PR #3, opened 2020-12-07
Added HomogeneousAtmosphere

leroyvn:
# Description

This PR closes eradiate/eradiate-issues#22. **Do not review until ready.**

# Checklist

### DIFF
--- a/eradiate/scenes/atmosphere/homogeneous.py
+++ b/eradiate/scenes/atmosphere/homogeneous.py
@@ -6,84 +6,105 @@ import eradiate
 from .base import Atmosphere
 from .radiative_properties.rayleigh import compute_sigma_s_air
 from ..core import SceneElementFactory
+from ..spectra import (
+    Spectrum,
+    UniformAlbedoSpectrum,
+    UniformCollisionCoefficientSpectrum,
+    validator_has_quantity
+)
 from ...util.attrs import (
     converter_or_auto,
-    validator_or_auto,
-    attrib_quantity,
-    converter_to_units,
-    validator_units_compatible,
-    validator_is_positive
+    validator_or_auto
 )
-from ...util.units import config_default_units as cdu
-from ...util.units import kernel_default_units as kdu
+from ...util.collections import onedict_value
 from ...util.units import ureg
 
 
-@SceneElementFactory.register("rayleigh_homogeneous")
+@SceneElementFactory.register("homogeneous")
 @attr.s()
-class RayleighHomogeneousAtmosphere(Atmosphere):
-    """Rayleigh homogeneous atmosphere scene element
-    [:factorykey:`rayleigh_homogeneous`].
+class HomogeneousAtmosphere(Atmosphere):
+    """Homogeneous atmosphere scene element [:factorykey:`homogeneous`].
 
-    This class builds an atmosphere consisting of a non-absorbing
-    homogeneous medium. Scattering uses the Rayleigh phase function and the
-    Rayleigh scattering coefficient of air at standard number density (
-    see :func:`sigma_s_air`).
+    This class builds an atmosphere consisting of a homogeneous medium.
+    Scattering uses the Rayleigh phase function.
 
     See :class:`~eradiate.scenes.atmosphere.base.Atmosphere` for undocumented
     members.
 
     .. rubric:: Constructor arguments / instance attributes
 
-    ``sigma_s`` (float or "auto"):
+    ``sigma_s`` (:class:`~eradiate.scenes.spectra.Spectrum` or "auto"):
         Atmosphere scattering coefficient value. If set to ``"auto"``,
         the scattering coefficient will be computed based on the current
         operational mode configuration using the :func:`sigma_s_air`
         function. Default: ``"auto"``.
 
-        Unit-enabled field (default unit: cdu[collision_coefficient]).
+        Can be initialised with a dictionary processed by
+        :class:`.SceneElementFactory`.
+
+    ``sigma_a`` (:class:`~eradiate.scenes.spectra.Spectrum`):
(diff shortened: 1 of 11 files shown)


leroyvn:
@nollety this PR is now ready for review. Would you mind taking a look?

nollety:
Sure! I'll take a look at it asap

eradiate/scenes/atmosphere/homogeneous.py
```diff
@@ -6,84 +6,105 @@
...
+
+    sigma_a = attr.ib(
+        factory=lambda: UniformCollisionCoefficientSpectrum(value=0.),
+        converter=Spectrum.converter("collision_coefficient"),
+        validator=[attr.validators.instance_of(Spectrum),
+                   validator_has_quantity("collision_coefficient")]
+    )
 
     @property
     def kernel_height(self):
```
nollety, line 72: Don't you think that the API docs look a little funny without the docstrings (same for `kernel_width`) since `kernel_offset` has a docstrings, but not the other `kernel_*` properties?

eradiate/scenes/atmosphere/homogeneous.py
```diff
@@ -6,84 +6,105 @@
...
+            return 10. / self._sigma_s.value
+        else:
+            return self.width
 
     @property
     def _albedo(self):
         """Return albedo."""
-        return 1.
+        return UniformAlbedoSpectrum(
+            value=self._sigma_s.value / (self._sigma_s.value + self.sigma_a.value)
```
nollety, line 89: I cannot try that at the moment, but what happens if one creates a homogeneous atmosphere with a zero scattering coefficient and a zero absorption coefficient? Do we check somewhere that the extinction coefficient is strictly positive?

eradiate/scenes/atmosphere/tests/test_homogeneous.py
```diff
@@ -37,43 +37,40 @@ def test_rayleigh_homogeneous(mode_mono, ref):
...
     with pytest.raises(ValueError):
-        RayleighHomogeneousAtmosphere(height=-100.)
+        HomogeneousAtmosphere(height=-100.)
 
     with pytest.raises(ValueError):
-        RayleighHomogeneousAtmosphere(width=-50.)
-
-    with pytest.raises(ValueError):
-        RayleighHomogeneousAtmosphere(sigma_s=-1e-7)
+        HomogeneousAtmosphere(width=-50.)
```
nollety, line 76: I'd add:
```python
with pytest.raises(ValueError):
    HomogeneousAtmosphere(sigma_s=0., sigma_a=0.)
```

because the albedo then is undefined.

eradiate/scenes/atmosphere/homogeneous.py
```diff
@@ -6,84 +6,105 @@
...
-        units_add_converter=False,
-        units_add_validator=False,
-    )  # TODO: turn into a Spectrum
+        converter=converter_or_auto(
+            Spectrum.converter("collision_coefficient")
+        ),
+        validator=validator_or_auto(
+            attr.validators.instance_of(Spectrum),
+            validator_has_quantity("collision_coefficient")
+        ),
```
nollety, line 61: Validate positive values only also?

eradiate/scenes/atmosphere/homogeneous.py
```diff
@@ -6,84 +6,105 @@
...
+    )
+
+    sigma_a = attr.ib(
+        factory=lambda: UniformCollisionCoefficientSpectrum(value=0.),
+        converter=Spectrum.converter("collision_coefficient"),
+        validator=[attr.validators.instance_of(Spectrum),
+                   validator_has_quantity("collision_coefficient")]
+    )
```
nollety, line 69: Validate positive values only also?

nollety (approved):
This looks great to me! My questions are about the validation of the `sigma_s` and `sigma_a` attributes, for the most part.

eradiate/scenes/atmosphere/homogeneous.py
leroyvn: Right. A pity that empty docstrings can't trigger some kind of inheritence from the parent class.

eradiate/scenes/atmosphere/homogeneous.py
leroyvn: No, but it's actually quite complicated. I currently don't have a reliable way (meaning something which won't break as soon as you'll start using a different spectrum type) to test for this upon object initialisation: we don't have spectrum arithmetics. So for the moment, I'll just leave this reponsibility to the user.

eradiate/scenes/atmosphere/tests/test_homogeneous.py
leroyvn: See my previous comment: this is unlikely to happen and checking for this in a way that makes sense, _i.e._ will really catch errors (say when you have non-uniform spectra with a few zeros in them), is quite challenging to implement.

eradiate/scenes/atmosphere/homogeneous.py
leroyvn: This should be taken care of by `Spectrum` IMO.

eradiate/scenes/atmosphere/homogeneous.py
leroyvn: This should be taken care of by Spectrum IMO.

leroyvn:
Thanks for the comments @nollety. Unfortunately, it's difficult to address the points you're mentioning due to how limited the `Spectrum` type is right now. I am in the process of refactoring it to make it more comprehensive and extensible (and add limited arithmetics to it). When we'll have this ready, we can improve validation.

PR #5, opened 2020-12-09
Spectrum refactoring

leroyvn:
# Description

This PR refactors `Spectrum`. Changes are as follows:

- The convoluted type-based ` Spectrum` subclassing system has been replaced with a simpler attribute-based tracking system for attached quantities. The new system uses the updated enum-based physical quantity specification introduced by cbde014b53c169c49fc31d33f7e51a23761586b9. Extending this system (_i.e._ adding a new `Spectrum` subclass or a new physical quantity) is much simpler than before.
- The `Spectrum` class still has a `convert()` method which allows for simplified dict-based initialisation for fields where the expected physical quantity is known (_e.g._ `ConstantIllumination.radiance` can be initialised with `{"type": "uniform", "value": 1.0}` and the appropriate quantity and units will be automatically added).
- The updated `UniformSpectrum` type supports basic arithmetics, currently as a proof of concept.
- The `util.attrs` module has been updated: we now distinguish between fields supporting units (which can be applied units when initialised with a dictionary) and fields which also have specified compatible units (which will be automatically checked for unit correctness and applied defaults). This information is stored as field metadata. Note that only the first is required to implement unit support; the second is used upon attribute definition to automate converter and validator definition, but similar behaviour can be implemented manually.
- Classes expecting spectra have been updated to use this new infrastructure.
- Moved `validator_has_quantity` to `util.attrs`.
- Various docstrings were updated or refreshed.
- Updated tutorial notebooks accordingly.

# Checklist

### DIFF
--- a/eradiate/scenes/biosphere.py
+++ b/eradiate/scenes/biosphere.py
@@ -12,25 +12,16 @@ import aabbtree
 import attr
 import numpy as np
 
-from .core import SceneElementFactory, SceneElement
-from .spectra import Spectrum, UniformReflectanceSpectrum, UniformTransmittanceSpectrum
+from .core import SceneElement, SceneElementFactory
+from .spectra import Spectrum
 from ..util.attrs import (
-    attrib_quantity, validator_has_len, validator_is_positive
+    attrib_quantity,
+    validator_has_len,
+    validator_has_quantity,
+    validator_is_positive
 )
 from ..util.units import config_default_units as cdu, ureg
 from ..util.units import kernel_default_units as kdu
-
-
-def _validator_has_quantity(quantity):
-    def f(_, attribute, value):
-        if value._quantity != quantity:
-            raise ValueError(
-                f"incompatible quantity '{value._quantity}' "
-                f"used to set field '{attribute.name}' "
-                f"(allowed: '{quantity}')"
-            )
-
-    return f
 
 
 @SceneElementFactory.register(name="homogeneous_discrete_canopy")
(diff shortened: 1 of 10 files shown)


leroyvn:
@schunkes @nollety Please have a quick look and tell me if you see something wrong.

eradiate/scenes/spectra.py
```diff
@@ -84,202 +102,117 @@ def f(value):
...
+    @property
+    def _values(self):
+        return self.value
+
+    def __add__(self, other):
+        # Preserve quantity field only if it is the same for both
+        # operands
+        if self.quantity is other.quantity and self.quantity is not None:
+            quantity = self.quantity
+        else:
```
schunkes, line 151: Why does this case have to be handled separately?

eradiate/scenes/biosphere.py
```diff
@@ -56,76 +47,58 @@ class HomogeneousDiscreteCanopy(SceneElement):
...
-        Seed for the random number generator
-
-        Default value: 1
+        Seed for the random number generator. Default: 1.
 
     ``avoid_overlap`` (bool):
-        If true, the scene element will attempt to place the leaves such
+        If ``True``, the scene element will attempt to place the leaves such
         that they do not overlap. If a leaf cannot be placed without overlap
-        after 1^6 tries, it will raise a RuntimeError.
```
schunkes, line 128: that would not have worked well :D

eradiate/scenes/spectra.py
leroyvn: I'm assuming you're referring to `self.quantity is not None`. Indeed, this can be removed.

eradiate/scenes/biosphere.py
leroyvn: :)

schunkes:
I think this looks fine :)

eradiate/scenes/biosphere.py
```diff
@@ -134,17 +107,13 @@ class HomogeneousDiscreteCanopy(SceneElement):
             different ``seed`` values and after a certain amount of
             failures, run it without overlap avoidance.
 
-        Default value: True
+    ``leaf_reflectance`` (float or :class:`~eradiate.scenes.spectra.Spectrum`):
+        Reflectance spectrum of the leaves in the cloud. Must be a reflectance
+        spectrum (dimensionless). Default: 0.5.
```
nollety, line 112: Why does it say "must be a reflectance spectrum" here? What does that sentence adds to the previous sentence, given that the accepted types are `float or :class:~eradiate.scenes.spectra.Spectrum`

nollety (approved):

eradiate/scenes/biosphere.py
leroyvn: This means that the spectrum must have `self.quantity == PhysicalQuantity.REFLECTANCE`. I still have to find a way to properly document this, but I don't think I can do it as part of this PR.

leroyvn:
Okay, I'll squash and merge!

PR #8, opened 2020-12-11
Allow unknown dataset attributes

nollety:
# Description

The dataset metadata validator will now allow unknown dataset attributes (i.e., other than `title`, `convention`, `source`, `history`, `references`), variables attributes and coordinates attributes (i.e., other than `standard_name`, `long_name` and `units`), as well as unknown coordinates/variables.

# Checklist

### DIFF
--- a/eradiate/util/xarray.py
+++ b/eradiate/util/xarray.py
@@ -9,7 +9,7 @@ from . import plot
 
 # -- Metadata processing -------------------------------------------------------
 
-def validate_metadata(data, spec, normalize=False):
+def validate_metadata(data, spec, normalize=False, allow_unknown=False):
     """Validate (and possibly normalise) metadata fields of the ``data``
     parameter based on a data specification.
 
@@ -23,6 +23,9 @@ def validate_metadata(data, spec, normalize=False):
     Parameter ``normalize`` (bool):
         If ``True``, also normalise metadata.
 
+    Parameter ``allow_unknown`` (bool):
+        If ``True``, allows unknown keys.
+
     Returns:
         If ``normalize`` is ``True``, normalised metadata dictionary; otherwise,
         unmodified metadata dictionary.
@@ -30,7 +33,7 @@ def validate_metadata(data, spec, normalize=False):
     Raises → ValueError:
         Got errors during metadata validation.
     """
-    v = cerberus.Validator(schema=spec.schema)
+    v = cerberus.Validator(schema=spec.schema, allow_unknown=allow_unknown)
     v.validate(data.attrs, normalize=normalize)
     if not v.errors:
         if normalize:
@@ -595,7 +598,7 @@ class EradiateDataArrayAccessor:
         """Wraps :func:`pplane`."""
         return pplane(self._obj, sza=sza, saa=saa)
 
-    def validate_metadata(self, var_spec, normalize=False):
+    def validate_metadata(self, var_spec, normalize=False, allow_unknown=True):
         """Validate the metadata for the wrapped :class:`~xarray.DataArray`.
         This function wraps :func:`validate_metadata`.
 
@@ -613,13 +616,19 @@ class EradiateDataArrayAccessor:
         # Validate variable metadata
         if var_spec.standard_name is not None:
             dataarray.attrs = validate_metadata(
-                dataarray, var_spec, normalize=normalize
+                data=dataarray,
+                spec=var_spec,
+                normalize=normalize,
+                allow_unknown=allow_unknown
             )
 
         # Validate dimension coordinate metadata
         for dim, coord_spec in var_spec.coord_specs.items():
             dataarray.coords[dim].attrs = validate_metadata(
-                dataarray.coords[dim], coord_spec, normalize=normalize
+                data=dataarray.coords[dim],
+                spec=coord_spec,
+                normalize=normalize,
+                allow_unknown=allow_unknown
             )
 
     def normalize_metadata(self, var_spec):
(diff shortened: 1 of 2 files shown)


leroyvn (commented):
Hi @yvan, the change you are proposing removes the possibility to reject unknown attributes (it's removed at the API level). I don't think this is desirable: validation should be conservative IMO. You have two possibilities, depending on when you want to allow for unknown fields:

- Use the [`allow_unknown`](https://docs.python-cerberus.org/en/stable/validation-rules.html#allow-unknown) validation rule in the validation schemas where you want to allow for unknown fields. Unknown variables will then always be allowed for the `Spec` subclass you'll modify.
- Add an `allow_unknown` parameter to the validation API (likely defaulting to `False` so as to retain existing behaviour). Validation will then still be conservative, with a possiblity to accept unknown attributes when relevant.

I would also appreciate if these changes could be unit tested if the behaviour is non-trivial and varies compared to the existing one.

nollety:
Thanks @leroyvn. I did not think of that. I'll go for the first possibility you suggest then: it seems like the most flexible option. I'll add unit tests too.

nollety:
I tried but I could not implement your first suggestion -- I lack proficiency with cerberus to do that -- so I went with your second suggestion, with the parameter `allow_unknown` defaulting to `False`. I also added the unit tests you requested.

leroyvn:
Hi @nollety, it seems to me that you have implemented a mix between the two approaches. If you're fine with the second, I don't think you actually need to change any Cerberus schema and the parameter forwarding you added to `validate_metadata()` is enough (don't forget to update the accessors, though).

eradiate/util/xarray.py
```diff
@@ -23,14 +23,17 @@ def validate_metadata(data, spec, normalize=False):
...
+
     Returns:
         If ``normalize`` is ``True``, normalised metadata dictionary; otherwise,
         unmodified metadata dictionary.
 
     Raises → ValueError:
         Got errors during metadata validation.
     """
-    v = cerberus.Validator(schema=spec.schema)
+    v = cerberus.Validator(schema=spec.schema, allow_unknown=allow_unknown)
```
leroyvn, line 36: It's all you have to do :)

eradiate/util/xarray.py
```diff
@@ -328,6 +331,9 @@ class DatasetSpec:
        Either all or none of ``convention``, ``title``, ``history``, ``source``
        and ``references`` must be ``None``.
 
+    ``allow_unknown`` (bool):
+        Allows unknown dataset attributes.
+
```
leroyvn, line 336: Not necessary

eradiate/util/xarray.py
```diff
@@ -357,6 +363,10 @@ class DatasetSpec:
         default=None,
         validator=attr.validators.optional(attr.validators.instance_of(str))
     )
+    allow_unknown = attr.ib(
+        default=False,
+        validator=attr.validators.instance_of(bool)
+    )
```
leroyvn, line 369: Not necessary

eradiate/util/xarray.py
```diff
@@ -656,7 +666,10 @@ def validate_metadata(self, dataset_spec, normalize=False):
         # Validate dataset metadata
         if dataset_spec.title is not None:
             dataset.attrs = validate_metadata(
-                dataset, dataset_spec, normalize=normalize
+                data=dataset,
+                spec=dataset_spec,
+                normalize=normalize,
+                allow_unknown=dataset_spec.allow_unknown
```
leroyvn, line 672: You can simply do `allow_unknown=allow_unknown` and make `allow_unknown` a parameter of `EradiateDatasetAccessor.validate_metadata()`. You should then propagate this behaviour to `EradiateDataArrayAccessor.validate_metadata()` as well.

nollety:
I understand. I thought that a dataset specification object (`DatasetSpec`) would indicate whether the unknown attributes are allowed or not, and not how `validate_data` is called . If `allow_unknown` is a parameter of `EradiateDatasetAccessor.validate_metadata` only, a given dataset `ds` will validate a specification or not based on how the method is called, i.e. `ds.ert.validate_data(spec, allow_unknown=True)` will not raise but `ds.ert.validate_data(spec, allow_unknown=False)` will.
Is that fine?

leroyvn:
If you go for solution 1, the specification tells whether or not unknown parameters are allowed. But then, there should not be any way to change this behaviour: `validate_metadata()` and its accessor proxies should not offer the option to tune this behaviour, since a user willing to control unknown parameters should use the specification for that.

If you go for solution 2, you implement `allow_unknown` forwarding in `validate_metadata()` and all accessor proxies. The lower-level function `validate_metadata()` will raise if a document doesn't validate, possibly tolerating unknown fields; you can then re-raise in accessor proxies if you want to make it clear that you know that this case happens (but honestly I think you don't have to). The behaviour is therefore that no spec allows for unknown parameters, but you allow for a bypass. If the user wants to differentiate behaviour between dataset, variable and/or coordinate metadata validations, they then have to do both steps manually (we're missing API components for that but I'm open to adding some).

nollety:
I went with solution 2 :)
I did not see an esay way to implement solution 1: according to the [cerberus usage docs](https://docs.python-cerberus.org/en/stable/usage.html#allowing-the-unknown), you can set `allow_unknown`:
- by setting the validator attribute of the same name, but the validator is not accessible from `DatasetSpec`,
- by editing directly the schema, but then `allow_unknown` only applies to nested mappings in that schema and attributes are not in a nested mapping of the `DatasetSpec` schema

nollety:
@leroyvn Following our discussion, `allow_unknown` is now passed to every calls to `validate_metadata()` with default value `True`, and is set to `False` when `validate_metadata()` is called within `normalize_metadata()`

leroyvn:
All good, let's go!

leroyvn:
@yvan would you mind squashing these commits prior to merging?

nollety:
@leroyvn there you go! :)

PR #11, opened 2020-12-16
Onedim solver refactoring

leroyvn:
# Description

This PR builds on #10 and closes eradiate/eradiate-issues#26. It finally moves `OneDimSolver` away from Cerberus-based object configuration and validation.

Main change:

- `OneDimSolverApp` is now configured and validated using attrs.

This was a fairly deep change and I have to revisit some code in the process.

Major changes:

- Refactored the SPP splitting code to improve its clarity. I added an attribute to control the SPP split threshold for testing. I also updated and modified the tests. I overall think that all radiancemeter measures should derive from a common type: we'll have to refactor that at some point.

Minor changes:

- Updated some docstrings.
- Updated a few function names for consistency.

To do:

- [x] Update CLI
- [x] Update tutorials
- [x] Update rest of documentation

# Checklist


eradiate/scenes/integrators.py
```diff
@@ -0,0 +1,181 @@
...
+@IntegratorFactory.register("volpathmis")
+@attr.s
+class VolPathMISIntegrator(MonteCarloIntegrator):
+    """A thin interface to the volumetric path tracer (with spectral multiple
+    importance sampling) kernel plugin
+    :cite:`Miller2019NullscatteringPathIntegral`.
+    """
+
+    use_spectral_mis = attr.ib(
+        default=None,
```
schunkes, line 159: Why is the default not `False` ?

eradiate/solvers/onedim/tests/test_app.py
```diff
@@ -132,67 +104,81 @@ def test_onedim_solver_app_run():
...
 
-    assert np.allclose(results_single["lo"], results_double["lo"], rtol=1e-5)
+    deviation_single = (
+            np.abs(results_double - results_single) / results_double
+    ).flatten()
+    deviation_single_split = (
+            np.abs(results_double - results_single_split) / results_double
+    ).flatten()
 
+    assert np.allclose(deviation_single, deviation_single_split, atol=5e-3)
```
schunkes, line 159: Isn't your test showing, that the SPP splitting mechanism is not working? Or do we expect the results to be close for 1e6 samples?

eradiate/scenes/integrators.py
leroyvn: `None` adds no entry to the kernel dictionary and lets Mitsuba apply its own defaults. If I apply a default and Mitsuba changes its own, mine will remain unchanged.

eradiate/solvers/onedim/tests/test_app.py
leroyvn: In order to demonstrate that SPP splitting works as expected, you have to show that it yields results similar to a computation without splitting. There is no perfect way of testing this: situations where double precision is really required are not something like this test case ([unless you plan on shooting 1e8+ rays](https://github.com/mitsuba-renderer/mitsuba2/issues/82)). 

What I did instead is force a split with a lower number of rays. With the values I used (I kept 1e6 for `spp`), the maximum SPP number in the second single-precision computation is 1e5: splitting occurs. But still, we have access, with 1e6 SPP, to an accurate value without splitting. I could arguably add something like this to also check that the single-precision reference is good:
```python
assert np.all(deviation_single < 1e-3)
```
(we still have quite a bit of noise though so the threshold may be inappropriate)

leroyvn: This however, indeed, doesn't show that splitting improves anything. For that, we'll need to run a much longer test case, and I don't think we can realistically put this in our automated testing suite.

eradiate/scenes/integrators.py
schunkes: Ahhhh

leroyvn:
I'm basically done with that now. I fixed a few things which were not working, but I also stumbled upon a bunch of bugs or impractical APIs which I can't reasonably fix right now. I'll make a final pass on docs and squash + merge along with #10.

leroyvn:
We're fine now: I'll squash and merge this!

PR #14, opened 2021-01-07
Added support to new us76_u86_4 datasets and throw away usa_mls

nollety:
# Description

I've updated the parts of the code that deal with the absorption datasets to support the new `us76_u86_4` datasets. The main new thing is the chunked datasets. These datasets have a smaller wavenumber range, 500 cm^-1, which makes them smaller in size and faster to load. The datasets are also named according to this convention:
```python
f"{engine}-{absorber}-{w_min}_{w_max}"
```
where
- `engine` is the engine used to compute the absorption cross section. Only the value `spectra` is supported at the moment, but future values could be: `hapi`, `radis`, `kspectrum`, etc
- `absorber` is the name of the absorbing gas species or mixture. Only the value `us76_u86_4` is supported at the moment, but future values could be: `n2`, `o2`, `co2`, `ch4`, `h2o`, `us62`, etc
- `w_min` and `w_max` are the minimum and maximum wavenumber values corresponding to this dataset

Example: `spectra-us76_u86_4-4000_25711` is the absorption dataset for the absorber `us76_u86_4` computed using the `SPECTRA` engine (unfortunately this name is a bit confusing) and its wavenumber range goes from 4000 to 25711 cm^-1.

This convention covers all cases I can think of so far

Changes are:
- `data.absorption_spectra`:
  - updated list of available datasets: got rid of mentions to obsolete `usa_mls` datasets.
  - added a function `available_datasets()` that scans the folder `resources/data/spectra/absorption` and returns the available (based on what the user has downloaded) datasets corresponding to a given absorber and wavenumber value.
- `radprops.absorption.compute_sigma_a()`: the function scans the available datasets before it loads the appropriate dataset, corresponding to the current wavelength value.
- radprops.rad_profile: updated `us76_u86_4-fullrange` to `spectra-us76_u86_4-4000_25711`

Note: if this pull request is approved, I'll remove `eradiate-data_us76_u86_4_fullrange.zip` from the ftp server --- it's replaced by `spectra-us76_u86_4-4000_25711.zip`.

# Checklist

### DIFF
--- a/eradiate/data/absorption_spectra.py
+++ b/eradiate/data/absorption_spectra.py
@@ -2,42 +2,227 @@
 
 **Category ID**: ``absorption_cross_section_spectrum``
 
+.. admonition:: Note
+
+   In the table below, wavenumber ranges are exact whereas wavelength ranges are approximative (numbers are rounded):w
+
+
 .. list-table:: Available data sets and corresponding identifiers
    :widths: 1 1 1 1
    :header-rows: 1
 
    * - Data set ID
      - Reference
-     - Spectral range [cm^-1]
-     - Spectral range [nm]
-   * - ``usa_mls-fullrange``
-     - ``eradiate-datasets_maker/scripts/spectra/acs/spectra/usa_mls.py``
-     - [18000, 19000]
-     - [~527, ~555]
-   * - ``usa_mls-fullrange``
-     - ``eradiate-datasets_maker/scripts/spectra/acs/spectra/usa_mls.py``
-     - [4000, 25711]
-     - [~389, 2500]
-   * - ``us76_u86_4-fullrange``
+     - Wavenumber range [cm^-1]
+     - Wavelength range [nm]
+   * - ``spectra-us76_u86_4-4000_25711``
      - ``eradiate-datasets_maker/scripts/spectra/acs/spectra/us76_u86_4.py``
-     - [4000, 25711]
-     - [~389, 2500]
+     - [4000.00413, 25710.993092]
+     - [389, 2500]
+   * - ``spectra-us76_u86_4-4000_4500``
+     - ``eradiate-datasets_maker/scripts/spectra/acs/spectra/us76_u86_4.py``
+     - [4000, 4500]
+     - [2222, 2500]
+   * - ``spectra-us76_u86_4-4500_5000``
+     - ``eradiate-datasets_maker/scripts/spectra/acs/spectra/us76_u86_4.py``
+     - [4500, 5000]
+     - [2000, 2222]
+   * - ``spectra-us76_u86_4-5000_5500``
+     - ``eradiate-datasets_maker/scripts/spectra/acs/spectra/us76_u86_4.py``
+     - [5000, 5500]
+     - [1818, 2000]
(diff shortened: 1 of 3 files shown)


eradiate/data/absorption_spectra.py
```diff
@@ -2,42 +2,221 @@
...
+     - [408, 416]
+   * - ``spectra-us76_u86_4-24500_25000``
+     - ``eradiate-datasets_maker/scripts/spectra/acs/spectra/us76_u86_4.py``
+     - [24500, 25000]
+     - [400, 408]
+   * - ``spectra-us76_u86_4-25000_25500``
+     - ``eradiate-datasets_maker/scripts/spectra/acs/spectra/us76_u86_4.py``
+     - [25000, 25500]
+     - [392, 400]
+   * - ``spectra-us76_u86_4-``
```
leroyvn, line 194: Aren't you missing a suffix here?

eradiate/data/absorption_spectra.py
```diff
@@ -2,42 +2,221 @@
...
 class _AbsorptionGetter(DataGetter):
     PATHS = {
-        "usa_mls-narrowrange": "spectra/absorption/usa_mls/narrowrange/*.nc",
-        "usa_mls-fullrange": "spectra/absorption/usa_mls/fullrange/*.nc",
-        "us76_u86_4-fullrange": "spectra/absorption/us76_u86_4/fullrange/*.nc",
-        "test": "tests/absorption/us76_u86_4/*.nc"
+        **{"spectra-us76_u86_4-4000_25711": "spectra/absorption/us76_u86_4/spectra-us76_u86_4-4000_25711/*.nc"},
+        **{f"{_US76_U86_4_PREF}-{x}_{x + 500}": f"{_US76_U86_4_PATH}/{_US76_U86_4_PREF}-{x}_{x + 500}/*.nc" for x in np.arange(4000, 25500, 500)},
+        **{f"{_US76_U86_4_PREF}-25500_25711": f"{_US76_U86_4_PATH}/{_US76_U86_4_PREF}-25500_25711/*.nc"},
+        **{"test": "tests/absorption/us76_u86_4/*.nc"}
```
leroyvn, line 219: Please format this to reduce line width to a reasonable value and remove unnecessary dict unpackings for improved legibility.

leroyvn (changes requested):
Hi @nollety, not much to say about this, just a bit of cleanup of the data index dict. I assume that the corresponding changes to data won't show up due to them being distributed outside of the data submodule, right?

eradiate/data/absorption_spectra.py
nollety: Yes, thanks for catching that!

nollety:
> Hi @nollety, not much to say about this, just a bit of cleanup of the data index dict. I assume that the corresponding changes to data won't show up due to them being distributed outside of the data submodule, right?

Hi @leroyvn, thanks for the feedback. You're right, the changes affect the datasets that are on the ftp server --- you need to download the last version of these datasets from the ftp server otherwise these changes will break the previous functionalities related to absorption calculations.

PR #16, opened 2021-01-13
us76_u86_4 fix

nollety:
# Description

This is a fix for the recent [8f6ff68](https://github.com/eradiate/eradiate/commit/8f6ff6815c3e03b572a846e1f41b3e2a3726da12) about supporting the new absorption datasets for the `us76_u86_4` gas mixture.

1. A function `available_datasets` was added to `data.absorption_spectra` module, that would scan the `resources/data/spectra/absorption/` directory. This function would work properly only if the python kernel is executed from the Eradiate root directory, which is not desirable. This was changed in the present pull request: the function finds the absorption datasets from whatever directory the python kernel was started in.

2. `US76RadProfile` default dataset was `spectra-us76_u86_4-4000_25711` (the large absorption data set) and was changed to `None`. The absorption coefficient values are computed with `compute_sigma_a` which will scan the `spectra/absorption` folder and automatically chooses the appropriate dataset to use.

# Checklist

- ~~[ ] The code is tested to prove its function~~

### DIFF
--- a/eradiate/data/absorption_spectra.py
+++ b/eradiate/data/absorption_spectra.py
@@ -198,7 +198,6 @@
 """
 
 import numpy as np
-import pathlib
 import xarray as xr
 
 from .core import DataGetter
@@ -248,7 +247,7 @@ class _AbsorptionGetter(DataGetter):
 
 @ureg.wraps(ret=None, args=("cm^-1", None, None), strict=False)
 def available_datasets(wavenumber, absorber="us76_u86_4", engine="spectra"):
-    """Returns the available datasets corresponding to a given wavenumber,
+    """Returns the available dataset(s) corresponding to a given wavenumber,
     absorber and absorption cross section engine.
 
     Parameter ``wavenumber`` (:class:`~pint.Quantity`):
@@ -261,20 +260,21 @@ def available_datasets(wavenumber, absorber="us76_u86_4", engine="spectra"):
         Engine used to compute the absorption cross sections.
 
     Returns → list or ``None``:
-        Available dataset ids.
+        Available dataset id(s).
     """
     if absorber == "us76_u86_4":
         if engine != "spectra":
             raise ValueError(f"engine {engine} is not supported.")
-        absorber_dir = pathlib.Path(f"resources/data/spectra/absorption/{absorber}")
-        subdirs = [f for f in absorber_dir.glob("*") if f.is_dir() and not f.name.startswith(".")]
         available = []
-        for subdir in subdirs:
-            _engine, _absorber, w_range = tuple(subdir.name.split("-"))
-            if _absorber == absorber and _engine == engine:
-                w_min, w_max = w_range.split("_")
-                if float(w_min) <= wavenumber < float(w_max):
-                    available.append(subdir.name)
+        for d in [_AbsorptionGetter.PATHS[k] for k in _AbsorptionGetter.PATHS
+                  if k != "test"]:
+            path = _presolver.resolve(d.strip("/*.nc"))
+            if path.is_absolute():
+                _engine, _absorber, w_range = tuple(path.name.split("-"))
+                if _absorber == absorber and _engine == engine:
+                    w_min, w_max = w_range.split("_")
+                    if float(w_min) <= wavenumber < float(w_max):
+                        available.append(path.name)
         return available if len(available) > 0 else None
     else:
         raise ValueError(f"absorber {absorber} is not supported.")
--- a/eradiate/radprops/rad_profile.py
+++ b/eradiate/radprops/rad_profile.py
@@ -37,6 +37,7 @@ import numpy as np
 import xarray as xr
 
 from eradiate import __version__
+from eradiate import data
 from .absorption import compute_sigma_a
 from .rayleigh import compute_sigma_s_air
 from ..thermoprops import us76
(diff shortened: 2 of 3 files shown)


eradiate/radprops/rad_profile.py
```diff
@@ -443,8 +443,7 @@ class US76ApproxRadProfile(RadProfile):
     )
 
     dataset = attr.ib(
-        default="spectra-us76_u86_4-4000_25711",
-        validator=attr.validators.in_({"spectra-us76_u86_4-4000_25711", "test"}),
+        default=None,
```
leroyvn, line 446: Without a validator, you won't get a meaningful error message when the user will enter wrong values. How does debugging work in that case? I think you should, at least, validate for the data type.

Then have a look below: what happens at line 484 if I set `self.dataset` to some random non-string value? Or to a string value which is not a dataset ID?

Finally, please don't forget to update the docstring when you update attributes.

leroyvn (commented):
Hi @nollety, I think your real problem is that you didn't use the path resolver to search for files. Your solution will break if I decide to scatter my data files in multiple locations; the path resolver can be used to prevent that.

Take a look at `solar_irradiance_spectral.py`:
```python
    PATHS = {
        "blackbody_sun": "spectra/solar_irradiance/blackbody_sun.nc",
        "meftah_2017": "spectra/solar_irradiance/meftah_2017.nc",
        "thuillier_2003": "spectra/solar_irradiance/thuillier_2003.nc",
        "whi_2008": "spectra/solar_irradiance/whi_2008_time_period_1.nc",  # alias
        "whi_2008_1": "spectra/solar_irradiance/whi_2008_time_period_1.nc",
        "whi_2008_2": "spectra/solar_irradiance/whi_2008_time_period_2.nc",
        "whi_2008_3": "spectra/solar_irradiance/whi_2008_time_period_3.nc",
    }
```
Paths are given relative to the data directory root; this is then passed to the path resolver.

Using this infrastructure, your absorber directory should be: `f"spectra/absorption/{absorber}"` (you can actually directly use the `PathResolver.glob()` method).

When you use the path resolver, don't forget that it's implemented as a singleton. See `solar_irradiance_spectral.py` for usage.

leroyvn:
Hi @nollety, any chance this can be solved this week?

nollety:
Hi @leroyvn, yes this on my today's to-do list!
I apologise for the delay, last week was too much work on absorption cross section computations, I should try to find a better balance between my tasks 😃

eradiate/radprops/rad_profile.py
nollety: > When you use the path resolver, don't forget that it's implemented as a singleton. See `solar_irradiance_spectral.py` for usage.

About the path resolver being a singleton, I have seen multiple instances of the `PathResolver` class, what do you mean then by that?

nollety:
I've added a validator for the `dataset` attribute of US76RadProfile. Note that we marked this attribute as "for testing/debugging purposes only"; that's why I did not feel the need to provide a validator.

eradiate/radprops/rad_profile.py
```diff
@@ -39,6 +39,7 @@
 from eradiate import __version__
 from .absorption import compute_sigma_a
 from .rayleigh import compute_sigma_s_air
+from ..data.absorption_spectra import _AbsorptionGetter
```
nollety, line 42: I assume that's not very clean, however I do not see another way to import the list of paths to the absorption datasets.

eradiate/radprops/rad_profile.py
nollety: Updated the docstrings!

nollety:
@leroyvn alright, I think this is ready for merge if you agree!

leroyvn (changes requested):
Hi @nollety, just realised that I didn't submit those comments yesterday... In short: you can use the getter getter (yes, that's not a typo ;) ) instead of importing the class directly. The other comments are more general and require deeper action.

eradiate/radprops/rad_profile.py
```diff
@@ -443,8 +444,9 @@ class US76ApproxRadProfile(RadProfile):
     )
 
     dataset = attr.ib(
-        default="spectra-us76_u86_4-4000_25711",
-        validator=attr.validators.in_({"spectra-us76_u86_4-4000_25711", "test"}),
+        default=None,
+        validator=attr.validators.optional(
+            attr.validators.in_(list(_AbsorptionGetter.PATHS.keys())))
```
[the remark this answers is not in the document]
nollety: The default behaviour lets `compute_sigma_a` finds the absorption dataset files automatically based on the thermophysical profile and wavelength/wavenumber value. The user must leave `dataset=None` to let that happen. In the tests, we call `compute_sigma_a` with `dataset_id="test"` because we cannot rely on the absorption data being downloaded, instead we want to use a small test dataset.

I will add that to the docstrings. I agree with your remark about `compute_sigma_a` not having defaults. I'll add an issue about that.

eradiate/data/absorption_spectra.py
```diff
@@ -248,7 +247,7 @@ def find(cls):
 
 @ureg.wraps(ret=None, args=("cm^-1", None, None), strict=False)
 def available_datasets(wavenumber, absorber="us76_u86_4", engine="spectra"):
```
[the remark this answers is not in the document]
nollety: No, you're right, we won't need this as soon as we drop the `spectra-us76_u86_4-4000_25711` aka the "big" absorption dataset.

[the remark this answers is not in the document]
nollety: I'll add that to the absorption module refactoring issue

leroyvn (approved):
Hi @nollety, seems good now! I'll merge this!

PR #17, opened 2021-01-13
Refactor hom dis canopy

schunkes:
# Description

This PR refactors the HomogeneousDiscreteCanopy class. It can now be constructed in two ways:
Through the parameters as before or through a file holding specifications for the individual leaves.

## Open questions

Variables like LAI and the canopy's extent are no longer attributes of the class. Should I add them anyways, to let the users get some information about the object? This is also problematic for the tests to this class. Currently we cannot test the properties of the canopy from the object.

It does seem strange to add attributes to the class that are only used for printing them out.

# Checklist

- [ ] The code is apropriately documented.


leroyvn:
Hi @schunkes, I can't review this now: you branched off from the draft `rami_solver_app` branch, which contains many other unrelated commits. Two possible solutions:

1. Wait until this branch is merged then rebase onto `main`, or
2. Branch off again from `main` by cherry picking relevant commits and force-pushing the feature branch.

leroyvn:
As for your questions:

> Variables like LAI and the canopy's extent are no longer attributes of the class. Should I add them anyways, to let the users get some information about the object? This is also problematic for the tests to this class. Currently we cannot test the properties of the canopy from the object.

Yes, these would be very useful. I think you can implement them as cached properties.

> It does seem strange to add attributes to the class that are only used for printing them out.

It's actually quite common: AFAIK that's what Python's `@property` feature exists for ;)

schunkes:
Superseded by #18

PR #18, opened 2021-01-13
Hom dis canopy new base

schunkes:
# Description
This PR supersedes #17 

# Checklist

- [ ] The code is apropriately documented.

### DIFF
--- a/eradiate/scenes/biosphere.py
+++ b/eradiate/scenes/biosphere.py
@@ -8,10 +8,12 @@
 """
 
 from abc import ABC
+__all__ = ["BiosphereFactory", "HomogeneousDiscreteCanopy"]
 
 import aabbtree
 import attr
 import numpy as np
+import os
 
 from .core import SceneElement
 from .spectra import Spectrum, SpectrumFactory
@@ -19,12 +21,12 @@ from ..util.attrs import (
     attrib_quantity,
     validator_has_len,
     validator_has_quantity,
-    validator_is_positive
+    validator_is_positive,
 )
 from ..util.factory import BaseFactory
 from ..util.units import config_default_units as cdu
 from ..util.units import kernel_default_units as kdu
-from ..util.units import ureg
+from ..util.units import ureg, ensure_units
 
 
 class BiosphereFactory(BaseFactory):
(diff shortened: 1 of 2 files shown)


schunkes:
Something we need to discuss:
Which informational variables should we include? I'd add the dimensions of the canopy for example.
Also: should the canopies have a function that prints all those infos nicely formatted?

schunkes:
I refactored the code again. I moved the leaf positioning into a separate method, which can now handle different shapes, cuboids, spheres and cylinders. Now the same code can be used for all canopies, which makes sense in my view.

I did not implement the overlap avoidance for spheres and cylinders and I wonder if we should remove it for the cube as well. Maybe we should ask Christian if this is really used.

leroyvn (commented):
Hi @schunkes, that's good! I made a series of comments, mostly on implementation details, programming style and documentation (which, if I'm not mistaken, you're working on now :) ).

leroyvn:
> Which informational variables should we include? I'd add the dimensions of the canopy for example.

Ideally, all statistical indicators used by the `from_parameters()` method should be accessible.

> Also: should the canopies have a function that prints all those infos nicely formatted?

I was actually thinking about tuning the `__repr__` function of our objects to make them more readable. This would require a little bit of thinking, ideas are welcome.

eradiate/scenes/biosphere.py
```diff
@@ -40,316 +40,468 @@ class BiosphereFactory(BaseFactory):
...
-
-    leaf_transmittance = attr.ib(
-        default=0.5,
-        converter=SpectrumFactory.converter("transmittance"),
-        validator=[
-            attr.validators.instance_of(Spectrum),
-            validator_has_quantity("transmittance")
-        ]
-    )
+            .. admonition:: Note
```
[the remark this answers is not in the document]
schunkes: Ah, I only saw it done this way in other documents.

eradiate/scenes/biosphere.py
```diff
@@ -40,316 +40,468 @@ class BiosphereFactory(BaseFactory):
...
+    while True:
+        rands = np.random.rand(2)
+        s1 = np.power(rands[0], 1. / mu)
+        s2 = np.power(rands[1], 1. / nu)
+        s = s1 + s2
+        if s <= 1:
+            return s1 / s
+
+
+def create_leaf_cloud(shape_dict, n_leaves, leaf_radius, mu, nu, avoid_overlap=False):
```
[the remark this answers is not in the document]
schunkes: That is a good thought, but it does not really work with the `shape_dict`.

What do you think: Should I add all four parameters (`cube_size`, `sphere_radius`, `cylinder_radius`, cylinder_height)` to the function signature?

eradiate/scenes/biosphere.py
```diff
@@ -40,316 +40,468 @@ class BiosphereFactory(BaseFactory):
...
+    if shape_dict["type"] == "cube":
+        try:
+            size = shape_dict["size"]
+        except KeyError:
+            raise KeyError("Parameter size is missing for leaf cloud creation.")
+        size_magnitude = size.to(ureg.m).magnitude
+        tree = aabbtree.AABBTree()
+
+        for i in range(n_leaves):
+            if not avoid_overlap:
```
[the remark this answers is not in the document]
schunkes: This one depends on the question if we add overlap avoidance to all shapes. I am not even sure we need it for the HomogeneousDiscreteCanopy, actually

schunkes:
> Ideally, all statistical indicators used by the `from_parameters()` method should be accessible.

Which are those?
The thing is, not all of them can be computed readily, I think. For example, `hdo` and `hvr` are interdependent. They can be used to specify the canopy, but especially for `hvr` I am not even entirely sure what it means. I adapted the docstring from raytran but there is some interpretation involved... And Yves said, these two parameters were added after his time.

> I was actually thinking about tuning the `__repr__` function of our objects to make them more readable. This would require a little bit of thinking, ideas are welcome.

That sounds nice, I'll try that out tomorrow.

schunkes:
I have to think more about `_lai` and `_n_leaves`. The information I pass to the class constructor actually makes it hard to compute these values.

schunkes:
> I was actually thinking about tuning the `__repr__` function of our objects to make them more readable. This would require a little bit of thinking, ideas are welcome.

As [this Post](https://www.pythontutorial.net/python-oop/python-__repr__/) suggests, we should rather implement the `__str__` method, than the `__repr__`

schunkes:
I added an implementation for the `__str__` method. There is just one thing there: I am not sure how to output the leaf radius. The text file canopy specification given by the RAMI scenes does allow for different leaf radii. We could give an average radius, but I am afraid, that users would be very confused if the method printed `Average leaf radius: 0.1m` when they specified only one...

eradiate/scenes/biosphere.py
```diff
@@ -229,10 +269,37 @@ class HomogeneousDiscreteCanopy(SceneElement):
     - The classmethod :meth:`~eradiate.scenes.biosphere.HomogeneousDiscreteCanopy.from_files`
       will read a set of definition files that specify the leaves of the canopy.
       Please refer to this method for details on the file format.
+
```
leroyvn, line 273: Typo on line 269 (one `s` too many)

leroyvn (changes requested):
Hi @schunkes,

After complete review and a bit of thinking, I'm concluding that storing transforms is not that much of a good approach and that you should rather compute them on the fly when calling `kernel_dict()`. The main motivation here is that otherwise you can't perform conversion to KDU, which complicates the implementation.

Therefore, I think you should rather store the basic information characterising the scatterers (meaning that you have to transfer a bit of code from your class constructors to `kernel_dict()`, with the added benefit of deduping this part). You can store all that data in large Numpy arrays to benefit from Pint's support and optimised memory usage.

> >  Ideally, all statistical indicators used by the from_parameters() method should be accessible.

> Which are those?
> The thing is, not all of them can be computed readily, I think. For example, hdo and hvr are interdependent. They can be used to specify the canopy, but especially for hvr I am not even entirely sure what it means. I adapted the docstring from raytran but there is some interpretation involved... And Yves said, these two parameters were added after his time.

I was hoping you could tell me that, I haven't looked into all the details of this ;) If you can't find any more, then don't bother.

> I have to think more about _lai and _n_leaves. The information I pass to the class constructor actually makes it hard to compute these values.

I assume leaf locations and sizes should be enough to compute that, right? That would be another added benefit of storing scatterer parameters rather than transforms.

eradiate/scenes/biosphere.py
```diff
@@ -325,6 +332,22 @@ class HomogeneousDiscreteCanopy(SceneElement):
         ]
     )
 
+    leaf_positions = attrib_quantity(
+        default=ureg.Quantity(np.array([[0, 0, 0]]), ureg.m),
+        units_compatible=cdu.generator("length"),
+    )
+
+    leaf_orientations = attrib_quantity(
```
leroyvn, line 340: Since this is meant to be unitless, you can actually declare it as a regular `attr.ib`: you don't need the extra bells and whistles for conversion. You'll however need to validate for consistency with the other attributes.

eradiate/scenes/biosphere.py
```diff
@@ -485,21 +527,17 @@ def from_parameters(cls, size=[30, 30, 3]*ureg.m, lai=3, mu=1.066, nu=1.853,
...
-                     center_position=center_position,
-                     transforms=transforms)
-
-        canopy._lai = lai
-        canopy._size = size * ureg.m
-        canopy._n_leaves = n_leaves
-
-        return canopy
+                     center_position=center_position, leaf_radius=leaf_radius,
+                     leaf_positions=positions, leaf_orientations=orientations)
```
leroyvn, line 540: Formatting ;)

eradiate/scenes/biosphere.py
```diff
@@ -623,14 +621,27 @@ def kernel_dict(self, ref=True):
             }
         }
 
-        for i, transform in enumerate(self.transforms):
+        #
```
leroyvn, line 624: Should there be a comment here?

eradiate/scenes/biosphere.py
```diff
@@ -281,6 +276,21 @@ class HomogeneousDiscreteCanopy(SceneElement):
         Number of leaves in the canopy. Used only to display information about
         the object.
 
+    ``leaf_positions`` (list[list[float]]):
+        Lists all leaf positions as three vectors in cartesian coordinates.
+        Default: [[0., 0., 0.]] m.
```
leroyvn, line 281: I'd rather set no default here or add validators to check that `leaf_positions` and `leaf_orientations` are consistent.

eradiate/scenes/biosphere.py
```diff
@@ -281,6 +276,21 @@ class HomogeneousDiscreteCanopy(SceneElement):
...
+
+        Unit-enabled field (default units: cdu[length])
+
+    ``leaf_orientations`` (list[list[float]]):
```
leroyvn, line 285: Same here.

eradiate/scenes/biosphere.py
```diff
@@ -538,77 +576,37 @@ def from_file(cls, file_path, leaf_reflectance=0.5, leaf_transmittance=0.5,
...
-                           abs(max_y-min_y),
-                           abs(max_z-min_z)] * ureg.m
+        positions = np.array(positions_) * ureg.m
+        orientations = np.array(orientations_) * ureg.dimensionless
 
-            return canopy
+        return cls(leaf_reflectance=leaf_reflectance,
+                     leaf_transmittance=leaf_transmittance,
+                     center_position=center_position, leaf_radius=radius,
+                     leaf_positions=positions, leaf_orientations=orientations)
```
leroyvn, line 605: Formatting

leroyvn (changes requested):
Alright @schunkes, I think we're almost there. Apart from the small details I'm mentioning, I think you haven't accounted for the suggestions Yves made about hints at physically relevant values for the `from_parameters()` constructor; could please make a sweep through the email he sent and add them where relevant? Also, I'd suggest rearranging the arguments (and docs) of t his function to match the priority order he mentioned during our discussion: this way, the "common" parameters will show before the "expert" ones.

We're also missing the dict interface (you'll have to reimplement `from_dict()`, but we'll add that as part of another PR. Please file an issue for that so that we keep track of it.

schunkes:
About the `from_dict` method: No other scene element has this, so I didn't implement this now. 

About the parameter order: This is already the ordering that Yves suggested. It is just that we don't have the canopy height isolated, because the scene element needs a size in all dimensions.

leroyvn:
Hi @schunkes, we're there! Good job, let's merge this. Would you mind squashing this PR into a single commit?

leroyvn (approved):

PR #20, opened 2021-01-15
39 reparametrize perspective camera

schunkes:
# Description

Here's my proposition. I could still add a few tests. Closes eradiate/eradiate-issues#39.

# Checklist

### DIFF
--- a/eradiate/scenes/measure.py
+++ b/eradiate/scenes/measure.py
@@ -496,7 +496,7 @@ class PerspectiveCameraMeasure(Measure):
     @target.validator
     @origin.validator
     def _target_origin_validator(self, attribute, value):
-        if np.all(self.target == self.origin):
+        if np.allclose(self.target, self.origin):
             raise ValueError(
                 f"While initializing {attribute}:"
                 f"Origin and target must not be equal,"
--- a/eradiate/scenes/measure.py
+++ b/eradiate/scenes/measure.py
@@ -23,11 +23,11 @@ from ..util.attrs import (
     validator_is_number,
     validator_is_positive,
     validator_is_vector3,
-    validator_quantity
+    validator_quantity,
 )
 from ..util.collections import is_vector3
 from ..util.factory import BaseFactory
-from ..util.frame import angles_to_direction, spherical_to_cartesian
+from ..util.frame import angles_to_direction
 from ..util.misc import always_iterable
 from ..util.units import config_default_units as cdu
 from ..util.units import kernel_default_units as kdu
@@ -84,6 +84,7 @@ class MeasureFactory(BaseFactory):
        .. factorytable::
           :factory: MeasureFactory
     """
+
     _constructed_type = Measure
     registry = {}
 
@@ -157,14 +158,17 @@ class TargetPoint(Target):
 
         Unit-enabled field (default: cdu[length]).
     """
+
     # Target point in CDU
     xyz = attrib_quantity(units_compatible=cdu.generator("length"))
 
     @xyz.validator
     def _xyz_validator(self, attribute, value):
         if not is_vector3(value):
-            raise ValueError(f"while validating {attribute.name}: must be a "
-                             f"3-element vector of numbers")
+            raise ValueError(
+                f"while validating {attribute.name}: must be a "
+                f"3-element vector of numbers"
+            )
 
     def kernel_item(self):
         """Return kernel item."""
@@ -201,6 +205,7 @@ class TargetRectangle(Target):
         Unit-enabled field (default: cdu[length]).
     """
 
+    # fmt: off
     # Corners of an axis-aligned rectangle in CDU
     xmin = attrib_quantity(
         converter=converter_quantity(float),
@@ -222,20 +227,25 @@ class TargetRectangle(Target):
         validator=validator_quantity(validator_is_number),
         units_compatible=cdu.generator("length")
     )
+    # fmt: on
 
     @xmin.validator
     @xmax.validator
     def _x_validator(self, attribute, value):
         if not self.xmin < self.xmax:
-            raise ValueError(f"while validating {attribute.name}: 'xmin' must "
-                             f"be lower than 'xmax")
+            raise ValueError(
+                f"while validating {attribute.name}: 'xmin' must "
+                f"be lower than 'xmax"
+            )
 
     @ymin.validator
     @ymax.validator
     def _y_validator(self, attribute, value):
         if not self.ymin < self.ymax:
-            raise ValueError(f"while validating {attribute.name}: 'ymin' must "
-                             f"be lower than 'ymax")
+            raise ValueError(
+                f"while validating {attribute.name}: 'ymin' must "
+                f"be lower than 'ymax"
+            )
 
     def kernel_item(self):
         """Return kernel item."""
@@ -249,14 +259,11 @@ class TargetRectangle(Target):
         dx = xmax - xmin
         dy = ymax - ymin
 
-        to_world = \
-            ScalarTransform4f.translate([0.5 * dx + xmin, 0.5 * dy + ymin, 0]) * \
-            ScalarTransform4f.scale([0.5 * dx, 0.5 * dy, 1])
+        to_world = ScalarTransform4f.translate(
+            [0.5 * dx + xmin, 0.5 * dy + ymin, 0]
+        ) * ScalarTransform4f.scale([0.5 * dx, 0.5 * dy, 1])
 
-        return {
-            "type": "rectangle",
-            "to_world": to_world
-        }
+        return {"type": "rectangle", "to_world": to_world}
 
 
 @MeasureFactory.register("distant")
@@ -298,6 +305,13 @@ class DistantMeasure(Measure):
     ``flip_directions`` (bool):
         If ``True``, sampled directions will be flipped. Default: False.
     """
+
+    # fmt: off
+    zenith = attrib_quantity(
+        default=ureg.Quantity(0., ureg.deg),
+        validator=validator_is_positive,
+        units_compatible=cdu.generator("angle"),
+    )
     film_resolution = attr.ib(
         default=(32, 32),
         validator=attr.validators.deep_iterable(
@@ -335,6 +349,7 @@ class DistantMeasure(Measure):
         default=None,
         converter=attr.converters.optional(bool)
     )
+    # fmt: on
 
     def sensor_info(self):
         """This method returns a tuple of sensor IDs and the corresponding SPP
@@ -366,6 +381,7 @@ class DistantMeasure(Measure):
 
         return np.reshape(data, (data.shape[0], data.shape[1]))
 
+    # fmt: off
     def kernel_dict(self, ref=True):
         result = {
             "type": "distant",
@@ -398,6 +414,7 @@ class DistantMeasure(Measure):
             result["flip_directions"] = self.flip_directions
 
         return {self.id: result}
+    # fmt: on
 
 
 @MeasureFactory.register("perspective")
@@ -406,9 +423,8 @@ class PerspectiveCameraMeasure(Measure):
     """Perspective camera scene element [:factorykey:`perspective`].
 
     This scene element is a thin wrapper around the ``perspective`` sensor
-    kernel plugin. It positions a perspective camera based on the a pair of
-    zenith and azimuth angles, following the convention used in Earth
-    observation. The film is a square.
+    kernel plugin. It positions a perspective camera based on a set of vectors,
+    specifying the origin, viewing direction and 'up' direction of the camera.
 
     .. rubric:: Constructor arguments / instance attributes
 
@@ -418,54 +434,53 @@ class PerspectiveCameraMeasure(Measure):
 
         Unit-enabled field (default: cdu[length]).
 
-    ``zenith`` (float):
-        Zenith angle. Default value: 0 deg.
-
-        Unit-enabled field (default: cdu[angle]).
-
-    ``azimuth`` (float):
-        Azimuth angle value. Default value: 0 deg.
-
-        Unit-enabled field (default: cdu[angle]).
-
-    ``distance`` (float):
-        Distance from the ``target`` point to the camera.
-        Default: 1 km.
+    ``origin`` (array[float]):
+        A 3-element vector specifying the position of the camera.
+        Default: [1, 1, 1] m.
 
         Unit-enabled field (default: cdu[length]).
 
-    ``res`` (int):
-        Resolution of the film in pixels. Default: 64.
+    ``up`` (array[float]):
+        A 3-element vector specifying the up direction of the camera.
+        This vector must be different from the camera's viewing direction,
+        which is given by ``target - origin``.
+        Default: [0, 0, 1].
+
+    ``film_width`` (int):
+        Horizontal resolution of the film in pixels. Default: 64.
+
+    ``film_height`` (int):
+        Vertical resolution of the film in pixels. Default: 64.
 
     ``spp`` (int):
         Number of samples per pixel. Default: 32.
     """
 
+    # fmt: off
     target = attrib_quantity(
         default=ureg.Quantity([0, 0, 0], ureg.m),
         validator=validator_has_len(3),
         units_compatible=cdu.generator("length"),
     )
 
-    zenith = attrib_quantity(
-        default=ureg.Quantity(0., ureg.deg),
-        validator=validator_is_positive,
-        units_compatible=cdu.generator("angle"),
-    )
-
-    azimuth = attrib_quantity(
-        default=ureg.Quantity(0., ureg.deg),
-        validator=validator_is_positive,
-        units_compatible=cdu.generator("angle"),
-    )
-
-    distance = attrib_quantity(
-        default=ureg.Quantity(1., ureg.km),
-        validator=validator_is_positive,
+    origin = attrib_quantity(
+        default=ureg.Quantity([1, 1, 1], ureg.m),
+        validator=validator_has_len(3),
         units_compatible=cdu.generator("length"),
     )
 
-    res = attr.ib(
+    up = attr.ib(
+        default=[0, 0, 1],
+        validator=validator_has_len(3)
+    )
+
+    film_width = attr.ib(
+        default=64,
+        converter=int,
+        validator=validator_is_positive
+    )
+
+    film_height = attr.ib(
         default=64,
         converter=int,
         validator=validator_is_positive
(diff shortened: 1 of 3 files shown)


eradiate/scenes/measure.py
```diff
@@ -483,29 +524,37 @@ def kernel_dict(self, ref=True):
...
-        origin = spherical_to_cartesian(distance, zenith, azimuth)
+        if self.origin_type == "angles":
+            origin = spherical_to_cartesian(distance, zenith,
+                                            azimuth, origin=target)
+        elif self.origin_type == "point":
+            origin = self.origin.to(kdu.get("length")).magnitude
+        else:
+            raise ValueError(f"Unsupported origin type: {self.origin_type}")
+        
         direction = origin / np.linalg.norm(origin)
```
leroyvn, line 535: You don't need this anymore, I think.

eradiate/scenes/measure.py
```diff
@@ -483,29 +524,37 @@ def kernel_dict(self, ref=True):
...
         if np.allclose(origin, target):
             raise ValueError("target is too close to the camera")
 
-        up = [np.cos(azimuth), np.sin(azimuth), 0] \
-            if np.allclose(direction, [0, 0, 1]) \
-            else [0, 0, 1]
+        if np.allclose(self.up, [0, 0, 0]):
+            self.up = [np.cos(azimuth), np.sin(azimuth), 0] \
+                if np.allclose(direction, [0, 0, 1]) \
+                else [0, 0, 1]
```
leroyvn, line 543: You can remove this too.

eradiate/scenes/measure.py
```diff
@@ -417,6 +436,17 @@ class PerspectiveCameraMeasure(Measure):
...
+        validator=validator_has_len(3),
+        units_compatible=cdu.generator("length"),
+    )
+
+    up = attr.ib(
+        default=[0, 0, 1],
+        validator=validator_has_len(3)
+    )
+
     zenith = attrib_quantity(
```
leroyvn, line 450: I'd like to get rid of zenith and azimuth parameters completely, the purpose of this interface change is to remove them in favour of something more flexible.

eradiate/scenes/measure.py
```diff
@@ -404,8 +415,16 @@ class PerspectiveCameraMeasure(Measure):
 
         Unit-enabled field (default: cdu[length]).
 
-    ``res`` (int):
-        Resolution of the film in pixels. Default: 64.
+    ``origin_type`` (str):
+        Distinguishes camera placement by origin point and zenith/azmiuth angles.
+        Can be set to ``point`` or ``angles``.
+        Default: ``angles``
```
leroyvn, line 421: No longer needed if you remove zenith and azimuth.

leroyvn (changes requested):
Hi @schunkes, good start. However the purpose of this issue is to change the interface, not to extend it. I'd like to replace the `distance`, `zenith` and `azimuth` parameters with the parameters of a look-at transform.

This behaviour would still be achievable with the updated `spherical_to_cartesian()`:
```python
# Old
PerspectiveCameraMeasure(zenith=30, azimuth=45, distance=10)
# New
PerspectiveCameraMeasure(target=[0, 0, 1], origin=spherical_to_cartesian(10, 30, 45, [0, 0, 1], up=[0, 0, 1])
```
If you really want to reimplement the old behaviour, I'd rather have it written as a class method constructor.

schunkes:
Good point. Fixed it.

eradiate/scenes/measure.py
```diff
@@ -388,24 +389,21 @@ class PerspectiveCameraMeasure(Measure):
...
 
-        Unit-enabled field (default: cdu[angle]).
+        Unit-enabled field (default: cdu[length]).
 
-    ``distance`` (float):
-        Distance from the ``target`` point to the camera.
-        Default: 1 km.
+    ``up`` (array[float]):
+        A 3-element vector specifying the vertical direction of the camera.
+        Default: [0, 0, 0].
```
leroyvn, line 400: I don't think this default can work (at least I don't see how). A better value would be [0, 0, 1] IMO. Also I'm a bit uncomfortable with calling this "vertical" because, well... If you set it to something else than [0, 0, 1], it's not the vertical anymore (assuming gravity is oriented to -Z, ofc).

Sticking with "up" is better IMO. Something like: "A 3-element vector specifying the "up" direction of the camera."

eradiate/scenes/measure.py
```diff
@@ -388,24 +389,21 @@ class PerspectiveCameraMeasure(Measure):
...
-    ``zenith`` (float):
-        Zenith angle. Default value: 0 deg.
-
-        Unit-enabled field (default: cdu[angle]).
-
-    ``azimuth`` (float):
-        Azimuth angle value. Default value: 0 deg.
+    ``origin`` (array[float]):
+        A 3-element vector specifying the position of the camera.
+        Default: [0, 0, 0] m.
```
leroyvn, line 394: Your defaults are currently `origin == target`: this will result in default params making `kernel_dict()` raise (BTW I think the origin-target distance, if checked, should be checked during the init sequence, not when expanding to dictionary). Maybe select a default value positioning `origin` a couple dozen metres aways from `target`?

eradiate/scenes/measure.py
```diff
@@ -479,33 +476,30 @@ def kernel_dict(self, ref=True):
...
-        distance = self.distance.to(kdu.get("length")).magnitude
-        zenith = self.zenith.to("rad").magnitude
-        azimuth = self.azimuth.to("rad").magnitude
-
-        origin = spherical_to_cartesian(distance, zenith, azimuth)
-        direction = origin / np.linalg.norm(origin)
+        origin = self.origin.to(kdu.get("length")).magnitude
 
         if np.allclose(origin, target):
             raise ValueError("target is too close to the camera")
```
leroyvn, line 482: Here is what I'm talking about: you should actually catch this during field validation. You can use a common validator to `origin` and `target` using the decorator validator syntax.

leroyvn (changes requested):
Hi @schunkes, a couple more details I noticed and we'll be done, the interface will be very nice.

eradiate/scenes/measure.py
```diff
@@ -479,33 +475,30 @@ def kernel_dict(self, ref=True):
...
 
-        up = [np.cos(azimuth), np.sin(azimuth), 0] \
-            if np.allclose(direction, [0, 0, 1]) \
+        direction = origin - target
+
+        self.up = [1, 0, 0] \
+            if np.dot(direction, [0, 0, 1]) == 0\
             else [0, 0, 1]
```
leroyvn, line 487: You're ignoring the value passed by the user for `self.up` here contrary to what you wrote in the docstring. I think the simplest is to drop that part and let the user set a reasonable value for `up` themselves.

Or, if you want to keep this behaviour, you can have it handled by a special value (_e.g._ `None`). In that case, it's probably best to implement it as part of the post-init step (`kernel_dict()` should preferably not alter the object's state). As you prefer (you'd have to document that ofc).

eradiate/util/attrs.py
```diff
@@ -460,3 +460,22 @@ def f(instance, attribute, value):
...
+        Parameter ``other`` (attribute of instance)
+            Other attribute, which must be different from 'value'
+
+        """
+        other_value = getattr(instance, other)
+        if np.all(value == other_value):
+            raise AttributeError(f"'{attribute}' must be different "
+                                 f"from '{other}'!")
+
+    return validator_vector_nequal_other
```
leroyvn, line 481: As we discussed on Slack, I don't think you need to introduce this to implement the validation procedure. I'd rather stick to the decorator-based approach, which doesn't require intermediate components.

eradiate/util/attrs.py
```diff
@@ -460,3 +460,22 @@ def f(instance, attribute, value):
...
+        to numpy arrays.
+
+        Parameter ``other`` (attribute of instance)
+            Other attribute, which must be different from 'value'
+
+        """
+        other_value = getattr(instance, other)
+        if np.all(value == other_value):
+            raise AttributeError(f"'{attribute}' must be different "
+                                 f"from '{other}'!")
```
leroyvn, line 479: `AttributeError` is not appropriate in such a case IMO (see _e.g._ [Python docs](https://docs.python.org/3/library/exceptions.html#AttributeError)). A `ValueError` is more appropriate. See [attrs docs](https://www.attrs.org/en/stable/api.html#attr.validators.in_) for examples of exceptions raised by validators.

leroyvn (changes requested):
Hi @schunkes, I solidified the discussions we had on Slack and spotted a few more small issues which should be easy to solve. When that will be done, we'll be good to go!

schunkes:
Alright, I am now using an ad hoc validator for target and origin and I removed the section of code, checking and overwriting `up`. Instead I added another validator, that makes sure, `up` is not the same as `origin + target`

eradiate/scenes/tests/test_measure.py
```diff
@@ -112,6 +112,9 @@ def test_perspective(mode_mono):
     d = PerspectiveCameraMeasure()
     assert KernelDict.empty().add(d).load() is not None
 
+    with pytest.raises(AttributeError):
+        d = PerspectiveCameraMeasure(origin=[0,0,0], target=[0,0,0])
```
leroyvn, line 116: Please try with a value other that [0, 0, 0] (this test could not detect the mistakes you made).

eradiate/scenes/tests/test_measure.py
```diff
@@ -112,6 +112,9 @@ def test_perspective(mode_mono):
     d = PerspectiveCameraMeasure()
     assert KernelDict.empty().add(d).load() is not None
 
+    with pytest.raises(AttributeError):
```
leroyvn, line 115: Does this test actually pass? Your validators raise `ValueError`s, I think.

leroyvn (changes requested):
Hi @schunkes, good! The validator code is elegant, we're almost there. I added a couple of comments, should be quick and easy to fix. Once you're done, please don't forget to rebase and squash, then make a pass on the checklist.

eradiate/scenes/measure.py
```diff
@@ -443,6 +443,22 @@ class PerspectiveCameraMeasure(Measure):
...
+                             f"Origin and target must not be equal,"
+                             f"got target = {self.target}, origin = {self.origin}")
+
+    @up.validator
+    def _up_validator(self, attribute, value):
+        direction = self.origin + self.target
+        if np.allclose(np.dot(direction, value), 0):
+            raise ValueError(f"While initializing {attribute}:"
+                             f"Up direction must differ from viewing direction,"
+                             f"got up = {self.up}, view direction = {direction}.")
```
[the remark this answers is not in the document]
schunkes: Ah crap, I mixed it up :sweat_smile:

eradiate/scenes/measure.py
```diff
@@ -443,6 +443,22 @@ class PerspectiveCameraMeasure(Measure):
         validator=validator_is_positive
     )
 
+    @target.validator
+    @origin.validator
+    def _target_origin_validator(self, attribute, value):
+        if np.all(self.target == self.origin):
```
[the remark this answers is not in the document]
schunkes: Strictly speaking yes: The tolerance for `allclose()` is non-zero but even a tiny difference would allow for a valid specification of a `look_at` transform.

eradiate/scenes/measure.py
```diff
@@ -447,6 +443,22 @@ class PerspectiveCameraMeasure(Measure):
...
+            raise ValueError(f"While initializing {attribute}:"
+                             f"Origin and target must not be equal,"
+                             f"got target = {self.target}, origin = {self.origin}")
+
+    @up.validator
+    def _up_validator(self, attribute, value):
+        direction = self.origin + self.target
```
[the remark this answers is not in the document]
schunkes: it's `target - origin` but yes, I made a mistake :)

eradiate/scenes/measure.py
```diff
@@ -201,6 +205,7 @@ class TargetRectangle(Target):
         Unit-enabled field (default: cdu[length]).
     """
 
+    # fmt: off
```
leroyvn, line 208: This will have unintended side effects: you're basically disabling formatting on half of the file. This is okay for editing, but you should not leave the marks when you're done: other maintainers working on the code will likely want to format code and they shouldn't be looking at formatter markup to have it work.

We're not enforcing black yet, it's a very disruptive change (I imagine you've been as scared as I have by how meaningless diffs become when you black-format the entire project); but I haven't changed my mind and I want it to be our formatting tool in the future. A possible solution for transition (haven't tried it yet but seems capable) is [darker](https://pypi.org/project/darker/): it will apply formatting only to regions of the code which have been modified since you've last committed. The tool is flexible and has many options as well as IDE integration instructions. Only minor issue: it's not packaged for Conda (but I can now create a recipe to package it myself).

```suggestion
```

eradiate/scenes/measure.py
```diff
@@ -222,20 +227,25 @@ class TargetRectangle(Target):
         validator=validator_quantity(validator_is_number),
         units_compatible=cdu.generator("length")
     )
+    # fmt: on
```
leroyvn, line 230:
```suggestion
```

eradiate/scenes/measure.py
```diff
@@ -298,6 +305,13 @@ class DistantMeasure(Measure):
     ``flip_directions`` (bool):
         If ``True``, sampled directions will be flipped. Default: False.
     """
+
+    # fmt: off
```
leroyvn, line 309:
```suggestion
```

eradiate/scenes/measure.py
```diff
@@ -335,6 +349,7 @@ class DistantMeasure(Measure):
         default=None,
         converter=attr.converters.optional(bool)
     )
+    # fmt: on
```
leroyvn, line 352:
```suggestion
```

eradiate/scenes/measure.py
```diff
@@ -366,6 +381,7 @@ def postprocess_results(self, sensor_ids, sensor_spp, results):
 
         return np.reshape(data, (data.shape[0], data.shape[1]))
 
+    # fmt: off
```
leroyvn, line 384:
```suggestion
```

eradiate/scenes/measure.py
```diff
@@ -398,6 +414,7 @@ def kernel_dict(self, ref=True):
             result["flip_directions"] = self.flip_directions
 
         return {self.id: result}
+    # fmt: on
```
leroyvn, line 417:
```suggestion
```

eradiate/scenes/measure.py
```diff
@@ -418,54 +434,53 @@ class PerspectiveCameraMeasure(Measure):
...
-    ``res`` (int):
-        Resolution of the film in pixels. Default: 64.
+    ``film_height`` (int):
+        Vertical resolution of the film in pixels. Default: 64.
 
     ``spp`` (int):
         Number of samples per pixel. Default: 32.
     """
 
+    # fmt: off
```
leroyvn, line 459:
```suggestion
```

eradiate/scenes/measure.py
```diff
@@ -476,6 +491,27 @@ class PerspectiveCameraMeasure(Measure):
         converter=int,
         validator=validator_is_positive
     )
+    # fmt: on
```
leroyvn, line 494:
```suggestion
```

eradiate/scenes/measure.py
```diff
@@ -505,43 +541,33 @@ def sensor_info(self):
         SPP levels that lead to numerical precision loss in results."""
         return [(self.id, self.spp)]
 
+    # fmt: off
```
leroyvn, line 544:
```suggestion
```

eradiate/scenes/measure.py
```diff
@@ -505,43 +541,33 @@ def sensor_info(self):
...
-                    "height": self.res,
+                    "width": self.film_width,
+                    "height": self.film_height,
                     "pixel_format": "luminance",
                     "component_format": "float32",
                     "rfilter": {"type": "box"}
                 }
             }
         }
+    # fmt: on
```
leroyvn, line 570:
```suggestion
```

eradiate/scenes/measure.py
```diff
@@ -605,6 +631,7 @@ class RadianceMeterHsphereMeasure(Measure):
         Default: ``"radiancemeter_hsphere"``.
     """
 
+    # fmt: off
```
leroyvn, line 634:
```suggestion
```

eradiate/scenes/measure.py
```diff
@@ -659,15 +686,14 @@ class RadianceMeterHsphereMeasure(Measure):
 
     _zenith_angles = attr.ib(default=None, init=False)  # Set during post-init
     _azimuth_angles = attr.ib(default=None, init=False)  # Set during post-init
+    # fmt: on
```
leroyvn, line 689:
```suggestion
```

eradiate/scenes/measure.py
```diff
@@ -751,6 +787,7 @@ def sensor_info(self):
         else:
             return [(self.id, self.spp)]
 
+    # fmt: off
```
leroyvn, line 790:
```suggestion
```

eradiate/scenes/measure.py
```diff
@@ -782,6 +819,7 @@ def kernel_dict(self, **kwargs):
             kernel_dict[sensor_id] = sensor_dict
 
         return kernel_dict
+    # fmt: on
```
leroyvn, line 822:
```suggestion
```

eradiate/scenes/measure.py
```diff
@@ -838,6 +876,8 @@ class RadianceMeterPlaneMeasure(Measure):
         Number of samples per (zenith, azimuth) pair.
         Default: 32.
     """
+
+    # fmt: off
```
leroyvn, line 880:
```suggestion
```

eradiate/scenes/measure.py
```diff
@@ -886,16 +926,13 @@ class RadianceMeterPlaneMeasure(Measure):
 
     _zenith_angles = attr.ib(default=None, init=False)  # Set during post-init
     _azimuth_angles = attr.ib(default=None, init=False)  # Set during post-init
+    # fmt: on
```
leroyvn, line 929:
```suggestion
```

eradiate/scenes/measure.py
```diff
@@ -982,6 +1029,7 @@ def sensor_info(self):
         else:
             return [(self.id, self.spp)]
 
+    # fmt: off
```
leroyvn, line 1032:
```suggestion
```

eradiate/scenes/measure.py
```diff
@@ -1013,3 +1061,4 @@ def kernel_dict(self, **kwargs):
             kernel_dict[sensor_id] = sensor_dict
 
         return kernel_dict
+    # fmt: on
```
leroyvn, line 1064:
```suggestion
```

leroyvn (changes requested):
Hi @schunkes, I made a couple of final comments and added change suggestions. Unless you disagree, I'll validate all that today and merge!

schunkes:
My idea for the black comments was, that these regions should actually never be formatted, because we chose the non compliant formatting here on purpose, to make the large dictionaries more readable and I would argue that it is nice to have all `attr.ib`s formatted in the same way. I feel it is easier to read them when you can train your eye on this sequence of `converter, validator, units` each on their own line.

leroyvn:
Fair point, my bad. Let's go with this, then!

eradiate/scenes/measure.py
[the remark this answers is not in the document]
leroyvn: I'll then just commit this suggestion out of consistency with regular floating point arithmetics practices. Let's keep our eyes open, we'll change it to strict equality if it proves to be necessary.

PR #26, opened 2021-02-02
Resolve "default max value for heterogeneous atmosphere automatic width"

nollety:
# Description

* Resolves eradiate/eradiate-issues#9

# Checklist

### DIFF
--- a/eradiate/scenes/atmosphere/heterogeneous.py
+++ b/eradiate/scenes/atmosphere/heterogeneous.py
@@ -256,10 +256,10 @@ class HeterogeneousAtmosphere(Atmosphere):
             min_sigma_s = sigma_s.min()
             if min_sigma_s > 0.:
                 width = 10. / min_sigma_s
-                if width > ureg.Quantity(1e5, "km"):
-                    width = ureg.Quantity(1e5, "km")
+                if width > ureg.Quantity(1e3, "km"):
+                    width = ureg.Quantity(1e3, "km")
             else:
-                raise ValueError("cannot automatically compute width when "
+                raise ValueError("cannot compute width automatically when "
                                  "scattering coefficient reaches zero")
         else:
             width = self.width


nollety:
Here is my preliminary findings:

The figure below shows the results of simulations performed with different atmosphere widths, from ten to a million km, in terms of the top-of-atmosphere outgoing spectral radiance at different viewing zenith angle (color code).

image

The edge effects are quantified by the deviations of the radiance values with respect to the radiance value for a width of a million of km, which is assumed to be free of edge effects and is taken as reference.

The number of samples per pixels is 1e5 in each simulations.

The results illustrates that the atmosphere width that eliminates edge effects depends on the viewing zenith angle (vza). For example, when `vza=2.`, the edge effects seem (visual inspection) to vanish around after a width of 200 or 300 km. When `vza=70°`, they disappear only when the width is around 1000 km and larger.

At 550 nm, it seems feasible to use a default atmosphere width value of **1000 km**. This value seems to garantee no edge effects for `vza` values smaller than **75°**.

Since scattering is greater at smaller wavelength values, I've reproduced the same experiment but at the wavelength of 400 nm instead of 550 nm. It's clear that the edge effects take "longer" (along the width axis) to disappear at 400 nm than at 550 nm, but overall the results are very similar to the 550 nm results.
 
image

Clearly, the current default value for the atmosphere width (`1e5`) is too large.
What do you think would be a good default value? @leroyvn

nollety:
Some more plots (relative differences):

rdiff_vs_width_lin_550 0

rdiff_vs_width_log_550 0

rdiff_vs_width_lin_400 0

rdiff_vs_width_log_400 0

For all `vza` <= 75°, the relative error is less than 1% when the atmosphere width is 1000 km.

Ygovs:
> Here is my preliminary findings:
> 
> The figure below shows the results of simulations performed with different atmosphere widths, from ten to a million km, in terms of the top-of-atmosphere outgoing spectral radiance at different viewing zenith angle (color code).
> 
> image
> 
> The edge effects are quantified by the deviations of the radiance values with respect to the radiance value for a width of a million of km, which is assumed to be free of edge effects and is taken as reference.
> 
> The number of samples per pixels is 1e5 in each simulations. The surface is a lambertian surface with a reflectance of 1. The illumination direction is the default direction (theta=0, phi=0). The measure is `toa_pplane`. The atmosphere is a 100 km high `us76_approx` heterogeneous atmosphere with 50 layers.
> 
> The results illustrates that the atmosphere width that eliminates edge effects depends on the viewing zenith angle (vza). For example, when `vza=2.`, the edge effects seem (visual inspection) to vanish around after a width of 200 or 300 km. When `vza=70°`, they disappear only when the width is around 1000 km and larger.
> 
> At 550 nm, it seems feasible to use a default atmosphere width value of **1000 km**. This value seems to garantee no edge effects for `vza` values smaller than **75°**.
> 
> Since scattering is greater at smaller wavelength values, I've reproduced the same experiment but at the wavelength of 400 nm instead of 550 nm. It's clear that the edge effects take "longer" (along the width axis) to disappear at 400 nm than at 550 nm, but overall the results are very similar to the 550 nm results.
> 
> image
> 
> Clearly, the current default value for the atmosphere width (`1e5 km`) is too large.
> What do you think would be a good default value? @leroyvn

I am not sure I understand the need to simulate atmosphere higher than 120km

nollety:
@Ygovs here this is about the atmosphere width, not the atmosphere height

leroyvn:
Hi @nollety, thanks for this PR. I'd have however preferred to have the discussion on atmosphere width and the irradiance dimension fix in 2 separate PRs to keep discussions separate (for next time, don't bother splitting it now).

Regarding atmosphere element width: 1e4 km is definitely safe, and you could even go for the threshold value associated with VZA = 80°. I even think that with further assessment of the validity range of the plane-parallel assumption, we could go below that value.

eradiate/util/xarray.py
```diff
@@ -517,45 +517,53 @@ def plane(hsphere_dataarray, phi, theta_dim="theta_o", phi_dim="phi_o", drop=Fal
...
-    # Reorder data
-    values_neg = values_neg.loc[{theta_dim: sorted(values_neg.coords[theta_dim].values)}]
-
-    # Combine negative and positive half-planes; drop the azimuth dimension
-    # (inserted as a scalar dimension afterwards)
-    result = xr.concat((values_neg, values_pos), dim=theta_dim).drop_vars(phi_dim)
+    theta_pos = hsphere.coords[theta_dim]
+    theta_neg = hsphere.coords[theta_dim][1:]
+    ds = xr.Dataset()
+    for var_name in ['lo', 'brf', 'brdf']:
```
leroyvn, line 526: You should not mention any field here: this component is supposed to handle any data, not just the data with the vars you're using here.

leroyvn (changes requested):
Regarding your proposal for the irradiance field fix: it's too intrusive IMO. I think it's safer to remove the plane slicing facilities from the dataset accessor and limit it to data arrays. That way, slicing must be explicitly specified for all variables in a dataset.

That being done, the actual change should be done in the solver app code, which will anyway be broken by the removal of this API.

nollety:
Hi @leroyvn, thanks for the feedback :)
I apologise for mixing two issues into one, next time I'll be careful to split thing up propertly!

I talked a bit with @Ygovs about the atmosphere width yesterday.
In the end, the optimal default width value is a matter of viewing zenith angle.
For simulations in plane-parallel atmosphere, viewing zenith angle larger than 75° will probably not be relevant because, as Yves explained to me yesterday, the plane-parallel approximation starts to be bad.
Considering that, and the fact that at `1e3` km, the results for `vza`=75° have already converged, we could go with `1e3` km. What do you think?
Given our current limitations to create vegetation scenes with large widths (from my understanding), and in the perspective of coupling atmosphere and vegetation scenes, I thought it was important to choose this value as small as possible. Of course, changing from `1e5` to `1e4` km is already a nice improvement ^^

leroyvn:
Works for me. Anyway, the value can then be set at solver app-level depending on measure setup.

nollety:
@leroyvn I reverted the commit attempting to fix the issue with the principal plane projection (and rebased); this is now a proper one-issue pull request! ^^

leroyvn:
Good, let's squash and merge this!

nollety:
@leroyvn this is ready to be merged ;)

leroyvn:
Let's go!

PR #29, opened 2021-02-11
Added boa_altitude attribute and bottom and top properties to atmosphere base class

nollety:
# Description
This is a suggestion.

## Changes
I've added a `boa_altitude` attribute and a `bottom` property and `top` property to the atmosphere base class. The atmosphere height is defined as `height = top - bottom`. The `boa_altitude` attribute behaves like the `toa_altitude` attribute.

## Motivation
These properties would help a lot in the handling of atmospheres that start at a non-zero altitudes, and also in the positioning of particles layers in the atmosphere.

For example, take an atmosphere that is defined by `boa_altitude = 2.0` and `toa_altitude = 120.0` (`height = 118.0`). Say that inside that atmosphere, you must place a particles layer with bottom and top altitudes of e.g. `3.0` and `5.0`. Given the `height` information only, you would place it at `3.0` vertical units above the bottom of the atmosphere -- assuming that the atmosphere starts at the altitude `0.0` -- where, in fact, it's only `1.0` unit above.

# Checklist

### DIFF
--- a/eradiate/scenes/atmosphere/base.py
+++ b/eradiate/scenes/atmosphere/base.py
@@ -44,8 +44,15 @@ class Atmosphere(SceneElement, ABC):
     )
 
     boa_altitude = attrib_quantity(
-        default=ureg.Quantity(0., "km"),
+        default="auto",
+        converter=converter_or_auto(converter_to_units(cdu.generator("length"))),
+        validator=validator_or_auto(
+            validator_units_compatible(ureg.m),
+            validator_is_positive
+        ),
         units_compatible=cdu.generator("length"),
+        units_add_converter=False,
+        units_add_validator=False
     )
 
     toa_altitude = attrib_quantity(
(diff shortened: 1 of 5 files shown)


leroyvn (commented):
Hi @nollety, that seems appropriate -- although we might have to think about a different system if such labelled locations start becoming more numerous. I'm however not too happy with the current way we check for all these; but that's a deeper design issue which I guess we'll address later.

For now, I just have one remark: you don't check for BOA/TOA altitude ordering for the homogeneous atmosphere class; any reason for that?

nollety:
Hi @leroyvn, I don't understand your remark about the BOA/TOA altitude ordering in the homogeneous atmosphere class. Here is the post-init sequence of the `Homogeneous` atmosphere class:
```python
    if self.bottom > self.top:
        raise ValueError(f"boa_altitude ({self.bottom}) must be "
                         f"smaller than toa_altitude ({self.top})")
```
It compares the `top` and `bottom` properties because `toa_altitude` and `boa_altitude` can be `"auto"` which is translated into a default value in the corresponding properties.

nollety:
@leroyvn after our discussion, the changes brought by this pull request could be useless so I've marked it as draft. I'll re-evaluate the need to have these kinds of attribute/properties for the atmosphere classes

nollety:
Similar properties are implemented by #106

PR #30, opened 2021-02-11
Update dataset attribute according to actual wavelength

nollety:
# Description

This is a quick fix for the issue with the `US76RadProfile.dataset` attribute being set a default value that is not appropriate outside a narrow wavelength range (~ 540 - 555 nm), and thus must be updated according to actual wavelength value.

# Checklist


leroyvn:
Closing (sloved by c1990f1).

PR #33, opened 2021-02-15
Refactored compute_sigma_a

nollety:
# Description

We have refactored `compute_sigma_a` so that it takes a dataset in input, and not a dataset id. `US76ApproxRadProfile` objects are now in charge of finding out what dataset to open and opening it.

The multiple datasets archive files on the ftp server are now put together into one single archive file [link](https://eradiate.eu/data/spectra-us76_u86_4.zip) (approx. 2.3 GB). This single archive file contains all 44 (wavenumber) * 64 (pressure) individual `us76_u86_4` datasets.

⚠️ Once/if this pull request is accepted, I'll remove the obsolete absorption datasets archive files.

# Checklist

### DIFF
--- a/docs/rst/getting_started/install.rst
+++ b/docs/rst/getting_started/install.rst
@@ -263,7 +263,7 @@ Downloading required data files
 
 Eradiate does not automatically ship all available data sets due to their size.
 In order to successfully run all tests and tutorials, the
-`us76_u86_4-4000_25711 data set <https://eradiate.eu/data/spectra-us76_u86_4-4000_25711.zip>`_
+`spectra-us76_u86_4 data set <https://eradiate.eu/data/spectra-us76_u86_4.zip>`_
 must be downloaded manually and placed in the ``resources/data`` directory.
 :ref:`This section <sec-user_guide-manual_download>` explains in detail where
 the data set can be found and where it must be placed exactly.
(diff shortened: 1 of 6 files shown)


eradiate/tests/system/test_sanity.py
```diff
@@ -144,4 +144,4 @@ def test_symmetry_zenith(variant_scalar_mono, surface, atmosphere):
 
     assert np.allclose(results_postprocessed["lo_zero"],
                        results_postprocessed["lo_pi"],
-                       rtol=5.e-3)
+                       rtol=1e-2)
```
leroyvn, line 147: Would mind making this a separate commit so that we can roll it back if it appears to be inappropriate?

leroyvn (commented):
Hi @nollety, this seems in line with what we discussed. I have a comment (see below) and a question: does your fix address the test execution time?

nollety:
Hi @leroyvn Thanks for the feedback, the `eradiate/radprops/tests` now takes less than 3s, compared to 20s before. 

I've just updated the installation instructions with the new link -- the section _Installation Guide - Downloading required data files_ looks to me like a repeat of _Data Guide - Adding manually downloaded data_ but I guess it's normal. 

The data guide did not require an update because the instructions to extract the archive into resources/data remain correct.

eradiate/tests/system/test_sanity.py
nollety: Oh sorry, it was left out. I did not mean to commit that. It's just that it's easier for me to see whether or not all tests pass and not having to filter out these failing sanity tests.

leroyvn:
Ok, so let's go with that! Please squash these into a single commit and we'll proceed with merging :)

nollety:
@leroyvn great, you can proceed with merging!

PR #32, opened 2021-02-12
Change tutorial engine: switch to sphinx-gallery

leroyvn:
# Description

This PR closes eradiate/eradiate-issues#45.

This PR ports tutorials to the [sphinx-gallery](https://sphinx-gallery.github.io) extension.

Todo:

- [x] Update RTD build to skip example execution (preview docs [here](https://eradiate.readthedocs.io/en/tutorial_engine/))
- [x] Sort examples nicely
- [x] Add download link or script for supplementary files
- [x] Document tutorials for users (how to run them)
- [x] Document tutorials for developers (how to create them, how to build the dev docs)

# Checklist

- [x] The code follows the relevant coding guidelines
- [x] The code generates no new warnings (technically it does, the new make target has a bug but it works)
- [x] The code is appropriately documented
- [x] The code is tested to prove its function
- [x] The feature branch is rebased on the current state of the `main` branch
- [x] I give permission that the Eradiate project may redistribute my contributions under the terms of its license

PR #34, opened 2021-02-16
Attribute docs

leroyvn:
# Description

This PR closes eradiate/eradiate-issues#30 and must be chained with #32. It implements a documentation system for attrs attributes and brings in the following:

- Attribute documentation is now attached to field declaration: it is easier to see when you have to update the documentation and the risk of changing a field and omitting docs updates is reduced.
- Attribute documentation is structured: there is a dedicated docs entry for the attribute description, its type and default values. I plan to add a documentation system for units as well.
- Attributes now carry their docs with them, meaning that they can be inherited. It's about time! This, in practice, means that all fields will always be displayed in the docs, regardless whether they are inherited or not.
- This docs system currently only supports strings. Pro: You can use restructuredText in them. Con: Automation could be improved. But the door is still open for more automation.

This update was long overdue and will make refactoring much easier, since now attrs attributes will carry their documentation with them. In particular, inherited fields can now be documented robustly and accurately. The system uses a flexible formatting mechanism, meaning that we won't go through too much pain if we decide to change our docstring format.

Todo:

- [x] Document this documentation system

# Checklist

- [x] ~~The code is tested to prove its function~~


leroyvn:
All squashed and rebased, let's merge!

PR #35, opened 2021-02-16
Mode refactoring

leroyvn:
# Description

This PR refactors the mode API and must be chained with #34. It moves the mode management internals to a private module. This add extra safety and compartmentalisation, and enables for a better controlled API definition (we can choose what to show and what to hide).

A price to pay for this is that I had to break the public API in the process: `eradiate.mode()` is now a function (I tried to maintain an instance-like syntax, but I couldn't figure out an elegant solution). I sorted out those details.

# Checklist


leroyvn:
All squashed and rebased, let's merge!

PR #38, opened 2021-02-18
Migrate to Pinttrs

leroyvn:
# Description

This PR closes eradiate/eradiate-issues#27 and eradiate/eradiate-issues#50. It migrates our code to the Pinttrs library together with deep changes to how our API and documentation are organised.

## Direct consequences of migration to Pinttrs

- The `eradiate.util.units` module is removed and replaced with a minimal `eradiate._units` private module, which only contains a few quantity and unit definitions. Members which should be exposed as part of the public API are forwarded appropriately.
- The `unit_enabled`decorator is gone. Therefore, it can no longer be used to add the `from_dict()` class constructor to classes. Fortunately, the `pinttr.interpret_units()` function makes the writing for such functions very simple.
- The `eradiate.util.exceptions.UnitsError` exception is now gone and replaced by `pinttr.exceptions.UnitsError`, which inherits from `pint.DimensionalityError`.

## Complete API reorganisation

While most API entries have been preserved, the whole module organisation has been rethought. The main idea is to remove the `util` subpackage, which has grown into a big bag of disparate items only related by the fact they didn't fit somewhere else. The scoping of some of the `util` modules was unclear and some of the code needed to be cleaned up or removed.

We make a distinction between private and public modules. Private modules, identified by a leading underscore contain internals not meant to be used outside of the scope of Eradiate itself: they are not part of the public API. Opposite are public modules, whose contents are meant to be completely exposed as part of Eradiate's API.

A middle ground exists between public and private modules: packages can be public and expose only a part of their contents as public API.

## Retired code

The `view` module has been retired.

## Documentation redesign

The API reference documentation, formerly generated using autosummary's recursion feature, has been completely rewritten. We still use autosummary, but every documented item has to be explicitly added to the documentation. Proper API entry classification is implemented by sectioning the API docs pages.

The new API organisation implies the exposure of private definitions using module imports (_e.g._ `eradiate.unit_context_config`, `eradiate.unit_context_kernel`, etc.). Documenting these requires some care; we might figure out a better way to both define and document these items (I have some metaclass work in the pipe).

## Follow-up

This PR will be followed by further refactorings and extensions:

- The `solvers` module hierarchy still requires deep refactoring and alignment.
- The xarray API will require some extension and changes to reflect the transition away from our so-called angular data arrays and data sets. In that process, a rewrite of the xarray and plotting tutorial will be necessary.

# Checklist


leroyvn:
We're getting close to release stage! I updated the documentation.

leroyvn:
Fire! Merging this.

PR #43, opened 2021-02-25
Solver module refactoring

leroyvn:
# Description

- Closes eradiate/eradiate-issues#40 
- Closes eradiate/eradiate-issues#55

This PR refactors the `solvers` sub-package. Changes are as follows:

- Moved all features up one level in the module hierarchy: 
  - `solvers.onedim.app` → `solvers.onedim`
  - `solvers.onedim.runner` → `solvers.core`
  - `solvers.onedim.postprocess` → `solvers.core`
  - `solvers.rami.app` → `solvers.rami`
- In practice, these new module are actually API entry points and the implementation of the various classes and functions still has deeper nesting.
- Added intermediate classes `Scene` and `SolverApp` as part of the new `solvers.core` module.
- Converted the `Runner` class to a lightweight function.
- Added support for `DistantMeasure` ray origin selection (there is still room for improvement but it can do for now).
- Refactored `OneDimSolverApp` to align it with `RamiSolverApp`.
- Updated API documentation.
- Update `OneDimSolverApp` tutorial.

Known issues:

- `OneDimSolverApp` measure proxies are gone and unlikely to return in the form they used to have. Ad-hoc factory-registered measures had very annoying issues and I'd rather substitute them with app-specific dictionary pre-processing using our beloved Cerberus. I can already hear you guys scream, but I don't want random convenience dict proxies tailored to a single use-case to invade the entire codebase. A good example of that is the request @Ygovs made to implicitly assume that atmospheres are `heterogeneous`; relevant for the 1D solver, but likely not in other cases.
- The SPP splitting feature is gone as well. It was made of two parts:
  - A measure-level component, which splits sensors in case of excess SPP
  - A solver-level component, which aggregates sensor results if a measure appears to have SPP split

  Both these components should be refactored as part of the improvement of the class hierarchy, and we currently don't really need SPP splitting. This feature will therefore be revived in a separate issue/PR.

A lot of the documentation is still to be updated, but it's a *lot* of work and it requires some discussion. I'll file a bunch of issues to keep track of what is to be done. In the meantime, we'll have to live to broken tutorials and outdated user docs.

Todo:

- [x] Double-check post-processing code for duplicates
- [x] Update examples
- [x] Update API docs
- [x] Update `OneDimSolver` user guide

# Checklist


leroyvn:
We're here! There is still a lot of work on docs (I'm sure there is still duplicate information and many pages are outdate now) but it's beyond the scope of this PR.

PR #48, opened 2021-03-10
Measure refactoring

leroyvn:
# Description

- Closes eradiate/eradiate-issues#59
- Closes eradiate/eradiate-issues#63

This PR refactors the measure module. Changes are as follows:

- The measure module is now split into a series of private submodules and has a clearly defined API.
- The `Measure` class was allocated more responsibility and notably implements the SPP splitting functionality, as well as:
  - sampler setup
  - film setup
  - kernel dictionary assembly (can be overridden)
- Child classes are expected to provide only the core elements of their kernel dicts 
- The radiancemeter measures were removed since they no longer serve any purpose; replacements will be implemented later
- The solver app sensor registry is gone, replaced by a structured `SensorInfo` data structure; we currently lose the search funcitonality, but it is no longer really useful and restoring it if it ever becomes necessary again will be very easy

To do:

- [x] Add a basic radiancemeter (thin interface to radiancemeter and radiancemeterarray)
- [x] Add SPP splitting integration test
- [x] Check if TinyDB is still required (maybe move to docs requirements?)

# Checklist


leroyvn:
Alright, all tests pass and the docs build. Let's merge this!

PR #50, opened 2021-03-12
Refactor KernelDict

leroyvn:
# Description

This PR refactors the `KernelDict` class. The code has been cleaned up and the interface has been slightly updated for improved consistency. Changes are as follows:

- `KernelDict` now inherits from `UserDict` (see [this post](https://treyhunner.com/2019/04/why-you-shouldnt-inherit-from-list-and-dict-in-python/#So_should_I_use_abstract_base_classes_or_UserDict_and_UserList?) for the motivation)
- The `empty()` class method has been replaced with a `new()` class method which also accepts initialisation content
- The `add()` method is now variadic and no longer accepts lists

# Checklist

### DIFF
--- a/eradiate/scenes/core.py
+++ b/eradiate/scenes/core.py
@@ -13,16 +13,21 @@ from .._units import unit_registry as ureg
 from ..exceptions import KernelVariantError
 
 
-class KernelDict(dict):
+class KernelDict(UserDict):
     """A dictionary designed to contain a scene specification appropriate for
     instantiation with :func:`~mitsuba.core.xml.load_dict`.
 
     :class:`KernelDict` keeps track of the variant it has been created with
     and performs minimal checks to help prevent inconsistent scene creation.
-    """
 
-    # TODO: add a content() method which returns a generator with the content of
-    #       the items() method without the 'type' entry
+    .. rubric:: Instance attributes
+
+    ``data`` (dict):
+        Wrapped dictionary.
+
+    ``variant`` (str):
+        Kernel variant for which the dictionary is created.
+    """
 
     def __init__(self, *args, **kwargs):
         """Initialise self and set :attr:`variant` attribute based on currently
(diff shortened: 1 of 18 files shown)


eradiate/scenes/core.py
```diff
@@ -76,43 +70,57 @@ def check(self):
...
-        else:
-            for key, value in content.items():
-                self[key] = value
-
-        return self
 
-    def normalize(self):
-        self["type"] = "scene"
+        for element in elements:
+            print(element)
```
schunkes, line 106: This should be removed.

eradiate/scenes/measure/_core.py
```diff
@@ -97,11 +98,10 @@ def postprocess_results(self, runner_results):
 
         return np.reshape(result, (result.shape[0], result.shape[1]))
 
-    def sensor_infos(self):
+    def sensor_infos(self) -> List[SensorInfo]:
```
schunkes, line 101: How do you use this return type hint? I mean, what do you do with it?

eradiate/solvers/core/_runner.py
```diff
@@ -40,7 +40,10 @@ def runner(kernel_dict):
     # Run computation
     from eradiate.kernel.core.xml import load_dict
 
-    kernel_scene = load_dict(kernel_dict)
+    try:
+        kernel_scene = load_dict(kernel_dict.data)
+    except AttributeError:
+        kernel_scene = load_dict(kernel_dict)
```
schunkes, line 46: Is this for backward compatibility? I think this should be removed in the final version then, right?

schunkes (commented):
Only have a handful of comments

eradiate/solvers/core/_runner.py
leroyvn: I did that to tolerate regular dicts, but it's indeed safer to just require a `KernelDict`. I'll change that.

eradiate/scenes/measure/_core.py
leroyvn: Ooops! I added this as I was experimenting with Sphinx and forgot to revert this file. Fixing that.

On a more general note: type hints are anyway purely optional, but we could use them for the following:

- Static type checking with mypy: this would allow us to remove some runtime type checking, which is always more pythonic.
- Strict runtime type checking with attrs: attrs fills automatically attributes' `type` field based on type hints. Combined with [attrs-strict](https://github.com/bloomberg/attrs-strict), this allows for boilerplate-free runtime type checking. Example:

  ```python
  import attr
  from attrs_strict import type_validator

  @attr.s
  class MyClass:
      foo: float = attr.ib(validator=type_validator())
  ```
  `foo` will be runtime type-checked with very little code. This, however, should be considered carefully: mypy should generally allow to avoid such things.
- Automated type information in docs: a Sphinx extension can fill in automatically the expected type in the docs. It's however incompatible with our current docstring format; but given how better our Sphinx theme has become, we can probably consider in the future switching to a standard docstring format. This would reduce the chance of desyncing code and docstrings.

nollety (approved):
Hi @leroyvn, I don't have any comments ; all these changes look fine to me :)

schunkes (approved):

leroyvn:
Thanks gentlemen, I'm merging this!

PR #53, opened 2021-03-24
presolver: Raise a user friendly ValueError when the ERADIATE_DIR env var is not set correctly

wint3ria:
# Description

The eradiate/_presolver.py used to fail with a cryptic error message when the ERADIATE_DIR variable was not set. This was an issue because a new user installing Eradiate is likely to forget to set this variable.

This simple patch add a user friendly error message with a correct ValueError in the case ERADIATE_DIR is not set.

# Checklist

### DIFF
--- a/eradiate/_presolver.py
+++ b/eradiate/_presolver.py
@@ -38,10 +38,30 @@ class PathResolver(metaclass=Singleton):
         ``[$PWD, $ERADIATE_DIR/resources/data]``
         """
 
+        eradiate_dir = os.getenv("ERADIATE_DIR")
+        if eradiate_dir is None:
+            raise ValueError(
+                "Environment variable ERADIATE_DIR is not set.\n"
+                "Please set this variable to the absolute path of the Eradiate installation folder.\n"
+                "If you are running Eradiate directly from the sources, " 
+                "then you can alternatively source the provided setpath.sh script."
+            )
+        
+        eradiate_path = Path(eradiate_dir).absolute()
+        eradiate_init = eradiate_path / "eradiate" / "__init__.py"
+        if not eradiate_init.exists():
+            raise ValueError(
+                f"Could not find {eradiate_init} file.\n"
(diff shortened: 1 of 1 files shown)


eradiate/_presolver.py
```diff
@@ -38,10 +38,17 @@ def reset(self):
         ``[$PWD, $ERADIATE_DIR/resources/data]``
         """
 
+        eradiate_dir = os.getenv("ERADIATE_DIR")
+        if eradiate_dir is None:
+            raise ValueError(f"Environment variable ERADIATE_DIR is not set.\n"
+                           + f"Please set this variable to the absolute path of the Eradiate installation folder.\n"
+                           + f"If you are running Eradiate directly from the sources, " + 
+                           "then you can alternatively source the provided setpath.sh script.")
```
leroyvn, line 46: No need for an f-string here; in addition, it's a good habit to avoid concatenating strings if it's just about breaking a long one during definition.

```suggestion
            raise ValueError(
                "Environment variable ERADIATE_DIR is not set.\n"
                "Please set this variable to the absolute path of the Eradiate installation folder.\n"
                "If you are running Eradiate directly from the sources, " 
                "then you can alternatively source the provided setpath.sh script."
            )
```

leroyvn (changes requested):
Hi @wint3ria, thanks for this PR, it's very useful indeed. I however suggest to check more thoroughly if `ERADIATE_DIR` is really the Eradiate source root directory. I guess detecting if the `$ERADIATE_DIR/eradiate/__init__.py` exists would do (or something similar).

eradiate/_presolver.py
wint3ria: Hum you're absolutely right, I will improve this format

I will also check the existence of this file

wint3ria: Done

leroyvn (approved):
Seems good to me, we can merge this!

PR #58, opened 2021-04-02
Remove alias to mitsuba

leroyvn:
# Description

This PR removes the Mitsuba alias in the kernel module. This feature, not really used anywhere, never worked exactly as intended. In addition, this proves to be problematic when it comes to adding kernel-related features: adding submodules or packages to the kernel module is cumbersome and requires unconventional logic.

This makes room for properly useful features, such as a logging wrapper or thin aliasing to plugins (to further help with scene design). Prior to proceeding, I'd like opinions.

To do:

- [x] Update kernel module docs

# Checklist


schunkes:
In what way did the alias prevent us from creating a logging wrapper or aliasing plugins? I always thought the `eradiate.kernel` alias was more about branding. The same way that you keep calling the the computational kernel in all meetings we have with the JRC and ESA :thinking:

schunkes (approved):
The code changes are pretty minor, so there's nothing to say. :)

leroyvn:
> In what way did the alias prevent us from creating a logging wrapper or aliasing plugins?

With the current system, any submodule import, *e.g.* `import eradiate.kernel.some_module`, is dispatched to `mitsuba.some_module` and results in an `ImportError` being raised: this can be quite problematic if we want to add kernel-specific functionality. I'll detail two of them below.

- Adding simplified entry points to Mitsuba's logging framework would naturally occur in a `eradiate.kernel.logging` module. See 88fe9532404ebdc31cd110c84504503c83e803ae.
- Adding a thin attrs-based interface to Mitsuba to make kernel-level work more streamlined when working with an IDE. I wanted to implement that in a `eradiate.kernel.pymts` subpackage (see 4d184939e50708c8d71154a9eb56bf34f823e33e and subsequent commits on the `pymts` branch). It's basically about generating dicts from objects in a fashion similar to what we do with the `KernelDict` workflow, but with simpler objects which basically define plugin parameters without any kind of processing or check. This would transfer the heavy lifting of dict creation to a kernel-specific component and would hopefully simplify the writing and debugging of `SceneElement` objects (if you write an incorrect plugin spec, your IDE will complain, and when we'll start using a static type checker, it will complain as well).

These are two examples of kernel-specific features which could be of use to us, but I can imagine more emerging in the future. While the branding point is still valid, there is, in practice, little point using the alias at all. I had to make very few changes to the end-user code when removing the alias: it's, in practice, only used by maintainers, who could be required to write 
```python
from eradiate.kernel import mitsuba  # obviously, import mitsuba is perfectly equivalent
mitsuba.set_variant("scalar_mono_double")
from mitsuba.core import foo
```
instead of
```python
import eradiate.kernel
eradiate.kernel.set_variant("scalar_mono_double")
from eradiate.kernel.core import foo
```

I think the added value of kernel-dedicated submodules outweighs module aliasing.

schunkes:
Ah, I understand :)

PR #63, opened 2021-05-04
Updated tutorial to use test version of solid_2017 data set

nollety:
# Description

I've updated the tutorial about solar irradiance spectrum data sets so that it does not require to download the full `solid_2017` data set (~80 MB) but instead uses a minimal version of that data set (~30 kB).

# Checklist


leroyvn:
Thanks @nollety, merging!

PR #65, opened 2021-05-05
Update us76_u86_4 absorption data set to new format

nollety:
# Description

Previously, an absorption data set consisted of a collection of NETCDF files -- one file per pressure-temperature pair -- that opened together into a single data set object -- using `xarray.open_mfdataset()` -- could be used to compute the absorption coefficient in the various pressure-temperature conditions that exist within the atmosphere. After testing, it turns out that combining these NETCDF files into one NETCDF file and opening the single file as a data set object -- now using `xarray.open_dataset()` -- result in lighter data sets and smaller opening times. It also reduce the total number of absorption data sets by a factor of ~100, thus ease the job of the absorption data getter.
This pull request is to apply this new format.

Changes:

* I have updated data guide and the `data.absorption_spectra` module to reflect this change of format for the absorption data sets.
* I have also uploaded a new archive file onto our ftp server with the new `us76_u86_4` absorption data sets. The archive file is now 60 MB (compared to 2.5 GB previously). That illustrates how lighter data sets are when combined. Each data set now corresponds to a 1000 cm^-1 wide wavenumber bin (16 to 500 nm wide in wavelength space) with a resolution of 0.01 cm^-1.
* I have updated the data guide, in particular the section about the `us76_u86_4` data sets.

If you have downloaded the `us76_u86_4` data sets, you will have to replace them by downloading the new ones ([here](eradiate.eu/data/us76_u86_4-spectra.zip)).

# Checklist


nollety:
Yes, but then not all absorption data sets will be part of the data submodule, only `us76_u86_4`. Also, I'm working on a higher resolution version of this data set -- the size of the archive file will likely increase by a factor of 10 which I assume will make the data set too big for the data submodule.

PR #71, opened 2021-05-14
Default heterogeneous atmosphere

nollety:
# Description

Resolves eradiate/eradiate-issues#67.

Added a default radiative properties profile so that the `HeterogeneousAtmosphere` default constructor does not require any argument.

The default radiative properties profile extends from 0 to 100 km and has 100 layers compared to the 50 indicated in eradiate/eradiate-issues#67. This is completely arbitrary and can be changed if needed.

# Checklist

### DIFF
--- a/eradiate/radprops/rad_profile.py
+++ b/eradiate/radprops/rad_profile.py
@@ -10,6 +10,7 @@ import pinttr
 import xarray as xr
 
 import eradiate
+from eradiate import path_resolver
 
 from .absorption import compute_sigma_a
 from .rayleigh import compute_sigma_s_air
@@ -29,8 +30,17 @@ from ..thermoprops.util import (
 from ..validators import all_positive
 
 
+def _to_quantity(variable):
+    try:
+        units = variable.units
+    except KeyError:
+        raise ValueError(f"{variable} has no units.")
+    finally:
+        return ureg.Quantity(variable.values, units)
+
+
 @ureg.wraps(
-    ret=None, args=("nm", "km", "km", "km^-1", "km^-1", "km^-1", None), strict=False
+    ret=None, args=("nm", "km", "km", "km^-1", "km^-1", "km^-1", ""), strict=False
 )
 def make_dataset(
     wavelength,
@@ -380,6 +390,15 @@ class ArrayRadProfile(RadProfile):
     def sigma_s(self, spectral_ctx=None):
         return self.sigma_t(spectral_ctx) * self.albedo(spectral_ctx)
 
+    @classmethod
+    def from_dataset(cls, path):
+        ds = xr.open_dataset(path_resolver.resolve(path))
+        z_level = _to_quantity(ds.z_level)
+        n_layers = ds.z_level.size - 1
+        albedo = _to_quantity(ds.albedo).reshape(1, 1, n_layers)
+        sigma_t = _to_quantity(ds.sigma_t).reshape(1, 1, n_layers)
+        return cls(albedo_values=albedo, sigma_t_values=sigma_t, levels=z_level)
+
     def to_dataset(self, spectral_ctx=None):
         if eradiate.mode().is_monochromatic():
             return make_dataset(
@@ -613,8 +632,8 @@ class US76ApproxRadProfile(RadProfile):
         profile = self.eval_thermoprops_profile()
         return make_dataset(
             wavelength=spectral_ctx.wavelength,
-            z_level=profile.z_level.values,
-            z_layer=profile.z_layer.values,
+            z_level=_to_quantity(profile.z_level),
+            z_layer=_to_quantity(profile.z_layer),
             sigma_a=self.sigma_a(spectral_ctx).flatten(),
             sigma_s=self.sigma_s(spectral_ctx).flatten(),
         )
@@ -994,8 +1013,8 @@ class AFGL1986RadProfile(RadProfile):
         """
         return make_dataset(
             wavelength=spectral_ctx.wavelength,
-            z_level=self.eval_thermoprops_profile().z_level.values,
-            z_layer=self.eval_thermoprops_profile().z_layer.values,
+            z_level=_to_quantity(self.eval_thermoprops_profile().z_level),
+            z_layer=_to_quantity(self.eval_thermoprops_profile().z_layer),
             sigma_a=self.sigma_a(spectral_ctx).flatten(),
             sigma_s=self.sigma_s(spectral_ctx).flatten(),
         )
--- a/eradiate/scenes/atmosphere/_heterogeneous.py
+++ b/eradiate/scenes/atmosphere/_heterogeneous.py
@@ -11,8 +11,8 @@ from ..._attrs import documented, parse_docs
 from ..._units import unit_context_kernel as uck
 from ..._units import unit_registry as ureg
 from ...radprops import RadProfileFactory
-from ...radprops.rad_profile import RadProfile
-from ...validators import is_file
+from ...radprops.rad_profile import AFGL1986RadProfile, RadProfile
+from ... import validators
 
 
 def write_binary_grid3d(filename, values):
@@ -108,14 +108,19 @@ class HeterogeneousAtmosphere(Atmosphere):
     This class builds a one-dimensional heterogeneous atmosphere. It expands as
     a ``heterogeneous`` kernel plugin, which takes as parameters a set of
     paths to volume data files. The radiative properties used to configure
-    :class:`.HeterogeneousAtmosphere` can be specified two ways:
+    :class:`.HeterogeneousAtmosphere` can be specified in two ways:
 
     - if the ``profile`` field is specified, kernel volume data files will be
       created using those data;
-    - if the ``profile`` field is not specified (*i.e.* set to ``None``), kernel
+    - if the ``profile`` field is set to ``None`` and
+      the ``albedo_fname`` and ``sigma_t_fname`` fields are specified, kernel
       volume data files will be read from locations set in the ``albedo_fname``
       and ``sigma_t_fname`` attributes (which then must be set to paths
       pointing to existing files).
+    - if the ``profile`` field is not specified and neither are the
+      ``albedo_fname`` and ``sigma_t_fname`` fields, then ``profile`` is set to
+      the default :class:`~eradiate.radprops.rad_profile.AFGL1986RadProfile`
+      radiative properties profile.
 
     If ``profile`` is specified:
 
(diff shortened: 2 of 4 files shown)


eradiate/xarray/util.py
```diff
@@ -0,0 +1,9 @@
+from eradiate import unit_registry as ureg
+
+
+def to_quantity(variable):
+    try:
+        units = variable.units
+    except KeyError:
+        raise ValueError(f"{variable} has no units.")
+    return ureg.Quantity(variable.values, units)
```
leroyvn, line 9: Stuffing many specific helper functions without well-though API integration is opening the door to lots of dead code in the long term. Since this function is tiny and you seem to use it in a single module, I'd rather have it there as a local `_to_quantity()` function.

eradiate/radprops/rad_profile.py
```diff
@@ -613,8 +734,8 @@ def to_dataset(self, spectral_ctx=None):
         profile = self.eval_thermoprops_profile()
         return make_dataset(
             wavelength=spectral_ctx.wavelength,
-            z_level=profile.z_level.values,
-            z_layer=profile.z_layer.values,
+            z_level=to_quantity(profile.z_level),
+            z_layer=to_quantity(profile.z_layer),
```
leroyvn, line 738: I think you're missing a conversion to unitless here (this should generate a warning).

eradiate/radprops/rad_profile.py
```diff
@@ -393,6 +394,126 @@ def to_dataset(self, spectral_ctx=None):
...
+            2.93870330e-08,
+            2.46196683e-08,
+            2.06121737e-08,
+            1.72454030e-08,
+            1.44203783e-08,
+            1.20527715e-08,
+            1.00710400e-08,
+            8.41472640e-09,
+            7.03205639e-09,
+            5.87861565e-09,
```
leroyvn, line 504: Not sure if hard-coding this in the Python module is best; maybe it should go to a CSV or NetCDF file in the data submodule?

eradiate/scenes/atmosphere/_heterogeneous.py
```diff
@@ -144,54 +149,36 @@ class HeterogeneousAtmosphere(Atmosphere):
         ),
         doc="Radiative property profile used. If set, volume data files will be "
         "created from profile data to initialise the corresponding kernel "
-        "plugin. If ``None``, :class:`.HeterogeneousAtmosphere` will assume "
-        "that volume data files already exist.",
+        "plugin.",
```
leroyvn, line 152: What happens now if `None` is used?

eradiate/scenes/atmosphere/_heterogeneous.py
```diff
@@ -203,19 +190,30 @@ def _sigma_t_fname_validator(self, attribute, value):
     _quantities = {"albedo": "albedo", "sigma_t": "collision_coefficient"}
 
     def __attrs_post_init__(self):
-        # Check for automatic width and height computation
+        # Check attributes coherence
```
leroyvn, line 193:
```suggestion
        # Check attributes consistency
```

leroyvn (changes requested):
Hi @nollety, I have a few comments. Since the default profile is arbitrary, you might want to leave the option of modifying it without a code update. As a general rule, hard-coded should be limited to physical constants IMO.

eradiate/radprops/rad_profile.py
nollety: I am not sure if I understand.
The `z_level` and `z_layer` coordinates in the thermophysical `profile` object will have length units. I tried again, I don't see a warning.

eradiate/xarray/util.py
nollety: Alright!

eradiate/radprops/rad_profile.py
leroyvn: If `make_dataset()` interprets units, then forget what I wrote. I just raise that point because `profile.z_level.values` is unitless while `to_quantity(profile.z_level)` is not.

eradiate/radprops/rad_profile.py
nollety: Using a NetCDF file seems more appropriate indeed.

eradiate/scenes/atmosphere/_heterogeneous.py
nollety: When `profile` is `None`, two things can happen:
* if `sigma_t_fname` and `albedo_fname` are not `None` (`toa_altitude` and `width` must also be set in that case), the volume data files are used to define the atmospheric medium's collision coefficients.
* if `sigma_t_fname` and `albedo_fname` are also `None`, then the default radiative properties profile -- defined by `radprops.rad_profile._DEFAULT_RAD_PROFILE` -- is used to define the atmospheric medium's collision coefficients.

eradiate/radprops/rad_profile.py
nollety: Ah, you were right that something is wrong; I found a mistake in the decorator of `radprops.rad_profile.make_dataset()`. I think everything is fine now.

eradiate/scenes/atmosphere/_heterogeneous.py
nollety: I think all that is becoming quite complex; the `HeterogeneousAtmosphere` has a very long `__attrs_post_init__` with multiple and nested if-else blocks.

leroyvn: Indeed. If it is impractical, then we must find a way to make it simpler. I'll try to figure out how. 

Still, the behaviour should be documented, unless this case is so unlikely to happen that it's not worth mentioning it.

leroyvn: After careful examination, here is my suggestion.

- Drop the `_DEFAULT_RAD_PROFILE` dataset and replace it with a factory default for `profile`.
- Restore the previous behaviour of the `HeterogeneousAtmosphere` init sequence. This implies that passing volume data files manually requires that the user explicitly sets `profile` to None, which is fine IMO.

I think it is simpler. A minor docs update will be required as well.

eradiate/scenes/atmosphere/_heterogeneous.py
```diff
@@ -11,7 +11,7 @@
 from ..._units import unit_context_kernel as uck
 from ..._units import unit_registry as ureg
 from ...radprops import RadProfileFactory
-from ...radprops.rad_profile import RadProfile
+from ...radprops.rad_profile import RadProfile, _DEFAULT_RAD_PROFILE
 from ...validators import is_file
```
leroyvn, line 15:
```suggestion
from ... import validators
```

eradiate/scenes/atmosphere/_heterogeneous.py
```diff
@@ -144,54 +149,36 @@ class HeterogeneousAtmosphere(Atmosphere):
...
         type=":class:`~eradiate.radprops.rad_profile.RadProfile` or None",
         default="None",
     )
 
     albedo_fname = documented(
-        attr.ib(default=None, converter=attr.converters.optional(Path)),
+        attr.ib(
+            default=None,
+            converter=attr.converters.optional(Path),
+            validator=attr.validators.optional(is_file),
```
leroyvn, line 161:
```suggestion
            validator=attr.validators.optional(validators.is_file),
```

eradiate/scenes/atmosphere/_heterogeneous.py
```diff
@@ -144,54 +149,36 @@ class HeterogeneousAtmosphere(Atmosphere):
...
-            try:
-                return is_file(self, attribute, value)
-            except FileNotFoundError:
-                raise
-
     sigma_t_fname = documented(
         attr.ib(
             default=None,
             converter=attr.converters.optional(Path),
+            validator=attr.validators.optional(is_file),
```
leroyvn, line 174:
```suggestion
            validator=attr.validators.optional(validators.is_file),
```

leroyvn (changes requested):
Hi @nollety, more comments below.

eradiate/scenes/atmosphere/_heterogeneous.py
```diff
@@ -138,60 +143,42 @@ class HeterogeneousAtmosphere(Atmosphere):
...
             validator=attr.validators.optional(attr.validators.instance_of(RadProfile)),
         ),
         doc="Radiative property profile used. If set, volume data files will be "
         "created from profile data to initialise the corresponding kernel "
-        "plugin. If ``None``, :class:`.HeterogeneousAtmosphere` will assume "
-        "that volume data files already exist.",
+        "plugin.",
         type=":class:`~eradiate.radprops.rad_profile.RadProfile` or None",
-        default="None",
+        default="AFGL1986RadProfile()",
```
leroyvn, line 154: This should generate a hyperlink:
```suggestion
        default=":class:`AFGL1986RadProfile() <.AFGL1986RadProfile>`",
```

eradiate/scenes/atmosphere/_heterogeneous.py
```diff
@@ -203,19 +190,29 @@ def _sigma_t_fname_validator(self, attribute, value):
...
+                        "toa_altitude cannot be set to 'auto' when 'albedo_fname' and 'sigma_t_fname' are not None"
+                    )
+                if self.width == "auto":
+                    raise ValueError(
+                        "width cannot be set to 'auto' when 'albedo_fname' and 'sigma_t_fname' are None"
+                    )
+            else:
                 raise ValueError(
-                    "toa_altitude cannot be set to 'auto' when profile is None"
+                    "both 'albedo_fname' and 'sigma_t_fname' must be set when 'profile' is not set."
```
leroyvn, line 214:
```suggestion
                    "both 'albedo_fname' and 'sigma_t_fname' must be set when 'profile' is None"
```

eradiate/scenes/atmosphere/_heterogeneous.py
```diff
@@ -203,19 +190,29 @@ def _sigma_t_fname_validator(self, attribute, value):
     _quantities = {"albedo": "albedo", "sigma_t": "collision_coefficient"}
 
     def __attrs_post_init__(self):
-        # Check for automatic width and height computation
+        # Check attributes consistency
```
leroyvn, line 193: The only write operations I see are at lines 201-202: if removing them makes sense (which I think is fine, since "auto" is the default value for these fields), then all this code should be transferred to validators.

eradiate/scenes/atmosphere/_heterogeneous.py
```diff
@@ -203,19 +190,29 @@ def _sigma_t_fname_validator(self, attribute, value):
...
         if self.profile is not None:
             if self.toa_altitude != "auto":
                 raise ValueError(
                     "toa_altitude must be set to 'auto' when profile is set"
                 )
         else:
-            if self.toa_altitude == "auto":
+            if self.albedo_fname is None and self.sigma_t_fname is None:
+                self.toa_altitude = "auto"
+                self.width = "auto"
```
leroyvn, line 202: Danger zone: This basically means that a user explicitly setting `toa_altitude` or `width` inconsistently will *not* get an error and won't get the object initialised as they expect. I think this should be removed.

eradiate/radprops/rad_profile.py
```diff
@@ -29,8 +30,17 @@
 from ..validators import all_positive
 
 
+def _to_quantity(variable):
+    try:
+        units = variable.units
```
leroyvn, line 35: I didn't know one could reference metadata as regular attributes, that's nice!

leroyvn (changes requested):
Hi @nollety, we're getting there. The `HeterogeneousAtmosphere` init sequence still needs some attention, I think you can completely get ride of the post-init step.

eradiate/radprops/rad_profile.py
nollety: Indeed, I like that a lot :)

eradiate/scenes/atmosphere/_heterogeneous.py
nollety: I agree

eradiate/scenes/atmosphere/_heterogeneous.py
nollety: I completely agree

eradiate/scenes/atmosphere/_heterogeneous.py
leroyvn: By the way: I am now sure that validators can be shared, so you can attach a single validator to all fields it is supposed to validate (this can include coupled validation). See `ArrayRadProfile` for an example.

nollety: Thanks for the tip! :)

eradiate/scenes/atmosphere/_heterogeneous.py
```diff
@@ -203,20 +215,6 @@ def _sigma_t_fname_validator(self, attribute, value):
...
-                )
-        else:
-            if self.toa_altitude == "auto":
-                raise ValueError(
-                    "toa_altitude cannot be set to 'auto' when profile is None"
-                )
-            if self.width == "auto":
-                raise ValueError("width cannot be set to 'auto' when profile is None")
-
         # Prepare cache directory in case we'd need it
```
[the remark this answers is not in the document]
leroyvn: Sorry, I missed that part. It's okay :)

leroyvn (approved):
Thanks @nollety, let's merge this!

PR #82, opened 2021-05-25
Explicit "auto" value for surface width

leroyvn:
# Description

This PR closes eradiate/eradiate-issues#83. It changes the way the surface's kernel object size is computed and allows for explicit setting of the `width` field to `"auto"`. Changes are as follows:

- `KernelDictContext`: `atmosphere_kernel_width` is renamed to a more general `override_surface_width`
- Added a `OverriddenValueWarning` used to notify the user that their configuration is being ignored
- `Surface.kernel_width()` now uses `override_surface_width` to force a kernel object value; it also issues warnings if relevant
- Added `Surface` unit tests
- Updated solvers to use surface face override infrastructure properly

# Checklist

### DIFF
--- a/eradiate/contexts.py
+++ b/eradiate/contexts.py
@@ -160,10 +160,12 @@ class KernelDictContext:
         default="True",
     )
 
-    atmosphere_kernel_width: Optional[pint.Quantity] = documented(
+    override_surface_width: Optional[pint.Quantity] = documented(
         pinttr.ib(default=None, units=ucc.deferred("length")),
-        doc="If relevant, stores the width of the kernel object associated with "
-        "the atmosphere.",
+        doc="If relevant, value which must be used as the surface width "
+        "(*e.g.* when surface size must match atmosphere or canopy size).\n"
+        "\n"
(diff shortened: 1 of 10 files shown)


leroyvn:
@nollety opinions? Does this solve the odd case you reported in eradiate/eradiate-issues#83?

nollety:
@leroyvn That looks great to me!

Is there a way -- I assume not -- to set `surface.width` to match `atmosphere.width`? Otherwise, the user has to trust eradiate that, during the translation to kernel objects, the surface width does indeed match the atmosphere width, because printing the eradiate scene object will only show that the atmosphere width is `1000 km` and the surface width is `"auto"`.

leroyvn:
No, there is not. You can set it explicitly, but `atmosphere.width` will always override `surface.width` upon call to `surface.kernel_width()` during `OneDimSolver.process()` execution. So basically, unless you have no atmosphere or canopy, there is no point setting `surface.width` to anything but `"auto"` .

leroyvn:
I'll merge this now. Thanks for your review @nollety!

PR #86, opened 2021-06-02
Add extra canopy abstractions to represent trees

schunkes:
# Description

Some abstract canopies in the RAMI framework are formed by leaf clouds plus trunks, this PR introduces a new abstraction that lets users combine the two and instantiate them with the currently existing machinery for discrete canopies.

I have added a class `Plant` that combines a `LeafCloud` with a cylindrical trunk and I refactored the `InstancedLeafCloud` into `InstancedPlant` to use the new class.

# Checklist

### DIFF
--- a/docs/rst/reference/scenes.rst
+++ b/docs/rst/reference/scenes.rst
@@ -56,8 +56,10 @@ Biosphere [eradiate.scenes.biosphere]
 .. autosummary::
    :toctree: generated/
 
+   CanopyElement
    LeafCloud
-   InstancedLeafCloud
+   AbstractTree
+   InstancedCanopyElement
    DiscreteCanopy
 
 .. dropdown:: **Parameters for LeafCloud generators**
@@ -68,6 +70,7 @@ Biosphere [eradiate.scenes.biosphere]
       _discrete.CuboidLeafCloudParams
       _discrete.SphereLeafCloudParams
       _discrete.CylinderLeafCloudParams
+      _dicsrete.ConeLeafCloudParams
 
 .. _sec-reference-scenes-surface:
 
--- a/eradiate/scenes/biosphere/__init__.py
+++ b/eradiate/scenes/biosphere/__init__.py
@@ -1,10 +1,18 @@
 from ._core import BiosphereFactory, Canopy
-from ._discrete import DiscreteCanopy, InstancedLeafCloud, LeafCloud
+from ._discrete import (
+    AbstractTree,
+    CanopyElement,
+    DiscreteCanopy,
+    InstancedCanopyElement,
+    LeafCloud,
+)
 
 __all__ = [
     "BiosphereFactory",
     "Canopy",
     "DiscreteCanopy",
-    "InstancedLeafCloud",
+    "InstancedCanopyElement",
     "LeafCloud",
+    "AbstractTree",
+    "CanopyElement",
 ]
--- a/eradiate/scenes/biosphere/_discrete.py
+++ b/eradiate/scenes/biosphere/_discrete.py
@@ -14,6 +14,7 @@ from ..core import SceneElement
 from ..spectra import Spectrum, SpectrumFactory
 from ... import validators
 from ..._attrs import documented, get_doc, parse_docs
+from ..._factory import BaseFactory
 from ...contexts import KernelDictContext
 from ...units import unit_context_config as ucc
 from ...units import unit_context_kernel as uck
@@ -55,7 +56,9 @@ def _leaf_cloud_positions_cuboid_avoid_overlap(
 ):
     """Compute leaf positions for a cuboid-shaped leaf cloud (square footprint).
     This function also performs conservative collision checks to avoid leaf
-    overlapping.
+    overlapping. This process might take a very long time, if the parameters
+    specify a very dense leaf cloud. Consider using
+    :func:`_leaf_cloud_positions_cuboid`.
     """
     n_attempts = int(n_attempts)  # For safety, ensure conversion to int
 
@@ -131,6 +134,25 @@ def _leaf_cloud_positions_cylinder(n_leaves, radius, l_vertical, rng):
         r = rand[1] * radius
         z = rand[2] * l_vertical
         positions[i, :] = [r * np.cos(phi), r * np.sin(phi), z]
+
+    return positions
+
+
+@ureg.wraps(ureg.m, (None, ureg.m, ureg.m, None))
+def _leaf_cloud_positions_cone(n_leaves, radius, l_vertical, rng):
+    """Compute leaf positions for a cone-shaped leaf cloud (vertical
+    orientation, tip pointing towards positive z.)"""
+
+    positions = np.empty((n_leaves, 3))
+
+    # uniform cone sampling from here:
+    # https://stackoverflow.com/questions/41749411/uniform-sampling-by-volume-within-a-cone
+    for i in range(n_leaves):
+        rand = rng.random(3)
+        h = l_vertical * (rand[0] ** (1 / 3))
+        r = radius / l_vertical * h * np.sqrt(rand[1])
+        phi = rand[2] * 2 * np.pi
+        positions[i, :] = [r * np.cos(phi), r * np.sin(phi), l_vertical - h]
 
     return positions
 
@@ -443,7 +465,72 @@ class CylinderLeafCloudParams(LeafCloudParams):
 
 @parse_docs
 @attr.s
-class LeafCloud(SceneElement):
+class ConicalLeafCloudParams(LeafCloudParams):
+    """
+    Advanced parameter checking class for the cone :class:`.LeafCloud`
+    generator.
+
+    .. seealso:: :meth:`.LeafCloud.cone`
+    """
+
+    _radius = documented(
+        pinttr.ib(default=1.0 * ureg.m, units=ucc.deferred("length")),
+        doc="Leaf cloud radius.\n\nUnit-enabled field (default: ucc[length]).",
+        type="float",
+        default="1 m",
+    )
+
+    _l_vertical = documented(
+        pinttr.ib(default=1.0 * ureg.m, units=ucc.deferred("length")),
+        doc="Leaf cloud vertical extent.\n\nUnit-enabled field (default: ucc[length]).",
+        type="float",
+        default="1 m",
+    )
+
+    @property
+    def radius(self):
+        return self._radius
+
+    @property
+    def l_vertical(self):
+        return self._l_vertical
+
+
+@parse_docs
+@attr.s
+class CanopyElement(SceneElement):
+    """
+    Abstract base class for objects that can be instantiated in an
+    :class:`.InstancedCanopyElement`.
+    """
+
+    def kernel_dict(self, ctx: Optional[KernelDictContext] = None) -> MutableMapping:
+        if not ctx.ref:
+            return self.shapes(ctx=ctx)
+        else:
+            return {**self.bsdfs(ctx=ctx), **self.shapes(ctx=ctx)}
+
+
+class CanopyElementFactory(BaseFactory):
+    """
+    This factory constructs objects whose classes are derived from
+    :class:`.CanopyElement`.
+
+    .. admonition:: Registered factory members
+       :class: hint
+
+       .. factorytable::
+          :factory: CanopyElementFactory
+    """
+
+    _constructed_type = CanopyElement
+    registry = {}
+
+
+@CanopyElementFactory.register("leaf_cloud")
+@parse_docs
+@attr.s
+class LeafCloud(CanopyElement):
     """
     A container class for leaf clouds in abstract discrete canopies.
     Holds parameters completely characterising the leaf cloud's leaves.
@@ -456,6 +543,7 @@ class LeafCloud(SceneElement):
       * :meth:`.LeafCloud.cuboid`;
       * :meth:`.LeafCloud.sphere`;
       * :meth:`.LeafCloud.cylinder`;
+      * :meth:`.LeafCloud.cone`;
 
     * :meth:`.LeafCloud.from_file` loads leaf positions and orientations from a
       text file;
@@ -589,7 +677,7 @@ class LeafCloud(SceneElement):
 
         .. seealso:: :class:`.CuboidLeafCloudParams`
 
-        The produced leaf cloud covers uniformly the
+        The produced leaf cloud uniformly covers the
         :math:`(x, y, z) \\in [-\\dfrac{l_h}{2}, + \\dfrac{l_h}{2}] \\times [-\\dfrac{l_h}{2}, + \\dfrac{l_h}{2}] \\times [0, l_v]`
         region. Leaf orientation is controlled by the ``mu`` and ``nu`` parameters
         of an approximated inverse beta distribution
@@ -664,18 +752,17 @@ class LeafCloud(SceneElement):
         """
         rng = np.random.default_rng(seed=seed)
         params = SphereLeafCloudParams(**kwargs)
-        leaf_positions = (
-            _leaf_cloud_positions_sphere(params.radius, params.n_leaves) * ureg.m,
-            rng,
+        leaf_positions = _leaf_cloud_positions_sphere(
+            params.radius, params.n_leaves, rng
         )
         leaf_orientations = _leaf_cloud_orientations(
             params.n_leaves, params.mu, params.nu, rng
         )
-        leaf_radii = _leaf_cloud_radii(params.n_leaves, params.leaf_radius) * ureg.m
+        leaf_radii = _leaf_cloud_radii(params.n_leaves, params.leaf_radius)
 
         # Create leaf cloud object
         return cls(
-            id=id,
+            id=params.id,
             leaf_positions=leaf_positions,
             leaf_orientations=leaf_orientations,
             leaf_radii=leaf_radii,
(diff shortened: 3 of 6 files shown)


schunkes:
I did not yet add or update any tests. First I'd like some feedback on my implementation.

eradiate/scenes/biosphere/_discrete.py
```diff
@@ -151,7 +170,7 @@ def _leaf_cloud_orientations(n_leaves, mu, nu, rng):
     return orientations
 
 
-@ureg.wraps(ureg.m, (None, ureg.m))
+@ureg.wraps(ureg.m, (None, ureg.m), strict=False)
```
leroyvn, line 173: Please keep `strict` to `True` and wrap arguments into appropriate units in calling code (see `test_leaf_cloud_radii()` for the corresponding unit test). ` strict=False` allows to pass unitless quantities and might result in hard-to-predict behaviour.

eradiate/scenes/biosphere/_discrete.py
```diff
@@ -202,6 +221,14 @@ class LeafCloudParams:
         type="float",
     )
 
+    position_offset = documented(
+        pinttr.ib(default=0.0 * ureg.m, units=ucc.deferred("length")),
+        doc="z-offset for leaf positions. Used for trees with trunks.\n\n"
+            "Unit-enabled field (default: ucc[length]).",
+        type="float",
+        default="0.0 m",
+    )
```
leroyvn, line 230: The overall idea is that those leaf clouds are generated in a reference frame and then positioned by the user. I think you can do without this field: just apply the shift after generating the cloud.

eradiate/scenes/biosphere/_discrete.py
```diff
@@ -511,6 +572,14 @@ class LeafCloud(SceneElement):
...
+
```
leroyvn, line 582: See above.

eradiate/scenes/biosphere/_discrete.py
```diff
@@ -867,9 +982,11 @@ def shapes(self, ctx=None):
         else:
             bsdf = self.bsdfs(ctx=ctx)[f"bsdf_{self.id}"]
 
+        offset_positions = [position + [0, 0, self.position_offset.m_as(kernel_length)] for position in self.leaf_positions.m_as(kernel_length)]
```
leroyvn, line 985: Black pass please 🙏

leroyvn (changes requested):
Hi @schunkes, thanks for tackling this issue. In addition to specific comments below, I'll summarise the discussion we had this morning.

Instead of chaging `LeafCloud` into something else, let's just keep it and introduce additional elements we'll need to account for trees with trunks. We can introduce a `CanopyElement` class, which will replace `LeafCloud` as the basic instantiable element. We then turn `InstancedLeafCloud` into `InstancedCanopyElement`. Our `LeafCloud` inherit from `CanopyElement`, and we add a `AbstractTree` which has a `LeafCloud` and trunk parameters.

Please find below some code summarising this:

```python
class DiscreteCanopy(Canopy):
    instanced_canopy_elements: List[InstancedCanopyElement] = attr.ib(...)
    ...

class InstancedCanopyElement(SceneElement):
    canopy_element: CanopyElement = attr.ib(...)
    instance_positions: ArrayLike = attr.ib(...)
    ...

class CanopyElement(SceneElement, ABC):
    ...

class LeafCloud(CanopyElement):
   ...

class AbstractTree(CanopyElement):
    leaf_cloud: LeafCloud = attr.ib(...)
    trunk_height = attr.ib(...)
    ...
```

schunkes:
Since I was concerned with theAPI getting to complicated, how about this:

Current factory call:

```python
Factory.create({
    "type": "discrete_canopy",
    "instanced_leaf_clouds": [
        {
            "instance_positions": [...],
            "leaf_cloud": {<params>}
        },
    ]
```

New factoy call:


```python
Factory.create({
    "type": "discrete_canopy",
    "canopy_elements": [
        {
            "type": "abstract_tree",
            "instance_positions": [...],
            "leaf_cloud": {<params>},
            "trunk_height": ...,
            "trunk_radius": ...,
            "trunk_reflectance": ...
        },
    ]
```

Notable changes:

- Drop the `instanced` prefix in the `DiscreteCanopy`'s parameter. The user does not really have to care if the canopy elements are instanced or not.
- Introduce a `type` to the `CanopyElement` call, eliminating an extra layer of abstraction. Otherwise the call would look like this:

```python
Factory.create({
    "type": "discrete_canopy",
    "instanced_canopy_elements": [
        {
            "instance_positions": [...],
            "canopy_element": {
                "type": "abstract_tree",
                 "leaf_cloud": {<params>},
                "trunk_height": ...,
                "trunk_radius": ...,
                "trunk_reflectance": ...
            }
        },
    ]
```

I think my proposal keeps all options for further classes derived from `CanopyElement` open and is easier to read and understand.

leroyvn:
This significantly obfuscates the underlying object model: inferring dict syntax from the corresponding function calls becomes impossible. My main concern in such cases is, as always, documentation. In order to see how practical your proposal can be, I suggest to:

1. Implement a stupid and simple `from_dict()` (the "ugly" one, shouldn't take much effort).
2. Implement your proposal as a `from_dict_brief()`, including documentation (docstrings and tutorial).
3. Compare both in terms of implementation and documentation, keep the best.

eradiate/scenes/biosphere/_discrete.py
schunkes: All done! ✨ 🍰 ✨

schunkes:
Okidoki, take another look :)

eradiate/scenes/biosphere/_discrete.py
```diff
@@ -14,6 +14,7 @@
 from ... import validators
 from ..._attrs import documented, get_doc, parse_docs
 from ...contexts import KernelDictContext
+from ..._factory import BaseFactory
```
leroyvn, line 17: isort pass please 🙏

eradiate/scenes/biosphere/_discrete.py
```diff
@@ -502,7 +495,34 @@ def l_vertical(self):
...
+        if not ctx.ref:
+            return self.shapes(ctx=ctx)
+        else:
+            return {**self.bsdfs(ctx=ctx), **self.shapes(ctx=ctx)}
+
+
+class CanopyElementFactory(BaseFactory):
+    """
+    This factory constructs objects whose classes are derived from
+    :class:`.Canopy`.
```
leroyvn, line 509: Docstring update please 🙏

eradiate/scenes/biosphere/_discrete.py
```diff
@@ -1125,6 +1117,11 @@ def shapes(self, ctx=None):
         kernel_height = self.trunk_height.m_as(kernel_length)
         kernel_radius = self.trunk_radius.m_as(kernel_length)
 
+        # offset the leaf cloud by the size of the trunk
+        self.leaf_cloud.leaf_positions += (
+            np.array([0, 0, kernel_height]).reshape((1, 3)) * kernel_length
+        )
```
leroyvn, line 1123: Danger-danger: you're mutating the object and you shouldn't.
```suggestion
        leaf_positions = self.leaf_cloud.leaf_positions + np.array([0, 0, kernel_height]).reshape((1, 3)) * kernel_length
```

eradiate/scenes/biosphere/_discrete.py
```diff
@@ -1178,12 +1175,11 @@ def from_dict(cls, d):
...
-        if d_copy["trunk_height"] > 0.0 * ureg.m:
-            d_copy["leaf_cloud"]["position_offset"] = d_copy["trunk_height"]
+        leaf_cloud_dict = d_copy.pop("leaf_cloud")
+
+        leaf_cloud = LeafCloud.convert(leaf_cloud_dict)
 
-        return cls(**d_copy)
+        return cls(leaf_cloud=leaf_cloud, **d_copy)
 
     def kernel_dict(self, ctx: Optional[KernelDictContext] = None) -> MutableMapping:
```
leroyvn, line 1184: If I'm not mistaken, `CanopyElement` provides this implementation already.

eradiate/scenes/biosphere/_discrete.py
```diff
@@ -1206,15 +1202,17 @@ class InstancedPlant(SceneElement):
...
+                attr.validators.instance_of(CanopyElement)
+            ),
+            converter=CanopyElementFactory.convert,
         ),
-        doc="Instanced plant. Can be specified as a dictionary, which will "
-        "be interpreted by :meth:`.Plant.from_dict`.",
-        type=":class:`Plant`",
+        doc="Instanced canopy element. Can be specified as a dictionary, which will "
+        "be interpreted by :meth:`.CanopyElement.from_dict`.",
+        type=":class:`CanopyElement`",
```
leroyvn, line 1215:
```suggestion
        type=":class:`.CanopyElement`",
```

eradiate/scenes/biosphere/_discrete.py
```diff
@@ -1272,8 +1270,10 @@ def from_file(cls, filename=None, plant=None):
         if not os.path.isfile(filename):
             raise FileNotFoundError(f"no file at {filename} found.")
 
-        if plant is None:
-            plant = Plant()
+        if canopy_element is None:
+            canopy_element = CanopyElement()
```
leroyvn, line 1274: This should break: `CanopyElement` is abstract. Instead, you can initialise with `LeafCloud()` or `AbstractTree()`.

eradiate/solvers/rami/_scene.py
```diff
@@ -122,7 +122,7 @@ def kernel_dict(self, ctx: KernelDictContext = None):
             if self.padding > 0:  # We must add extra instances if padding is requested
                 canopy = self.canopy.padded(self.padding)
                 ctx.override_surface_width = (
-                    max(self.canopy.size[:2]) * 2.0 * self.padding + 1.0
+                    max(self.canopy.size[:2]) * (2.0 * self.padding + 1.0)
```
leroyvn, line 125: Thanks for fixing that!

eradiate/scenes/biosphere/_discrete.py
```diff
@@ -909,6 +1028,159 @@ def bsdfs(self, ctx=None):
...
+        pinttr.ib(default=0.0 * ureg.m, units=ucc.deferred("length")),
+        doc="Trunk height. If unset defaults to zero, yielding no trunk.\n\n"
+        "Unit-enabled field (default: ucc[length]).",
+        type="float",
+        default="0.0 m",
+    )
+
+    trunk_radius = documented(
+        pinttr.ib(default=0.0 * ureg.m, units=ucc.deferred("length")),
+        doc="Trunk radius. If unset defaults to zero, yielding no trunk.\n\nUnit-enabled field (default: ucc[length]).",
```
leroyvn, line 1058:
```suggestion
        doc="Trunk radius. If unset defaults to zero, yielding no trunk.\n"
        "\n"
        "Unit-enabled field (default: ucc[length]).",
```

eradiate/scenes/biosphere/_discrete.py
```diff
@@ -909,6 +1028,159 @@ def bsdfs(self, ctx=None):
...
+        ),
+        doc="Instanced leaf cloud. Can be specified as a dictionary, which will "
+        "be interpreted by :meth:`.LeafCloud.from_dict`.",
+        type=":class:`LeafCloud`",
+        default="None",
+    )
+
+    trunk_height = documented(
+        pinttr.ib(default=0.0 * ureg.m, units=ucc.deferred("length")),
+        doc="Trunk height. If unset defaults to zero, yielding no trunk.\n\n"
```
leroyvn, line 1050: I think the "no trunk" case is irrelevant since you already have `LeafCloud` for this.

eradiate/scenes/biosphere/_discrete.py
```diff
@@ -909,6 +1028,159 @@ def bsdfs(self, ctx=None):
...
+        )
+
+        if ctx.ref:
+            bsdf = {"type": "ref", "id": f"bsdf_{self.id}"}
+        else:
+            bsdf = self.bsdfs(ctx=ctx)[f"bsdf_{self.id}"]
+
+        shapes_dict = self.leaf_cloud.shapes(ctx=ctx)
+
+        shapes_dict["trunk_cyl"] = {
```
leroyvn, line 1132: I think if you put multiple trees in your scene, this entry will be overwritten. Mind the shape naming pattern, see other classes for examples.

eradiate/scenes/biosphere/_discrete.py
```diff
@@ -909,6 +1028,159 @@ def bsdfs(self, ctx=None):
...
+            "type": "cylinder",
+            "bsdf": bsdf,
+            "radius": kernel_radius,
+            "p0": [0, 0, -0.1],
+            "p1": [0, 0, kernel_height],
+        }
+
+        shapes_dict["trunk_cap"] = {
```
leroyvn, line 1140: Ditto

leroyvn (changes requested):
Hi @schunkes, a few things to change. I'd appreciate if your bugfixes could land directly on `main`, obviously correct things need no review.

eradiate/scenes/biosphere/_discrete.py
schunkes: I don't see another way to get the offset information into the `leaf_cloud` object. I don't think it is a good idea to add the offset as a parameter to the `kernel_dict` method.
If anything, the leaf cloud can get a new member variable again, which holds the offset and which is then written into by `AbstractTree` :man_shrugging: 
Or do you think I should modify all the transformation matrices after I called `leaf_cloud.shape()` ?

schunkes: Also, using `+=` does not work here anyways. (ref: https://techoverflow.net/2019/05/22/how-to-fix-numpy-typeerror-cannot-cast-ufunc-subtract-output-from-dtypefloat64-to-dtypeint64-with-casting-rule-same_kind/)

leroyvn: > I don't see another way to get the offset information into the leaf_cloud object. I don't think it is a good idea to add the offset as a parameter to the kernel_dict method.

The context system should handle this kind of cases. The best is probably to subclass `KernelDictContext` (*e.g.* as a local member of `LeafCloud`), add a `translate` field to it which will take a 3-vector as value and initialise an instance of this from a parent `KernelDictContext`.

> Also, using += does not work here anyways. (ref: https://techoverflow.net/2019/05/22/how-to-fix-numpy-typeerror-cannot-cast-ufunc-subtract-output-from-dtypefloat64-to-dtypeint64-with-casting-rule-same_kind/)

I think it would with `np.array([0, 0, kernel_height], dtype=float).reshape((1, 3))`.

leroyvn (changes requested):
See offline comments for a more elegant way to implement leaf cloud displacement.

leroyvn (changes requested):

docs/rst/reference/scenes.rst
```diff
@@ -56,8 +56,10 @@ Biosphere [eradiate.scenes.biosphere]
...
    :toctree: generated/
 
+   CanopyElement
    LeafCloud
-   InstancedLeafCloud
+   AbstractTree
+   InstancedCanopyElement
    DiscreteCanopy
 
 .. dropdown:: **Parameters for LeafCloud generators**
```
leroyvn, line 65: Would be nice to also add a link to `_discrete.ConeLeafCloudParams`

eradiate/scenes/biosphere/__init__.py
```diff
@@ -1,10 +1,12 @@
 from ._core import BiosphereFactory, Canopy
-from ._discrete import DiscreteCanopy, InstancedLeafCloud, LeafCloud
+from ._discrete import DiscreteCanopy, InstancedCanopyElement, LeafCloud, AbstractTree, CanopyElement
```
leroyvn, line 2: isort / Black pass please 🙏

eradiate/scenes/biosphere/tests/test_biosphere_discrete.py
```diff
@@ -323,57 +324,244 @@ def test_leaf_cloud_kernel_dict(mode_mono):
...
+
+    # A LeafCloud instance can be loaded from a file on the hard drive
+    tree = AbstractTree(leaf_cloud=LeafCloud.from_file(tempfile_leaves))
+    assert len(tree.leaf_cloud.leaf_positions) == 5
+    assert np.allclose(tree.leaf_cloud.leaf_radii, 0.1 * ureg.m)
+    # Produced kernel dict is valid
+    assert KernelDict.new(tree.leaf_cloud, ctx=ctx).load()
+
+    # Use regular constructor if no "type" parameter is specified
+    assert AbstractTree(leaf_cloud=LeafCloud.from_dict(
```
leroyvn, line 355: isort / Black pass please 🙏

eradiate/scenes/biosphere/_discrete.py
```diff
@@ -947,18 +1066,161 @@ def bsdfs(self, ctx=None):
...
+            Dictionary containing parameters passed to the selected constructor.
+            Unit fields are pre-processed with :func:`pinttr.interpret_units`.
+        """
+
+        # Interpret unit fields if any
+        d_copy = pinttr.interpret_units(d, ureg=ureg)
+
+        leaf_cloud_dict = d_copy.pop("leaf_cloud")
+
+        leaf_cloud = LeafCloud.convert(leaf_cloud_dict)
```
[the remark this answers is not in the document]
leroyvn: Did you have the opportunity to check this?

leroyvn (changes requested):
Hi @schunkes, nice! I have a couple of minor comments. Once this is taken care of, please feel free to squash and merge :)

eradiate/scenes/biosphere/_discrete.py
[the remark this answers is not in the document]
schunkes: Your approach would lead to a name collision in the `AbstractTree` constructor. On one hand I pass the instantiated leaf cloud from Line 1233 as the `leaf_cloud` argument but the `*d_copy` would add another `**kwarg` with the name `leaf_cloud` :thinking:

PR #90, opened 2021-06-15
Particle layers

nollety:
# Description

Partially resolves eradiate/eradiate-issues#74 (this pull request is part of a series of pull requests to completely resolve eradiate/eradiate-issues#74).

This pull request defines a class `ParticleLayer` to represent particle layers in the atmosphere (i.e. clouds and aerosols).

Particle layer objects are created like this:

```python
from eradiate import unit_registry as ureg
from eradiate.radprops.particles import ParticleLayer

layer = ParticleLayer(
  bottom=ureg.Quantity(5, "km"),
  top=ureg.Quantity(7, "km"),
  vert_dist={
    "type": "gaussian",
  },
  tau_550=0.2,
  data_set="tests/radprops/rtmom_aeronet_desert.nc",
)
```

The class implements `eval_albedo()`, `eval_sigma_t()` methods to evaluate the collision coefficients in the particle layer, as well as a `eval_phase()` method to return the scattering phase function.

Note: this pull request does not integrate the particle layer(s) in the atmosphere, this is the job of #91.

# Checklist

### DIFF
--- a/eradiate/radprops/particle_dist.py
+++ b/eradiate/radprops/particle_dist.py
@@ -15,7 +15,7 @@ from scipy.stats import expon, norm
 
 from .._attrs import documented, parse_docs
 from .._factory import BaseFactory
-from ..units import to_quantity
+from ..validators import all_positive
 from ..units import unit_context_config as ucc
 from ..units import unit_registry as ureg
 
@@ -41,30 +41,6 @@ class ParticleDistribution(ABC):
     where :math:`f_i` is the particle number fraction in the layer division
     :math:`i`.
     """
-    bottom = documented(
-        pinttr.ib(
-            units=ucc.deferred("length"),
-            converter=pinttr.converters.to_units(ucc.deferred("length")),
-            validator=pinttr.validators.has_compatible_units,
-        ),
-        doc="Layer bottom altitude.\n\nUnit-enabled field (default: ucc[length]).",
-        type="float",
-    )
-    top = documented(
-        pinttr.ib(
-            units=ucc.deferred("length"),
-            converter=pinttr.converters.to_units(ucc.deferred("length")),
-            validator=pinttr.validators.has_compatible_units,
-        ),
-        doc="Layer top altitude.\n\nUnit-enabled field (default: ucc[length]).",
-        type="float",
-    )
-
-    @bottom.validator
-    @top.validator
-    def _bottom_and_top_validator(instance, attribute, value):
-        if instance.bottom >= instance.top:
-            raise ValueError("bottom altitude must be lower than top altitude")
 
     @classmethod
     def from_dict(cls, d: dict) -> ParticleDistribution:
@@ -106,7 +82,7 @@ class ParticleDistributionFactory(BaseFactory):
 @ParticleDistributionFactory.register("uniform")
 @parse_docs
 @attr.s
-class Uniform(ParticleDistribution):
+class UniformParticleDistribution(ParticleDistribution):
     r"""
     Uniform particle distribution.
 
@@ -122,22 +98,14 @@ class Uniform(ParticleDistribution):
     """
 
     def eval_fraction(self, z: pint.Quantity) -> np.ndarray:
-        if np.any(z < self.bottom) or np.any(z > self.top):
-            raise ValueError(
-                f"Altitude values do not lie between layer "
-                f"bottom ({self.bottom}) and top ({self.top}) "
-                f"altitudes. Got {z}."
-            )
-        else:
-            x = z.magnitude
-            f = np.ones(len(x))
-            return f / f.sum()  # Normalise fractions
+        f = np.ones(len(z))
+        return f / f.sum()
 
 
 @ParticleDistributionFactory.register("exponential")
 @parse_docs
 @attr.s
-class Exponential(ParticleDistribution):
+class ExponentialParticleDistribution(ParticleDistribution):
     r"""
     Exponential particle distribution.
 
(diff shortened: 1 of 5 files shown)


eradiate/radprops/particles.py
```diff
@@ -0,0 +1,423 @@
...
+            ],
+        ),
+        doc="Top altitude of the particle layer.\n"
+        "\n"
+        "Unit-enabled field (default: ucc[length]).",
+        type="float",
+        default="1 km.",
+    )
+
+    vert_dist = documented(
```
leroyvn, line 87: I'd rather go with a more explicit `distribution` (the "vertical" part is anyway unambiguous at this stage); would that be fine?

eradiate/radprops/particles.py
```diff
@@ -0,0 +1,423 @@
...
+        doc="Particles radiative properties data set path.",
+        type="str",
+    )
+
+    def __attrs_post_init__(self):
+        # propagate 'bottom' and 'top' to particle vertical distribution config
+        if isinstance(self.vert_dist, dict):
+            d = self.vert_dist
+            d.update({"bottom": self.bottom, "top": self.top})
+            self.vert_dist = ParticleDistributionFactory.convert(d)
```
leroyvn, line 157: This conversion operation won't be applied if you pass a `ParticleDistribution`: isn't that a problem?

eradiate/radprops/particles.py
```diff
@@ -0,0 +1,423 @@
...
+        # determine layers number based on vertical distribution type
+        if self.n_layers is None:
+            if isinstance(self.vert_dist, Uniform):
+                self.n_layers = 1
+            elif isinstance(self.vert_dist, Exponential):
+                self.n_layers = 8
+            elif isinstance(self.vert_dist, Gaussian):
+                self.n_layers = 16
+            elif isinstance(self.vert_dist, Array):
+                self.n_layers = 32
```
leroyvn, line 168: Seems complicated to me, and will break if you add a new `ParticleDistribution` subclass. Is there any strong con against going with the same value for all distributions?

eradiate/radprops/particles.py
```diff
@@ -0,0 +1,423 @@
...
+                is_positive,
+                pinttr.validators.has_compatible_units,
+            ],
+        ),
+        doc="Top altitude of the particle layer.\n"
+        "\n"
+        "Unit-enabled field (default: ucc[length]).",
+        type="float",
+        default="1 km.",
+    )
```
leroyvn, line 85: Since you want these two to be identical to what is in the distribution, how about turning them into properties? 

If you need these fields just for a nicer syntax during object construction, you can update `from_dict()` to handle this.

leroyvn (commented):
Hi @nollety, thanks for this PR. I made a few comments: I think the `bottom` and `top` fields could be turned into properties. Your tests require a new data file: was it already added to the data repo?

eradiate/radprops/particles.py
nollety: Yes, that would perfectly fine with me. Actually, I went for `distribution` at first but @Ygovs commented that this was ambiguous as it could be confused with the particles **size distribution** and he asked that I changed it to something more explicit.

eradiate/radprops/particles.py
nollety: Hmm, yes that's a problem actually

eradiate/radprops/particles.py
nollety: You're right, I guess there is no strong con against going with the same value for all distributions.

eradiate/radprops/particles.py
nollety: I do want (or need) that `bottom` and `top` are identical in both the particle layer and the particle distribution, yes

nollety: I was just thinking about removing them from the `ParticleDistribution` constructor, since the particle distribution can be defined independently of `bottom` and `top`, and evaluated later given the values of `bottom` and `top` in the `ParticleLayer` object.

nollety: The interface would look like this:

```python
from eradiate.radprops.particle_dist import Gaussian

ParticleLayer(
  bottom=ureg.Quantity(4, "km"),
  top=ureg.Quantity(5, "km"),
  distribution=Gaussian(),
  tau_550=0.1,
  data_set="tests/radprops/rtmom_aeronet_desert.nc"
```

instead of 

```python
ParticleLayer(
  distribution=Gaussian(
    bottom=ureg.Quantity(4, "km"),
    top=ureg.Quantity(5, "km"),
  ),
  tau_550=0.1,
  data_set="tests/radprops/rtmom_aeronet_desert.nc"
```

What do we prefer?

nollety: The second one looks more complicated to me, the first one is more intuitive I feel

nollety:
@leroyvn about the new data file required by the tests: yes, it's already in eradiate-data and the submodule was updated already in the previous pull request.

nollety:
@leroyvn as discussed in the office, I re-defined the `ParticleDistribution` class so that it does not have `bottom` and `top` attributes but only attributes specific to the distribution types (e.g. mean, std, etc). The attributes bottom and top are defined only once, at the `ParticleLayer`-level. The particle distribution is evaluated with `ParticleDistribution.eval_fraction()` on an altitude mesh determined by the `ParticleLayer` object. I think this is much simpler than before, the interface now looks like that:

```python
ParticleLayer(
  bottom=ureg.Quantity(2, "km"),
  top=ureg.Quantity(4, "km"),
  distribution=Gaussian(
    mean=ureg.Quantity(3.0, "km"),
    std=ureg.Quantity(300, "m"),
  ),
  tau_550=0.1,
  n_layers=32,
  dataset="tests/radprops/rtmom_aeronet_desert.nc",
)
```

eradiate/radprops/particle_dist.py
```diff
@@ -288,27 +220,33 @@ class Array(ParticleDistribution):
...
-        type=":class:`xarray.DataArray`",
+        doc="Particle distribution data array. Number fraction as a "
+        "function of altitude (``z``).\n"
+        "Note that number fraction do not need to be normalised.",
+        type=":class:`~xarray.DataArray`",
         default="``None``",
     )
 
+    @data_array.validator
+    def _validate_data_array(instance, attribute, value):
```
leroyvn, line 231: This function is defined twice, isn't it a problem?

eradiate/radprops/particles.py
```diff
@@ -0,0 +1,386 @@
...
+
+    @bottom.validator
+    @top.validator
+    def _bottom_and_top_validator(instance, attribute, value):
+        if instance.bottom >= instance.top:
+            raise ValueError("bottom altitude must be lower than top altitude")
+
+    distribution = documented(
+        attr.ib(
+            default={"type": "uniform"},
```
leroyvn, line 88: Style: I'd prefer if you'd use the constructor itself here:
```suggestion
            factory=UniformParticleDistribution,
```

leroyvn (commented):
Hi @nollety, seems good to me. I just have a couple of minor comments. One thing I just realised, though: you named particle distribution classes without a suffix, which contradicts our usual naming pattern and increases risk for name collisions. I'd appreciate if you could remedy this (*e.g.* `Gaussian` → `GaussianParticleDistribution`, etc.) either as part of this PR or in another. Apart from that, nothing to add, it's good for me!

eradiate/radprops/particle_dist.py
nollety: I assume it is, I renamed that function

nollety:
Hi @leroyvn, thanks for the feedback :)

I applied your suggestions and renamed the particle distribution classes to comply with our conventions for classes' names.

nollety:
@leroyvn, rebased and squashed!

leroyvn:
Seems good to me, let's go with this!

PR #96, opened 2021-06-22
Added a system test (lambertian BRF)

nollety:
# Description

Adds a system test for the measured BRF of a lambertian surface.
The test simply checks that the measured BRF matches the surface's reflectance.
Initially the idea of this test was to catch BRF deviations around 0.0° viewing zenith angle, as observed previously (e.g. eradiate/eradiate-issues#34).

# Checklist

### DIFF
--- a/eradiate/tests/system/test_onedim_lambertian_brf.py
+++ b/eradiate/tests/system/test_onedim_lambertian_brf.py
@@ -1,0 +1,127 @@
+"""Test cases with OneDimSolverApp and a lambertian surface."""
+
+import os
+
+import matplotlib
+import matplotlib.pyplot as plt
+import numpy as np
+import pytest
+
+import eradiate
+
+matplotlib.rcParams["text.usetex"] = True
+
+eradiate_dir = os.environ["ERADIATE_DIR"]
+output_dir = os.path.join(eradiate_dir, "test_report", "generated")
+
+
+def ensure_output_dir(path):
+    if not os.path.isdir(path):
+        os.makedirs(path)
+
+
+def test_measured_lambertian_brf(mode_mono_double):
+    r"""
+    Measured lambertian BRF
+    =======================
+
+    This test case checks that measured BRF matches the Lambertian surface's
+    reflectance.
+
+    Rationale
+    ---------
+
+    * Geometry: a square surface with unit size and a Lambertian BRDF with
+      reflectance :math:`\rho \in [0.0, 0.3, 0.5, 0.7, 1.0]`.
+    * Illumination: Directional illumination with a zenith angle
+      :math:`\theta \in [0.0, 30.0, 60.0]°`.
+    * Sensor: Distant reflectance measure covering a plane (1001 angular points,
+      1 sample per pixel) and targeting (0, 0, 0).
+      We take an odd number of angular points to ensure that the special value
+      :math:`\theta = 0°` is included.
+
+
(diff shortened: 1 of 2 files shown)


eradiate/tests/system/test_onedim_lambertian_brf.py
```diff
@@ -0,0 +1,119 @@
...
+    .. image:: generated/plots/test_measured_lambertian_brf_0.0.png
+       :width: 75%
+
+    .. image:: generated/plots/test_measured_lambertian_brf_30.0.png
+       :width: 75%
+
+    .. image:: generated/plots/test_measured_lambertian_brf_60.0.png
+       :width: 75%
+    """
+    spp = int(1e5)
```
leroyvn, line 60:
```suggestion
    spp = 1
```

eradiate/tests/system/test_onedim_lambertian_brf.py
```diff
@@ -0,0 +1,119 @@
...
+
+    Rationale
+    ---------
+
+    * Geometry: a square surface with unit size and a Lambertian BRDF with
+      reflectance :math:`\rho \in [0.0, 0.3, 0.5, 0.7, 1.0]`.
+    * Illumination: Directional illumination with a zenith angle
+      :math:`\theta \in [0.0, 30.0, 60.0]°`.
+    * Sensor: Distant reflectance measure covering a plane (7 angular points,
+      :math:`10^5` samples per pixel) and targeting (0, 0, 0).
```
leroyvn, line 40: A single sample per pixel should already yield a correct value.

eradiate/tests/system/test_onedim_lambertian_brf.py
```diff
@@ -0,0 +1,119 @@
...
+    .. image:: generated/plots/test_measured_lambertian_brf_60.0.png
+       :width: 75%
+    """
+    spp = int(1e5)
+    n_vza = 7
+    zenith_values = [0.0, 30.0, 60.0]
+    reflectance_values = [0.0, 0.3, 0.5, 0.7, 1.0]
+
+    results = {}
+    for zenith in zenith_values:
```
leroyvn, line 66: `zenith` is not explicit enough IMO, I suggest something like `illumination_zenith` for improved clarity.

leroyvn (changes requested):
Hi @nollety, thanks for this PR. I think the sample count you're using is overkill: instead, you could remove the `slow` mark and increase the film resolution (I just tried with 1 SPP and a 1000 pixel film). I also have a few minor comments (see below).

nollety:
@leroyvn, thanks for your feedback! Indeed, I forgot about the atmosphere not being there in this test case series... 10^5 is definitely too much ;)

nollety:
@leroyvn I don't know if you noticed but I changed a previous system test (test_onedim_symmetry) to take an odd number of viewing zenith angles so that the special value 0° is included. This allows to check (visually and/or numerically) that there is no dip at 0°.

nollety:
@leroyvn, rebased and squashed!

leroyvn:
> I don't know if you noticed but I changed a previous system test (test_onedim_symmetry) to take an odd number of viewing zenith angles so that the special value 0° is included. This allows to check (visually and/or numerically) that there is no dip at 0°.

I saw it, it's a good thing to do indeed!

Looks all good to me, let's merge that!

PR #101, opened 2021-06-28
Refactoring discrete

schunkes:
# Description

I refactored the `_discrete` module, splitting it into multiple modules, because the monolithic module had become quite large.

The `_discrete` module only holds code for the discrete canopy itself now.

# Checklist


leroyvn (changes requested):
Hi @schunkes, thanks for this PR, modules are now much better scoped! I have a few general comments about file naming, and a few in-source comments (see below).

* Is there a particular reason why the `canopy_element` should be public? (*i.e.* directly accessed by users) Judging by the imports in `__init__.py`, I assume not. Then I'd rather have it underscore-prefixed to explicitly signal to the user that the entry point is the `eradiate.scenes.biosphere` module.
* Test file naming should follow a naming pattern which will reduce risk for name collisions and help mapping to the module where tested components are located. Consequently, I'd prefer if `test_abstract_tree.py` would be `test_biosphere_tree_abstract.py`.

The latter point matters because if multiple files in the test suite have the same name, pytest will crash.

leroyvn:
Looks good, let's merge this!

PR #103, opened 2021-06-29
Mesh based canopy

schunkes:
# Description

This PR introduces a mesh based canopy element.
It is actually made up of two classes: `MeshTree` and `MeshTreeElement`. The former simply holds a list of objects of the latter class, since one tree might be made up of multiple meshes (e.g. the leaves and the trunk/branches). The latter holds the mesh and its radiative properties.

The branch is not yet rebased on the current state of main because that is changing a lot right now, anyways.

# Checklist

- [] The feature branch is rebased on the current state of the `main` branch


leroyvn:
Look good! Let's go!

PR #102, opened 2021-06-28
New heterogeneous atmosphere design

nollety:
Resolves eradiate/eradiate-issues#74.

# Description

This pull request introduces a new heterogeneous atmosphere design. 

Prerequisite pull requests:

* #111 
* #112

## Motivation

This new design was motivated by the need to integrate `ParticleLayer` objects in the heterogeneous atmosphere. The integration proved to be rather complex, for various reasons including:

* the level altitude meshes do not align
* the radiative properties must be "blended" together, including the phase functions

which is why this new design is suggested.

## New design layout

The new design is illustrated by the UML diagram below:

uml_diagram

* `MolecularAtmosphere` is the molecular part of the atmosphere, i.e. without particles. Each type of molecular atmospheres (e.g. AFGL1986) is associated to a specific class method (specified by `"construct"` in the corresponding kernel dict).
* `ParticleLayer` objects are 1D layers of particles (e.g. a cloud layer).
* `HeterogeneousAtmosphere` is an object that blends a `MolecularAtmosphere` object with one or more `ParticleLayer` objects. It has special methods that lets it align the altitude meshes of its components, compute the total collision coefficients and phase function blending ratios based on the individual radiative properties of each component, etc.

## Changes

* added a module `scenes.atmosphere._heterogeneous_new` where `HeterogeneousAtmosphereNew` is defined. We append __new_ or _New_ where necessary, in order to avoid breaking the existing implementation of `HeterogeneousAtmosphere`. Once the new design is completely implemented and tested, then it will be time to rename these. Corresponding tests are in `scenes/atmosphere/tests/test_heterogeneous_new.py`.
* added a module `_molecules` where `MolecularAtmosphere` is defined. Corresponding tests are in `scenes/atmosphere/tests/test_molecules.py`.
* fixed a typo to `_ROOTDIR` value in `eradiate/data/absorption_spectra.py`
* updated documentation

### `ParticleLayer`

`ParticleLayer` now inherits `Atmosphere` as indicated in the UML diagram above.
`ParticleLayer` is then now a scene element, which means

```python
from eradiate.contexts import KernelDictContext
from eradiate.scenes.atmosphere import ParticleLayer
from eradiate.units import unit_registry as ureg

ctx = KernelDictContext()
particle_layer = ParticleLayer(
    bottom = 0.0 * ureg.km,
    top = 1.5 * ureg.km,
    tau_550 = 0.1 * ureg.dimensionless,
    n_layers=9,
)
[...]

PR #106, opened 2021-06-30
Updated the Atmosphere interface

nollety:
# Description

This pull request updates the `Atmosphere` classes interface in anticipation of the ongoing work in #102.

# Changes

* `Atmosphere`:
  * `toa_altitude` attribute is removed from `Atmosphere`.
  * a `height` property is added.
  * following abstract properties are added:
    * `bottom`: returns the bottom altitude of the atmosphere
    * `top`: returns the top altitude of the atmosphere, i.e. the same value that `toa_altitude` hold
  * following abstract method is added:
    * `eval_width`; returns the width of the atmosphere
* `HomogeneousAtmosphere`:
  * added `_bottom` (defaults to 0 km) and `_top` (defaults to 10 km) attributes
  * implemented `bottom` and `top` properties and `eval_width()` method.
  * updated related tests
* `HeterogeneousAtmosphere`:
  * removed `height()` method
  * implemented `bottom` and `top` properties and `eval_width` method
  * changed `albedo_fname` and `sigma_t_fname` to hold only the file names (not the full path) : these parameters cannot be `None` (default to `"albedo.vol"` and `"sigma_t.vol"`)
  * added properties `albedo_file` `sigma_t_file` to return full path to these volume data files.
  * removed obsolete tests
* updated system test `test_onedim_symmetry`

`HeterogeneousAtmosphere` can no longer be instanciated with volume data files (this feature complicated the design a lot). `HeterogeneousAtmosphere` must have a non-`None` profile ; by default, `profile` is set to `US76ApproxRadProfile()`.

# Checklist

### DIFF
--- a/eradiate/kernel/tests/test_transform.py
+++ b/eradiate/kernel/tests/test_transform.py
@@ -1,0 +1,73 @@
+import enoki as ek
+
+from eradiate.kernel.transform import map_cube, map_unit_cube
+
+
+def test_map_unit_cube(mode_mono):
+    """
+    Returns a transformation that maps old cube vertices to new cube vertices.
+    """
+    trafo = map_unit_cube(1, 2, 3, 4, 5, 6)
+
+    vertices = [
+        [0, 0, 0],
+        [1, 0, 0],
+        [0, 1, 0],
+        [1, 1, 0],
+        [0, 0, 1],
+        [1, 0, 1],
+        [0, 1, 1],
+        [1, 1, 1],
+    ]
+
+    new_vertices = [
+        [1, 3, 5],
+        [2, 3, 5],
+        [1, 4, 5],
+        [2, 4, 5],
+        [1, 3, 6],
+        [2, 3, 6],
+        [1, 4, 6],
+        [2, 4, 6],
+    ]
+
+    for vertex, new_vertex in zip(vertices, new_vertices):
+        assert ek.allclose(
+            trafo.transform_point(vertex),
+            new_vertex,
+        )
+
+
+def test_map_cube():
+    """
+    Returns a transformation that maps old cube vertices to new cube vertices.
+    """
+    trafo = map_cube(1, 2, 3, 4, 5, 6)
+
+    vertices = [
+        [-1, -1, -1],
+        [1, -1, -1],
+        [-1, 1, -1],
+        [1, 1, -1],
+        [-1, -1, 1],
+        [1, -1, 1],
+        [-1, 1, 1],
+        [1, 1, 1],
+    ]
+
+    new_vertices = [
+        [1, 3, 5],
+        [2, 3, 5],
+        [1, 4, 5],
+        [2, 4, 5],
+        [1, 3, 6],
+        [2, 3, 6],
+        [1, 4, 6],
+        [2, 4, 6],
+    ]
+
+    for vertex, new_vertex in zip(vertices, new_vertices):
+        assert ek.allclose(
+            trafo.transform_point(vertex),
+            new_vertex,
+        )
--- a/eradiate/kernel/transform.py
+++ b/eradiate/kernel/transform.py
@@ -1,0 +1,80 @@
+from typing import Any
+
+
+def map_unit_cube(
+    xmin: float, xmax: float, ymin: float, ymax: float, zmin: float, zmax: float
+) -> Any:
+    """
+    Map the unit cube to [xmin, xmax] x [ymin, ymax] x [zmin, zmax].
+
+    .. note::
+        You must select a Mitsuba variant before calling this function.
+
+    Parameter ``xmin`` (float):
+        Minimum X value.
+
+    Parameter ``xmax`` (float):
+        Maximum X value.
+
+    Parameter ``ymin`` (float):
+        Minimum Y value.
+
+    Parameter ``ymax`` (float):
+        Maximum Y value.
+
+    Parameter ``zmin`` (float):
+        Minimum Z value.
+
+    Parameter ``zmax`` (float):
+        Maximum Z value.
+
+    Returns → :class:`~mitsuba.core.ScalarTransform4f`
+        Transform matrix.
+    """
+    from mitsuba.core import ScalarTransform4f
+
+    scale_trafo = ScalarTransform4f.scale([xmax - xmin, ymax - ymin, zmax - zmin])
+    translate_trafo = ScalarTransform4f.translate([xmin, ymin, zmin])
+    return translate_trafo * scale_trafo
+
+
+def map_cube(
+    xmin: float, xmax: float, ymin: float, ymax: float, zmin: float, zmax: float
+) -> Any:
+    """
+    Map the cube [-1, 1]^3 to [xmin, xmax] x [ymin, ymax] x [zmin, zmax].
+
+    .. note::
+        You must select a Mitsuba variant before calling this function.
+
+    Parameter ``xmin`` (float):
+        Minimum X value.
+
+    Parameter ``xmax`` (float):
+        Maximum X value.
+
+    Parameter ``ymin`` (float):
+        Minimum Y value.
+
+    Parameter ``ymax`` (float):
+        Maximum Y value.
+
+    Parameter ``zmin`` (float):
+        Minimum Z value.
+
+    Parameter ``zmax`` (float):
+        Maximum Z value.
+
+    Returns → :class:`~mitsuba.core.ScalarTransform4f`
+        Transform matrix.
+    """
+    from mitsuba.core import ScalarTransform4f
+
+    half_dx = (xmax - xmin) * 0.5
+    half_dy = (ymax - ymin) * 0.5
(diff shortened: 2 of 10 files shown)


eradiate/kernel/tests/test_transform.py
```diff
@@ -0,0 +1,74 @@
+import enoki as ek
```
leroyvn, line 1: These tests should fail (it's only by luck if they don't). Mitsuba imports should be function-scoped, and you should use a mode selection fixture for all your tests to ensure that you're always selecting the same mode. Corrections below.

eradiate/kernel/tests/test_transform.py
```diff
@@ -0,0 +1,74 @@
...
+from mitsuba.core import Point3f
```
leroyvn, line 2:
```suggestion
```

eradiate/kernel/tests/test_transform.py
```diff
@@ -0,0 +1,74 @@
...
+
+from eradiate.kernel.transform import map_cube, map_unit_cube
+
+
+def test_map_unit_cube():
```
leroyvn, line 7:
```suggestion
def test_map_unit_cube(mode_mono):
```

eradiate/kernel/tests/test_transform.py
```diff
@@ -0,0 +1,74 @@
...
+    ]
+
+    for vertex, new_vertex in zip(vertices, new_vertices):
+        assert ek.allclose(
+            trafo.transform_point(vertex),
+            new_vertex,
+        )
+
+
+def test_map_cube():
```
leroyvn, line 42: Same here.

eradiate/kernel/transform.py
```diff
@@ -0,0 +1,78 @@
...
+
+    scale_trafo = ScalarTransform4f.scale([xmax - xmin, ymax - ymin, zmax - zmin])
+    translate_trafo = ScalarTransform4f.translate([xmin, ymin, zmin])
+    return translate_trafo * scale_trafo
+
+
+def map_cube(
+    xmin: float, xmax: float, ymin: float, ymax: float, zmin: float, zmax: float
+) -> Any:
+    """Map the cube [-1, 1]^3 to [xmin, xmax] x [ymin, ymax] x [zmin, zmax].
```
leroyvn, line 43:
```suggestion
    """
    Map the cube [-1, 1]^3 to [xmin, xmax] x [ymin, ymax] x [zmin, zmax].
```

eradiate/kernel/transform.py
```diff
@@ -0,0 +1,78 @@
+from typing import Any
+
+
+def map_unit_cube(
+    xmin: float, xmax: float, ymin: float, ymax: float, zmin: float, zmax: float
+) -> Any:
+    """Map the unit cube to [xmin, xmax] x [ymin, ymax] x [zmin, zmax].
```
leroyvn, line 7: We try to start all docstrings with a newline (required for class docstrings, good practice for others):

```suggestion
    """
    Map the unit cube to [xmin, xmax] x [ymin, ymax] x [zmin, zmax].
```

eradiate/scenes/atmosphere/_core.py
```diff
@@ -18,9 +19,15 @@
 class Atmosphere(SceneElement, ABC):
     """
     An abstract base class defining common facilities for all atmospheres.
+
+    An atmosphere consists of a kernel medium (with a phase function) attached
+    to a kernel shape.
+
+    .. note::
+        The shape type is restricted to cuboid shapes at the moment.
```
leroyvn, line 27:
```suggestion
       The shape type is restricted to cuboid shapes at the moment.
```

eradiate/scenes/atmosphere/_core.py
```diff
@@ -41,27 +48,8 @@ class Atmosphere(SceneElement, ABC):
...
-                pinttr.converters.to_units(ucc.deferred("length"))
-            ),
-            validator=validators.auto_or(
-                pinttr.validators.has_compatible_units, validators.is_positive
-            ),
-            units=ucc.deferred("length"),
-        ),
-        doc='Atmosphere width. If set to ``"auto"``, a value will be estimated to '
+        doc="Atmosphere width (i.e., horizontal spatial extension)."
+        'If set to ``"auto"``, a value will be estimated to '
```
leroyvn, line 52: You should transition to the new `AUTO` singleton here.

eradiate/scenes/atmosphere/_core.py
```diff
@@ -71,66 +59,48 @@ class Atmosphere(SceneElement, ABC):
...
-        If ``toa_altitude`` is set to ``"auto"``, a value of 100 km is returned;
-        otherwise, ``toa_altitude`` is returned.
-        """
-        if self.toa_altitude == "auto":
-            return ureg.Quantity(100.0, ureg.km)
-        else:
-            return self.toa_altitude
+    # --------------------------------------------------------------------------
+    #                       Abstract instance methods
+    # --------------------------------------------------------------------------
```
leroyvn, line 64: Rather than whether methods are abstract or not, I'd prefer to categorise components based on their intent as part of the interface this class defines. In that case, relevant categories would be:

- fields and init sequence (no banner needed)
- properties (characterise the object itself)
- kernel dictionary generation

eradiate/scenes/atmosphere/_homogeneous.py
```diff
@@ -16,11 +24,42 @@ class HomogeneousAtmosphere(Atmosphere):
...
+    cuboid shape.
     """
 
-    sigma_s = documented(
+    _bottom: pint.Quantity = documented(
+        pinttr.ib(
+            default=ureg.Quantity(0.0, ureg.km),
+            units=ucc.deferred("length"),
+            converter=pinttr.converters.to_units(ucc.deferred("length")),
+            validator=pinttr.validators.has_compatible_units,
```
leroyvn, line 37: Not required (having this handled by default is the whole point of Pinttrs ;) ).

eradiate/scenes/atmosphere/_homogeneous.py
```diff
@@ -16,11 +24,42 @@ class HomogeneousAtmosphere(Atmosphere):
...
+        type="float",
+        default="0 km",
+    )
+
+    _top: pint.Quantity = documented(
+        pinttr.ib(
+            default=ureg.Quantity(10.0, ureg.km),
+            units=ucc.deferred("length"),
+            converter=pinttr.converters.to_units(ucc.deferred("length")),
+            validator=pinttr.validators.has_compatible_units,
```
leroyvn, line 49: Ditto

eradiate/scenes/atmosphere/_homogeneous.py
```diff
@@ -16,11 +24,42 @@ class HomogeneousAtmosphere(Atmosphere):
...
+        ),
+        doc="Atmosphere's top altitude.\n\nUnit-enabled field (default: ucc[length]).",
+        type="float",
+        default="10 km.",
```
leroyvn, line 53: Any reason to change the default from 100 to 10 km?

eradiate/tests/system/test_onedim_symmetry.py
```diff
@@ -83,7 +83,12 @@ def test_symmetry_zenith(mode_mono_double, surface, atmosphere):
         }[surface],
         atmosphere={
             "none": None,
-            "homogeneous": {"type": "homogeneous", "sigma_s": 1.0e-2},
+            "homogeneous": {
+                "type": "homogeneous",
+                "sigma_s": 1.0e-2,
+                "top": 100,
+                "top_units": "km",
+            },
```
leroyvn, line 91: In order to characterise more reliably the test scene, I'd suggest:
```suggestion
            "homogeneous": {
                "type": "homogeneous",
                "sigma_s": 1.0e-2 * ureg.km ** -1,
                "top": 1.0e2 * ureg.km,
            },
```
This way, your test slab always has an optical thickness equal to 1, no matter what strange thing default units could do in the future.

leroyvn (changes requested):
Hi @nollety, thanks for this PR. Your kernel-level updates need a few minor changes, nothing critical. I also made a few minor comments, not necessarily calling to action.

eradiate/scenes/atmosphere/_homogeneous.py
nollety: 100 km felt a bit too large, given that the atmosphere is homogeneous. 
The air number density' scale height is roughly 10 km so that when one approximates the earth's atmosphere as a homogeneous medium, it is somehow reasonable to me to assume that there is air below 10 km and nothing above.
Of course, the thermophysical variables vary significantly in that 10 km "homogeneous" atmosphere so the assumption is very crude, but 100 km would be even worse.
Also, 10 km is approximately the height of the troposphere, the first layer of the Earth's atmosphere.
Altogether, it made some sense to go for that value for a homogeneous atmosphere.

leroyvn:
All good, let's go and merge!

PR #117, opened 2021-09-08
Use keyword-only parameters and fields for required kwargs

leroyvn:
# Description

In some places, we added checks to verify that required keyword function parameters are indeed passed by the user. This is awkward and Python 3 actually provides [keyword-only parameters](https://www.python.org/dev/peps/pep-3102/) which achieve this.

This PR replaces parameter value checks with kw-only params where relevant. In addition, since the `SceneElement` base class defines a keyword argument field `id`, it follows that every additional field declared by child classes must either have a default value or be kw-only: the same logic therefore applies.

# Checklist

### DIFF
--- a/eradiate/scenes/biosphere/_core.py
+++ b/eradiate/scenes/biosphere/_core.py
@@ -191,7 +191,7 @@ class InstancedCanopyElement(SceneElement):
     @classmethod
     def from_file(
         cls,
-        filename: t.Optional[os.PathLike] = None,
+        filename: os.PathLike,
         canopy_element: t.Optional[CanopyElement] = None,
     ):
         """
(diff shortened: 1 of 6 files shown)


eradiate/scenes/biosphere/tests/test_biosphere_leaf_cloud.py
```diff
@@ -1,5 +1,6 @@
 import os
 import tempfile
+from fileinput import filename
```
schunkes, line 3: How do you use this? The docs for this method are very short and I can't make sense of them like this :sweat_smile:

schunkes (commented):
Are these all files, where the kw-only approach is applicable? That's so few :thinking:

eradiate/scenes/biosphere/tests/test_biosphere_leaf_cloud.py
leroyvn: Thanks for spotting that, I think it sneaked while I was cleaning up and sorting imports.

leroyvn:
Thanks for you feedback @schunkes. You actually made me realise that the kwarg syntax is actually usable for positional arguments, provided that they are not variadic. So it's actually not that often that we need it. I instead converted relevant params to positional: making them kw-only is mostly about syntactic considerations: we can require a kwarg syntax from the user by making functions kw-only. In the end, I think we shouldn't do that before we define a clear policy (_i.e._ which functions will be kw-only).

There is however one case where it's really required (`SceneElement` child class constructors typically). This is staying.

PR #118, opened 2021-09-14
Distant sensor array

schunkes:
# Description

This PR will introduce a variant of the `DistantMeasure`, that lets users specify a list of directions, which will be treated individually, instead of defining a hemisphere or plane to record leaving radiance on.

There are open questions which must be addressed before this code can be merged:

- The `_postprocess_add_directions` method in `DistantArrayMeasure` adds a new coordinate to the dataset, which lets should let users select the results based on the direction they specified for a sensor. However xarray does not support the use of lists as coordinate entries, so for the moment I add the directions as strings, which is at least informative.
- Since I cannot use xarray's builting averaging methods here, I have to select the correct subset of results along the 'sensor_id' dimension and use `numpy.average`. But this leads to some issues with squeezed dimensions and I suspect it will be problematic if the `spectral_coordinate` dimension is not a scalar.

# Checklist

### DIFF
--- a/eradiate/scenes/measure/_core.py
+++ b/eradiate/scenes/measure/_core.py
@@ -432,7 +432,6 @@ class MeasureResults:
         spectral_coords, sensor_ids, film_size = self._to_dataset_helper_coord_values(
             self.raw
         )
-
         # Collect radiance values
         data = self._to_dataset_helper_data_values(
             self.raw, spectral_coords, sensor_ids, film_size
@@ -482,20 +481,59 @@ class MeasureResults:
         if not aggregate_spps:
             return result
 
-        # Aggregate SPPs and drop sensor ID dimension
-        # Note: This will not work if multiple sensors are used for a purpose
-        # other than SPP splitting; should this happen, this part must be
-        # updated
-        weights = result.data_vars["spp"]
-        result_aggregated = result.assign(
-            {"raw": result["raw"].weighted(weights).mean(dim="sensor_id")},
+        # Else ...
+        sensor_id_prefixes = sorted(
+            set(
+                [
+                    sensor_id.split("_spp")[0]
+                    for sensor_id in list(result["sensor_id"].values)
+                ]
+            )
         )
-        result_aggregated = result_aggregated.assign(
-            {"spp": result_aggregated.data_vars["spp"].sum(dim="sensor_id")},
-        )
-        result_aggregated = result_aggregated.drop_dims("sensor_id")
 
-        # Copy metadata (lost during aggregation)
+        raw_aggregated = []
+        spp_aggregated = []
+        for sensor_id_prefix in sensor_id_prefixes:
+            # Select SPP-varying components for the current sensor ID prefix
+            sensor_ids = [
+                id
+                for id in result.coords["sensor_id"].values
+                if id.startswith(sensor_id_prefix)
+            ]
+            # Aggregate sample counts for the current prefix
+            weights = xr.where(
+                result.coords["sensor_id"].isin(sensor_ids),
+                result.data_vars["spp"],
+                0,
+            )
+            raw_aggregated.append(
+                result["raw"]
+                .weighted(weights)
+                .mean(dim="sensor_id")
+                .expand_dims({"sensor_id": [sensor_id_prefix]}, axis=1)
+            )
+            # Mask spp values so as to only include selected values
+            # The sensor_id coordinate is redefined to strip the spp suffix
+            weights_spp = xr.where(result.coords["sensor_id"].isin(sensor_ids), 1, 0)
+            spp_aggregated.append(
+                result.data_vars["spp"]
+                .weighted(weights_spp)
+                .sum(dim="sensor_id")
+                .expand_dims({"sensor_id": [sensor_id_prefix]}, axis=1)
+            )
+
+        raw = xr.combine_by_coords(raw_aggregated)["raw"]
+        spp = xr.combine_by_coords(spp_aggregated)["spp"]
+
+        result_aggregated = xr.Dataset(
+            data_vars={
+                "raw": raw,
+                "spp": spp,
+                "sensor_id": sensor_id_prefixes,
+            }
+        )
+
+        # Copy metadata
         for var in list(result_aggregated.data_vars) + list(result_aggregated.coords):
             result_aggregated[var].attrs = result[var].attrs.copy()
 
@@ -761,6 +799,13 @@ class Measure(SceneElement, ABC):
     def sensor_infos(self) -> t.List[SensorInfo]:
         """
         Return a tuple of sensor information data structures.
+        Sensor ids are generated with a set of suffixes.
+        E.g. in the case of spp splitting, multiple ids, with
+        the suffix '_sppXX' are generated with XX being
+        an integer number.
+
+        Subclasses may override this method and add other suffixes
+        for their specific purposes.
 
         Returns → list[:class:`.SensorInfo`]:
             List of sensor information data structures.
@@ -772,7 +817,8 @@ class Measure(SceneElement, ABC):
 
         else:
             return [
-                SensorInfo(id=f"{self.id}_{i}", spp=spp) for i, spp in enumerate(spps)
+                SensorInfo(id=f"{self.id}_spp{i}", spp=spp)
+                for i, spp in enumerate(spps)
             ]
 
     def _split_spp(self) -> t.List[int]:
--- a/eradiate/scenes/measure/_distant_array.py
+++ b/eradiate/scenes/measure/_distant_array.py
@@ -1,0 +1,324 @@
+from __future__ import annotations
+
+import typing as t
+
+import attr
+import numpy as np
+import pint
+import warnings
+import xarray as xr
+
+from ._core import SensorInfo, measure_factory
+from ._distant import DistantMeasure
+from ..illumination import DirectionalIllumination
+from ...attrs import documented, parse_docs
+from ...units import symbol
+from ...units import unit_context_kernel as uck
+from ...units import unit_registry as ureg
+
+
+@measure_factory.register(type_id="distant_array")
+@parse_docs
+@attr.s
+class DistantArrayMeasure(DistantMeasure):
+    """
+    Distant sensor array measure scene element [``distantarray``].
+
+    This measure scene element is a thin wrapper around the ``DistantSensor``
+    measures. It records either the radiancealong the directions given by the
+    ``directions`` parameter.
+    """
+
+    directions: np.ndarray = documented(
+        attr.ib(
+            default=np.array([[0.0, 0.0, -1.0]]),
+            converter=np.array,
+        ),
+        doc="A sequence of 3-vectors specifying distant sensor directions.",
+        type="array-like",
+        default="[[0, 0, -1]]",
+    )
+
+    @classmethod
+    def from_angles(cls, angles: t.List[t.Tuple[float, float]], **kwargs):
+        """
+        Contruct a :class:`.DistantArrayMeasure` from directions specified in
+        (zenith, azimuth) angle pairs.
+
+        Parameter ``angles`` (array-like):
+            Directions given by 2-vectors of zenith and azimuth angles
+            in degrees.
+
+         Parameter ``**kwargs``:
+            Keyword arguments forwarded to the :class:`.DistantArrayMeasure`
+            constructor.
+
+        Returns → :class:`.DistantArrayMeasure`:
+            Created measure object.
+        """
+        if isinstance(angles, pint.Quantity):
+            directions_converted = angles.to(ureg.rad).magnitude
+        else:
+            directions_converted = np.deg2rad(angles)
+
+        directions = [
+            [
+                np.sin(theta) * np.cos(phi) * -1,
+                np.sin(theta) * np.sin(phi) * -1,
+                np.cos(theta) * -1,
+            ]
+            for (theta, phi) in directions_converted
+        ]
+        return cls(directions=directions, **kwargs)
+
+    @classmethod
+    def azimuthal_ring(
+        cls,
+        zenith_angle: float,
+        direction: t.Optional[t.List] = None,
+        azimuth_resolution: float = 0,
+        azimuth_steps: float = 0,
+        **kwargs,
+    ):
+        """
+        Construct a :class:`.DistantArrayMeasure`, covering a ring of constant zenith angle in the
+        hemisphere given by the direction parameter.
+        The angular resolution along the azimuthal ring can be defined by either the resolution or
+        a number of steps. If both are given, the number of steps takes precedence and a warning is
+        issued.
+
+        Parameter ``zenith_angle`` (float):
+            Zenith angle that defines the ring in the hemisphere.
+            Zenith is measured relative to the ``direction`` parameter
+
+            This parameter can be given as a float in degrees or a :class:`pint.Quantity`.
+
+        Parameter ``direction`` (array-like):
+            Direction that defines the hemisphere in which the azimuthal ring is positioned.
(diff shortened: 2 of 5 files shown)


leroyvn:
Hi @schunkes, a couple of general comments in addition to answers to your questions.

## Sensor direction coordinates

You can keep the sensor dimension linear, make it a simple integer, and add direction data as non-dimension coordinates, quite similar to what is done for `DistantMeasure`. We can then provide helpers to generate optimised direction sets and repack the data in order to make it sliceable (see *Side notes*).

## Dimension issues

I think the best is to allocate a dimension specifically dedicated to SPP split. You'll therefore have 2 dimensions varying sensor ID: a proper `sensor_id` and a `sensor_id_spp` dimension. Collapsing the second dimension should then be easy. In order to do so, you'll have to modify the dataset. I can see two possibilities:

- update the low-level dataset format (not great because it means that the most raw data we output is already processed);
- reindex the data prior to collapsing the SPP dimension (probably better, a little more 'arcane' but not too difficult to experiment thanks to the full access you have to the raw data).

The final assignment would still be something like:

```python
result_aggregated = result.assign(
    {"raw": result["raw"].weighted(weights).mean(dim="sensor_id_spp")},
)
```

## Side notes

(We'll take care of those after this prototype is ready.)

- I think we'll make this measure series the default distant measure and rename the hemispherical one `HDistant*Measure`.
- We'll have to provide a few helpers for users to easily configure directions and format results (*e.g.* plane, ring, hemisphere). How this will be implemented is to be discussed.

eradiate/scenes/measure/_distantsensorarray.py
```diff
@@ -0,0 +1,218 @@
...
+
+from ._core import SensorInfo, measure_factory
+from ._distant import DistantMeasure
+from ..illumination import DirectionalIllumination
+from ...attrs import documented, parse_docs
+from ...units import symbol
+from ...units import unit_context_kernel as uck
+
+
+@measure_factory.register(type_id="distantarray")
```
leroyvn, line 17: Consistency: I'd rather have this called `distant_array` (better correspondence between camel and snake case).

eradiate/scenes/measure/_distantsensorarray.py
```diff
@@ -0,0 +1,218 @@
...
+        doc="A sequence of 3-vectors specifying distant sensor directions.",
+        type="array-like",
+        default="[[0, 0, 1]]",
+    )
+
+    @directions.validator
+    def _directions_validator(self, attribute, value):
+        if value.shape[1] != 3:
+            raise ValueError(
+                f"While initializing {attribute}: "
```
leroyvn, line 43:
```suggestion
                f"While validating {attribute.name}: "
```
(validators can also be run upon setter call)

eradiate/scenes/measure/_distantsensorarray.py
```diff
@@ -0,0 +1,218 @@
...
+                f"Expected shape (N, 3), got {value.shape}"
+            )
+
+    def sensor_infos(self) -> List[SensorInfo]:
+        """
+        Return a tuple of sensor information data structures.
+
+        Returns → list[:class:`.SensorInfo`]:
+            List of sensor information data structures.
+        """
```
leroyvn, line 53: No need to repeat the docstring if it is inherited. However, it would be nice if the parent method  would document the rationale underlying sensor ID definition.

eradiate/scenes/measure/_distantsensorarray.py
```diff
@@ -0,0 +1,218 @@
...
+                for id in ids
+                for i, spp in enumerate(spps)
+            ]
+
+    def _postprocess_add_directions(self, ds: xr.Dataset) -> xr.Dataset:
+        ds = ds.assign_coords(
+            {
+                "directions": (
+                    "sensor_id",
+                    [str(direction) for direction in self.directions],
```
leroyvn, line 73: IMO the best would be to store viewing angle values as non-dimension coordinates (see my general comment).

eradiate/scenes/measure/_distantsensorarray.py
```diff
@@ -0,0 +1,218 @@
...
+            result.append(d)
+
+        return result
+
+    @property
+    def film_resolution(self) -> Tuple[int, int]:
+        return (1, 1)
+
+
+@measure_factory.register(type_id="distantradiancearray")
```
leroyvn, line 115:
```suggestion
@measure_factory.register(type_id="distant_array_radiance")
```

eradiate/scenes/measure/_distantsensorarray.py
```diff
@@ -0,0 +1,218 @@
...
+@parse_docs
+@attr.s
+class DistantRadianceArrayMeasure(DistantArrayMeasure):
```
leroyvn, line 118:
```suggestion
class DistantArrayRadianceMeasure(DistantArrayMeasure):
```

eradiate/scenes/measure/_distantsensorarray.py
```diff
@@ -0,0 +1,218 @@
...
+    """
+    Distant radiance array measure scene element [``distantradiancearray``].
```
leroyvn, line 120:
```suggestion
    Distant radiance array measure scene element [``distant_array_radiance``].
```

eradiate/scenes/measure/_distantsensorarray.py
```diff
@@ -0,0 +1,218 @@
...
+
+        ds = self._postprocess_add_directions(ds)
+
+        return ds
+
+
+@measure_factory.register(type_id="distantreflectancearray")
+@parse_docs
+@attr.s
+class DistantReflectanceArrayMeasure(DistantArrayMeasure):
```
leroyvn, line 147:
```suggestion
class DistantArrayReflectanceMeasure(DistantArrayMeasure):
```

eradiate/scenes/measure/_distantsensorarray.py
```diff
@@ -0,0 +1,218 @@
...
+            "long_name": "top-of-atmosphere outgoing spectral radiance",
+            "units": symbol(uck.get("radiance")),
+        }
+
+        ds = self._postprocess_add_directions(ds)
+
+        return ds
+
+
+@measure_factory.register(type_id="distantreflectancearray")
```
leroyvn, line 144:
```suggestion
@measure_factory.register(type_id="distant_array_reflectance")
```

eradiate/scenes/measure/_distantsensorarray.py
```diff
@@ -0,0 +1,218 @@
...
+@parse_docs
+@attr.s
+class DistantReflectanceArrayMeasure(DistantArrayMeasure):
+    """
+    Distant reflectance array measure scene element [``distantreflectancearray``].
```
leroyvn, line 149:
```suggestion
    Distant reflectance array measure scene element [``distant_array_reflectance``].
```

eradiate/scenes/measure/_distantsensorarray.py
```diff
@@ -0,0 +1,218 @@
...
+from ..illumination import DirectionalIllumination
+from ...attrs import documented, parse_docs
+from ...units import symbol
+from ...units import unit_context_kernel as uck
+
+
+@measure_factory.register(type_id="distantarray")
+@parse_docs
+@attr.s
+class DistantArrayMeasure(DistantMeasure):
```
leroyvn, line 20: Isn't this class supposed to be an interface definition only? It should not be registered to the factory and should inherit from `ABC` (just to mention explicitly that it is an abstract class).
```suggestion
class DistantArrayMeasure(DistantMeasure, ABC):
```

leroyvn (changes requested):
A few more focused comments. I guess the tests aren't aligned yet (you have a `measure_type` kwarg which I didn't see in your lib code).

leroyvn:
> 
> 
> I have an idea for the averaging, but it does sound a bit hacky: Instead of extracting the relevant results for one sensor, those which should be averaged for spp aggregation, we could create a padded `weights` array, which holds the corresponding spp value for all relevant sensors and 0 otherwise. In that case we could use the xarray mean method and would not have to bother with the other dimensions. 🤔
> 
> EDIT: Creating the padded `weights` is actually very easy with the `xr.where()` method. I am just not sure how to properly merge my results with the Dataset afterwards...

Just saw this: extending the weights array is maybe another solution, I hadn't though about it. I'll let you ponder options, we can discuss this later on ;)

schunkes:
I still haven't addressed your comments but I managed to get the raw data aggregation and assembly of the new Dataset to work :) Would you mind taking a look at it?

eradiate/scenes/measure/_core.py
```diff
@@ -466,20 +465,55 @@ def to_dataset(self, aggregate_spps: bool = False) -> xr.Dataset:
...
+            )
+            # spps themselves must not be scaled, but just masked to include
+            # the right values in the sum
+            weights_spp = xr.where(result.coords["sensor_id"].isin(relevant_ids), 1, 0)
+            spp_aggregated.append(
+                result.data_vars["spp"]
+                .weighted(weights_spp)
+                .sum(dim="sensor_id")
+                .expand_dims({"sensor_id": [id_sensor]}, axis=1)
+            )
```
leroyvn, line 504:
```suggestion

            # Mask spp values so as to only include selected values
            # The sensor_id coordinate is redefined to strip the spp suffix
            weights_spp = xr.where(result.coords["sensor_id"].isin(sensor_ids), 1, 0)
            spp_aggregated.append(
                result.data_vars["spp"]
                .weighted(weights_spp)
                .sum(dim="sensor_id")
                .expand_dims({"sensor_id": [sensor_id_prefix]}, axis=1)
            )
```

eradiate/scenes/measure/_core.py
```diff
@@ -466,20 +465,55 @@ def to_dataset(self, aggregate_spps: bool = False) -> xr.Dataset:
...
+
+        raw_redone = xr.combine_by_coords(raw_aggregated)
+        spp_redone = xr.combine_by_coords(spp_aggregated)
```
leroyvn, line 507:
```suggestion
        raw = xr.combine_by_coords(raw_aggregated)["raw"]
        spp = xr.combine_by_coords(spp_aggregated)["spp"]
```

eradiate/scenes/measure/_core.py
```diff
@@ -466,20 +465,55 @@ def to_dataset(self, aggregate_spps: bool = False) -> xr.Dataset:
...
+
+        result_aggregated = xr.Dataset(
+            data_vars={
+                "raw": raw_redone["raw"],
+                "spp": spp_redone["spp"],
+                "sensor_id": list(measure_ids_sensor),
+            }
```
leroyvn, line 514:
```suggestion
        result_aggregated = xr.Dataset(
            data_vars={
                "raw": raw,
                "spp": spp,
                "sensor_id": sensor_id_prefixes,
            }
```
(`sorted()` already returns a list)

leroyvn (commented):
Hi @schunkes, I made suggestions to clarify the logic of your aggregation code (I hope I understood correctly!). It's quite comprehensive IMO, so we can go with that.

schunkes:
Here we go, I included your feedback and I added a *from_angles* constructor.

leroyvn (changes requested):
Hi @schunkes, that's nice. I have a couple of comments. 

I also would like you to make some changes to the aggregation code (if not the ones I suggested, something in a similar spirit) because I'm quite sure that if I go through that code in 6 months, it will take me much more effort to remember the logics than if variables are named more consistently and comments a little more detailed.

eradiate/scenes/measure/_distant_array.py
```diff
@@ -0,0 +1,221 @@
...
+
+from ._core import SensorInfo, measure_factory
+from ._distant import DistantMeasure
+from ..illumination import DirectionalIllumination
+from ...attrs import documented, parse_docs
+from ...units import symbol
+from ...units import unit_context_kernel as uck
+
+
+@measure_factory.register(type_id="distant_array")
```
leroyvn, line 18: This is an ABC: it's not supposed to be registered to the factory.
```suggestion
```

eradiate/scenes/measure/_distant_array.py
```diff
@@ -0,0 +1,221 @@
...
+@parse_docs
+@attr.s
+class DistantArrayMeasure(DistantMeasure, ABC):
+    """
+    Distant sensor array measure scene element [``distantarray``].
+
+    This measure scene element is a thin wrapper around the ``DistantSensor``
+    measures. It records either the radiance or reflectance along the directions given by the
+    ``directions`` parameter.
```
leroyvn, line 27: This class technically doesn't: only its concrete child classes do.

eradiate/scenes/measure/_distant_array.py
```diff
@@ -0,0 +1,221 @@
...
+            result.append(d)
+
+        return result
+
+    @property
+    def film_resolution(self) -> Tuple[int, int]:
+        return (1, 1)
+
+
+@measure_factory.register(type_id="distant_array_radiance")
```
leroyvn, line 118: Let's add an alias:
```suggestion
@measure_factory.register(type_id="distant_array")
@measure_factory.register(type_id="distant_array_radiance")
```

eradiate/scenes/measure/_distant_array.py
```diff
@@ -0,0 +1,221 @@
...
+@parse_docs
+@attr.s
+class DistantArrayRadianceMeasure(DistantArrayMeasure):
+    """
+    Distant array radiance measure scene element [``distant_array_radiance``].
```
leroyvn, line 123:
```suggestion
    Distant array radiance measure scene element [``distant_array``, ``distant_array_radiance``].
```

eradiate/scenes/measure/_distant_array.py
```diff
@@ -0,0 +1,221 @@
...
+
+    This measure scene element is a thin wrapper around the ``DistantSensor``
+    measures. It records either the radiance or reflectance along the directions given by the
+    ``directions`` parameter.
```
leroyvn, line 127:
```suggestion
    This measure scene element is a thin wrapper around the ``DistantSensor``
    measures. It records radiance along the directions given by the ``directions`` 
    parameter.
```

eradiate/scenes/measure/_distant_array.py
```diff
@@ -0,0 +1,221 @@
...
+@measure_factory.register(type_id="distant_array_reflectance")
+@parse_docs
+@attr.s
+class DistantArrayReflectanceMeasure(DistantArrayMeasure):
+    """
+    Distant array reflectance measure scene element [``distant_array_reflectance``].
+
+    This measure scene element is a thin wrapper around the ``DistantSensor``
+    measures. It records either the radiance or reflectance along the directions given by the
+    ``directions`` parameter.
```
leroyvn, line 156:
```suggestion
    This measure scene element is a thin wrapper around the ``DistantSensor``
    measures. It records the radiance and reflectance along the directions given by the
    ``directions`` parameter.
```

schunkes:
Alright, take another look. I think the dict interface is good, the keywords feel natural and fitting to me.

eradiate/scenes/measure/_core.py
```diff
@@ -466,20 +465,58 @@ def to_dataset(self, aggregate_spps: bool = False) -> xr.Dataset:
...
+            data_vars={
+                "raw": raw,
+                "spp": spp,
+                "sensor_id": sensor_id_prefixes,
+            }
         )
-        result_aggregated = result_aggregated.drop_dims("sensor_id")
 
-        # Copy metadata (lost during aggregation)
         for var in list(result_aggregated.data_vars) + list(result_aggregated.coords):
```
leroyvn, line 520:
```suggestion
        # Copy metadata
        for var in list(result_aggregated.data_vars) + list(result_aggregated.coords):
```

PR #120, opened 2021-09-20
Molecular absorption cross section computer using hapi

nollety:
Resolves eradiate/eradiate-issues#110

# Description

Adds a wrapper around [hapi](https://github.com/hitranonline/hapi) to compute molecular absorption cross section spectra dynamically.

# Checklist

- [ ] add `hapi` to the dependencies (build and upload to anaconda first)
- [ ] cite HAPI
- [ ] better docstrings
- [ ] add module to documentation

### DIFF
--- a/eradiate/radprops/sad.py
+++ b/eradiate/radprops/sad.py
@@ -1,0 +1,34 @@
+"""
+HAPI wrapper.
+"""
+
+import hapi
+import pint
+import xarray as xr
+
+# TODO: parse https://hitran.org/docs/iso-meta/ to automatically update isotopologues and corresponding abundances
+ISOTOPOLOGUES = dict(
+    H2O=[1, 2, 3, 4, 5, 6, 129],
+    CO2=[7, 8, 9, 10, 11, 12, 13, 14, 121, 15, 120, 122],
+    O3=[16, 17, 18, 19, 20],
+)
+
+
+def fetch_table(
+    molecule: str, wavelength_min: pint.Quantity, wavelength_max: pint.Quantity
+) -> None:
+    hapi.fetch_by_ids(
+        TableName="auto",
+        iso_id_list=ISOTOPOLOGUES[molecule],
+        numin=(1 / wavelength_max).m_as("cm^-1"),
+        numax=(1 / wavelength_min).m_as("cm^-1"),
+    )
+
+
+def compute_absorption_cross_section(
+    molecule: str, wavelength_min: pint.Quantity, wavelength_max: pint.Quantity
+) -> xr.DataArray:
+    fetch_table(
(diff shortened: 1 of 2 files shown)


nollety:
Quick demo:

```python
da = compute_absorption_cross_section(
    molecule="CO2",
    wavenumber_min=ureg.Quantity(5900, "cm^-1"),
    wavenumber_max=ureg.Quantity(5910, "cm^-1"),
    wavenumber_step=ureg.Quantity(0.01, "cm^-1"),
    pressure=ureg.Quantity([1.0, 0.6], "atm"),
    temperature=ureg.Quantity(280.0, "K"),
    hitran_data_dir="./",
    to_wavelength=False
)
da.plot(hue="p", linewidth=.5, yscale="log", figsize=(16,8))
```

The above code produces the figure below
output

nollety:
Closed because inactive

zxdawn:
Looks interesting! Any idea of the progress of adding HITRAN?

leroyvn:
Hi @zxdawn, a lot has happened since this issue was closed. We have decided to not support spectrum computation libraries directly because accurate molecular absorption modelling actually requires aggregating data from multiple sources. Instead, we provide a documented dataset format for both the line-by-line and correlated-k band modes. We provide default datasets, but users can also provide their if they wish. See the molecular absorption data documentation [here](https://eradiate.readthedocs.io/en/stable/rst/user_guide/data/absorption.html).

zxdawn:
Thanks @leroyvn . I suppose if I use hapi2 to generate a similar NC file, it should work.

BTW, how did you calculate the absorption cross section? I didn't find it in the Guide. (I wanna use the absorption cross section data to generate my own spectral response function.)

leroyvn:
Are you looking to generate CKD or LBL datasets?

We use the [Radis library](https://radis.readthedocs.io/en/latest/) to generate line-by-line spectra (its algorithm is rather fast and the interface is relatively simple), then incorporate additional datasets to account for effects that are not covered by the HITRAN LBL database (e.g. ozone photodissociation or collision induced absorption of certain species). Note that Eradiate works with absorption coefficient datasets, which factor in species number densities and are indexed against atmospheric profile state variables (pressure, temperature, species concentrations).

One thing to note: we are currently working on a new and extended CKD dataset with more species.

List thing: I don't understand the relationship between the (sensor?) spectral response function and the absorption cross-section, maybe you want to elaborate a bit?

zxdawn:
Ha, thanks! The Radis library is really useful.

Sorry for the confusion. I plan to use the LBL dataset to calculate absorption cross-section and then simulate spectral radiance. SRF is applied later to convert the high-resolution spectral radiance to low-resolution sensor signal.

PR #125, opened 2021-09-22
Distant array measure bugfix

schunkes:
# Description

This PR introduces a series of bugfixes for the `DistantSensorArray`.

First of all sensor id handling was updated. Sorting the sensor ids as strings produces a wrong order, since "100" comes before "2". Therefore spp aggregation is now handled by converting the sensor id's numerical part into an integer and introducing this as a temporary coordinate into the DataArrays. After spp aggregation, this temporary coordinate is used to correctly arrange the results in the Dataset, instead of the string sensor id.

The major issue I have with this approach is, that I had to alter the default `sensor_infos` method. This now adds a "ms0" suffix to the sensor ids, even if the `DistantMeasure` derivatives are used, which use only one mitsuba sensor plugin by default.

This needs to be made more robust. It crashes a lot of tests: The postprocessing tests inject raw results with sensor ids, that do not have a "msXXX" suffix.

Also the direction computation was updated. It turns out, that setting the direction to `[0, 0, 1]` will create a sensor that records the radiance leaving towards that direction, instead of creating a sensor that would point towards this direction and "look at the scene from below".

Finally I added a test case that assures the correctness of the generated directions in the `azimuthal_ring` constructor.

# Checklist

### DIFF
--- a/eradiate/scenes/measure/_core.py
+++ b/eradiate/scenes/measure/_core.py
@@ -18,7 +18,7 @@ from ..core import KernelDict, SceneElement
 from ... import ckd, converters, validators
 from ..._factory import Factory
 from ..._mode import ModeFlags
-from ..._util import ensure_array, natsort_alphanum_key
+from ..._util import deduplicate, ensure_array, natsort_alphanum_key
 from ...attrs import AUTO, AutoType, documented, get_doc, parse_docs
 from ...ckd import Bin
 from ...contexts import (
@@ -499,14 +499,11 @@ class MeasureResults:
         if not aggregate_spps:
             return result
 
-        # Else ...
-        sensor_id_prefixes = sorted(
-            set(
-                [
-                    sensor_id.split("_spp")[0]
-                    for sensor_id in list(result["sensor_id"].values)
-                ]
-            )
+        sensor_id_prefixes = deduplicate(
+            [
+                sensor_id.split("_spp")[0]
+                for sensor_id in list(result["sensor_id"].values)
+            ]
         )
 
         raw_aggregated = []
@@ -516,7 +513,7 @@ class MeasureResults:
             sensor_ids = [
                 id
                 for id in result.coords["sensor_id"].values
-                if id.startswith(sensor_id_prefix)
+                if id.split("_spp")[0] == sensor_id_prefix
             ]
             # Aggregate sample counts for the current prefix
             weights = xr.where(
@@ -528,7 +525,10 @@ class MeasureResults:
                 result["raw"]
                 .weighted(weights)
                 .mean(dim="sensor_id")
-                .expand_dims({"sensor_id": [sensor_id_prefix]}, axis=1)
+                .expand_dims(
+                    {"sensor_id": [sensor_id_prefix]},
+                    axis=1,
+                )
                 .to_dataset()
             )
             # Mask spp values so as to only include selected values
@@ -538,12 +538,16 @@ class MeasureResults:
                 result.data_vars["spp"]
                 .weighted(weights_spp)
                 .sum(dim="sensor_id")
-                .expand_dims({"sensor_id": [sensor_id_prefix]}, axis=1)
+                .expand_dims(
+                    {"sensor_id": [sensor_id_prefix]},
+                    axis=1,
+                )
                 .to_dataset()
             )
 
-        raw = xr.combine_by_coords(raw_aggregated)["raw"]
-        spp = xr.combine_by_coords(spp_aggregated)["spp"]
+        raw = xr.concat(raw_aggregated, "sensor_id")["raw"]
+
+        spp = xr.concat(spp_aggregated, "sensor_id")["spp"]
 
         result_aggregated = xr.Dataset(
             data_vars={
@@ -611,13 +615,14 @@ class MeasureResults:
             Sensor film size as a (int, int) pair.
         """
         spectral_coords = set()
-        sensor_ids = set()
+        sensor_ids = list()
         film_size = np.zeros((2,), dtype=int)
 
         for spectral_coord, val in raw.items():
             spectral_coords.add(spectral_coord)
             for sensor_id, data in val["values"].items():
-                sensor_ids.add(sensor_id)
+                if sensor_id not in sensor_ids:
+                    sensor_ids.append(sensor_id)
                 film_size = np.maximum(film_size, data.shape[:2])
 
         if eradiate.mode().has_flags(ModeFlags.ANY_CKD):
@@ -862,11 +867,11 @@ class Measure(SceneElement, ABC):
         spps = self._split_spp()
 
         if len(spps) == 1:
-            return [SensorInfo(id=f"{self.id}", spp=spps[0])]
+            return [SensorInfo(id=f"{self.id}_ms0", spp=spps[0])]
 
         else:
             return [
-                SensorInfo(id=f"{self.id}_spp{i}", spp=spp)
+                SensorInfo(id=f"{self.id}_ms0_spp{i}", spp=spp)
                 for i, spp in enumerate(spps)
             ]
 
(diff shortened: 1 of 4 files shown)


eradiate/scenes/measure/_core.py
```diff
@@ -510,7 +509,10 @@ def to_dataset(self, aggregate_spps: bool = False) -> xr.Dataset:
                 result["raw"]
                 .weighted(weights)
                 .mean(dim="sensor_id")
-                .expand_dims({"sensor_id": [sensor_id_prefix]}, axis=1)
+                .expand_dims(
+                    {"sensor_id": [int(sensor_id_prefix.split("ms")[1])]},
```
schunkes, line 513: This is the place in the code, that I was talking about. Here I expect the measure id to be of the form "something_ms0" and I split the string at "ms" to get the numerical part of the ID. I do this because I cannot tell `xr.combine_by_coords()` to use the alphanumerical sorting method you added. 

I could instead compute this number further up in the code. Basically at the beginning of the loop and perform a check: If `sensor_id_prefix.split("ms")[1]`  is not an empty string, use it and cast it to int, otherwise just use the `sensor_id_prefix` as is. What do you think?

leroyvn:
I was thinking about this. I see a few possible options, not necessarily incompatible with each other.

1. Make the ordering more robust by padding ID indexes with zeros (this should be done by the `sensor_infos()` method). This is clean IMO and should not require much effort, and you can simply not use natural order sorting.
2. Use a lower-level concatenation routine which won't reorder results. This is also safe IMO, since you actually don't want this step to mess with ordering (if I understand correctly).
3. Further refactor the sensor ID generation and low-level post-processing to eventually create one per suffix, indexed by integer values.

eradiate/scenes/measure/_core.py
```diff
@@ -481,15 +481,14 @@ def to_dataset(self, aggregate_spps: bool = False) -> xr.Dataset:
...
-            set(
-                [
-                    sensor_id.split("_spp")[0]
-                    for sensor_id in list(result["sensor_id"].values)
-                ]
-            )
-        )
+        # https://stackoverflow.com/questions/480214/how-do-you-remove-duplicates-from-a-list-whilst-preserving-order
+        seen = set()
+        seen_add = seen.add
```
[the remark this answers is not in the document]
schunkes: This is in the solution posted in the stackoverflow post: The author says it speeds up execution. I have not tested that though.

leroyvn:
> Also the direction computation was updated. It turns out, that setting the direction to [0, 0, 1] will create a sensor that records the radiance leaving towards that direction, instead of creating a sensor that would point towards this direction and "look at the scene from below".

Yeah that's a side effect of merging the `distant` and `hdistant` sensors back then. Another option is to use the `flip_directions` plugin parameter to have directions point towards the target location.

eradiate/scenes/measure/_core.py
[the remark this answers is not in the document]
leroyvn: Fair enough (although the frequency with which we'll use this routine probably doesn't justify saving on method derefencing xD). Now that I think about it, Python actually has always had an ordered dict on which you can rely to implement such order-preserving deduplication. I'll commit a utility function with options.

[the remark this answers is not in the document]
leroyvn: I just pushed an update to `main`, you can rebase to get it.

schunkes:
> I was thinking about this. I see a few possible options, not necessarily incompatible with each other.
> 
>     1. Make the ordering more robust by padding ID indexes with zeros (this should be done by the `sensor_infos()` method). This is clean IMO and should not require much effort, and you can simply not use natural order sorting.
> 
>     2. Use a lower-level concatenation routine which won't reorder results. This is also safe IMO, since you actually don't want this step to mess with ordering (if I understand correctly).
> 
>     3. Further refactor the sensor ID generation and low-level post-processing to eventually create one per suffix, indexed by integer values. This is slighlty more involved but opens the door to arbitrarily complex multi-sensor setups, with interesting post-processing workflows which can exploit dimension detection.
> 
> 
> All in all, I'd say 1 is a quick, easy and clean fix; 2 may be desirable to avoid suprises in the future; 3 should be on our roadmap for future improvement.

About option one: I agree that it would be easiest to pad sensor IDs with leading zeros, but that would still leave me with being forced to modify the `core.sensor_infos()' method, which I think is messy.

About option two: You mean, instead of using `combine_by_coords()' I should assemble the DataArray manually? 

About option three: I think I have a hard time imagining a generalized framework for this, so I think I'd keep this in memory and reconsider if we encounter more cases that warrant suffixes to sensor ids.

leroyvn:
> About option one: I agree that it would be easiest to pad sensor IDs with leading zeros, but that would still leave me with being forced to modify the core.sensor_infos() method, which I think is messy.

I don't get it, wouldn't you implement this padding policy as part of `Measure.sensor_infos()`?

> About option two: You mean, instead of using combine_by_coords() I should assemble the DataArray manually?

Tell me if I'm wrong: you want to concatenate data arrays along a single dimension and preserve ordering, right? In that case,  `concat()` is the tool to go for, no need to use `combine_by_coords()` and lose control on data array ordering.

> About option three: I think I have a hard time imagining a generalized framework for this, so I think I'd keep this in memory and reconsider if we encounter more cases that warrant suffixes to sensor ids.

Each `Measure` subclass can have its own set of sensor dimensions (`spp` at least, possibly `ms`, etc.). This defines the number of dimensions the data array will have, the structure of sensor IDs, and even allows for a generic implementation of aggregation pipelines.

schunkes:
I was thinking that I'd prefer not to change the `core.sensor_infos()` method if only the specialized class actually needs the change.

On the other hand, I am changing the `core.to_dataset()` method, so it really doesn't matter.

I think I will adopt things from your suggestions one and two: Add the proper padding to sensor ids and use xarrays `concat` to combine the DataArrays.

eradiate/scenes/measure/_core.py
```diff
@@ -481,15 +481,17 @@ def to_dataset(self, aggregate_spps: bool = False) -> xr.Dataset:
...
+        # seen_add = seen.add
+        sensor_id_prefixes = [
+            sensor_id.split("_spp")[0]
+            for sensor_id in list(result["sensor_id"].values)
+            if not (
+                sensor_id.split("_spp")[0] in seen
+                or seen.add(sensor_id.split("_spp")[0])
             )
-        )
+        ]
```
leroyvn, line 494: After rebasing, you should change this to
```
sensor_id_prefixes = deduplicate([
    sensor_id.split("_spp")[0] for sensor_id in list(result["sensor_id"].values)
])
```

eradiate/scenes/measure/tests/test_measure_core.py
```diff
@@ -96,6 +96,7 @@ def test_measure_results(mode_mono):
     ds = results.to_dataset(aggregate_spps=True)
     assert set(ds.data_vars) == {"raw", "spp"}
     assert set(ds.coords) == {"sensor_id", "w", "x", "y"}
+    print(ds["sensor_id"])
```
leroyvn, line 99:
```suggestion
```

leroyvn (changes requested):
Hi @schunkes, that should do for now. Now I think I get what you meant: You didn't want the parent `Measure.sensor_infos()` to emit a `ms` suffix if the measure is not multi-sensor, right? I think indeed allow for all suffixes to be optional requires a couple of architectural changes. I'll create an issue for that.

I have a couple of minor comments in addition, nothing much.

eradiate/scenes/measure/tests/test_measure_core.py
schunkes: They always slip through -.-

schunkes:
> Hi @schunkes, that should do for now. Now I think I get what you meant: You didn't want the parent `Measure.sensor_infos()` to emit a `ms` suffix if the measure is not multi-sensor, right? I think indeed allow for all suffixes to be optional requires a couple of architectural changes. I'll create an issue for that.
> 

Yes that's exactly my thought :)

PR #127, opened 2021-09-23
Added a check to KernelDict.__setitem__(): The method will now issue …

schunkes:
…a warning if a scene element ID appears more than once in a KernelDict.

closes eradiate/eradiate-issues#89

# Description
A generic warning is raised, mentioning the id in question and stating that scene elements will be overwritten.

Writing the tests I found, that this anyways applies only if you use `KernelDict.add()`. If you create one large dictionary with all your scene elements manually, their IDs will be different, because the are the keys in this manually created dict.

# Checklist

### DIFF
--- a/eradiate/scenes/core.py
+++ b/eradiate/scenes/core.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import typing as t
+import warnings
 from abc import ABC, abstractmethod
 from collections import abc as collections_abc
 
@@ -79,6 +80,13 @@ class KernelDict(collections_abc.MutableMapping):
         return self.data.__iter__()
 
     def __setitem__(self, k, v):
+        try:
+            self.data.__getitem__(k)
+            warnings.warn(
+                f"Duplicate key '{k}' will be overwritten. Are you trying to add scene elements with duplicate IDs to this KernelDict?"
+            )
+        except KeyError:
+            pass
         return self.data.__setitem__(k, v)
 
     def check(self) -> None:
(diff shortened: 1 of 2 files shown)


schunkes:
Two points for discussion:

Should I add a custom warning type to be more specific in the tests?
Is the phrasing of the warning okay?

eradiate/scenes/core.py
```diff
@@ -79,6 +80,13 @@ def __iter__(self):
         return self.data.__iter__()
 
     def __setitem__(self, k, v):
+        try:
+            self.data.__getitem__(k)
+            warnings.warn(
+                f"Duplicate scene element id '{k}' found. Scene elements will be overwritten."
+            )
+        except KeyError:
+            pass
```
leroyvn, line 89: I think a simple presence check would be more legible. Also, you're not setting scene element IDs here, so I'd make the message more general:
```suggestion
        if k in self.data:
            warnings.warn(f"Duplicate key '{k}' will be overwritten.")
        return self.data.__setitem__(k, v)
```

leroyvn (commented):
Hi @schunkes, thanks for this patch. I just have a small comment. The key point here is that `KernelDict` does operates at the kernel level and is basically a supercharged dict. It therefore doesn't know about scene elements or IDs, it only sees keys and values.

eradiate/scenes/core.py
schunkes: Well but the second message does not tell the user anything about where this issue comes from. If you don't know how the `KernelDict` works, you have no idea how to fix this. :(

leroyvn: Okay, so how about something like this:
```
Duplicate key '{k}' will be overwritten. Are you trying to add scene elements with duplicate IDs to this KernelDict?
```
This points the user to the potential issue without being definitive.

schunkes: That's nice :)

leroyvn:
Let's go! Thanks @schunkes.

PR #129, opened 2021-09-27
Fixed tests that depend on datasets to download

nollety:
# Description

A few tests were relying on the fact that the absorption data sets for the `us76_u86_4` absorber have been downloaded into the `resources/data/spectra/absorption` directory, which is not the case after a fresh install, and should not be required by the tests suite. These tests were refactored to use the dedicated absorption tests data sets.

# Checklist


leroyvn (approved):
Hi @nollety, thanks for fixing this. Looks good to me, I'll merge this now.

PR #131, opened 2021-10-05
Render mode

schunkes:
# Description

We can already discuss this code, but the spectra are still giving me troubles. I can't instantiate sRGB spectra directly, so I can't test them properly. There is an open issue in the mitsuba issue tracker mitsuba-renderer/mitsuba2#506

Points to discuss:

- measure/_core.py line 472: What is the unit of the bindex?
- Media don't support RGB spectra, but since the RGB mode is for qualitative renders only, I figured it would be okay to pick a monochromatic value in the middle of the visible spectrum for the medium parameters. Do you agree?

# Checklist


eradiate/scenes/measure/_core.py
```diff
@@ -443,8 +459,29 @@ def to_dataset(self, aggregate_spps: bool = False) -> xr.Dataset:
         if not self.raw:
             raise ValueError("no raw results to convert to xarray.Dataset")
 
-        # Collect spectral coordinate label
-        spectral_coord_label = eradiate.mode().spectral_coord_label
+        if eradiate.mode().has_flags("ANY_MONO"):
```
leroyvn, line 462:
```suggestion
        if eradiate.mode().has_flags(ModeFlags.ANY_MONO):
```

eradiate/scenes/measure/_core.py
```diff
@@ -443,8 +459,29 @@ def to_dataset(self, aggregate_spps: bool = False) -> xr.Dataset:
...
+            spectral_coord_label = eradiate.mode().spectral_coord_label
+            spectral_coord_metadata = {
+                "long_name": "wavelength",
+                "units": symbol(ucc.get("wavelength")),
+            }
+        elif eradiate.mode().has_flags("ANY_CKD"):
```
leroyvn, line 468:
```suggestion
        elif eradiate.mode().has_flags(ModeFlags.ANY_CKD):
```

eradiate/scenes/measure/_core.py
```diff
@@ -443,8 +459,29 @@ def to_dataset(self, aggregate_spps: bool = False) -> xr.Dataset:
...
+            spectral_coord_label = eradiate.mode().spectral_coord_label
+            spectral_coord_metadata = {
+                "long_name": "bindex",
+                "units": symbol(ucc.get("wavelength")),  # I don't know what to put here
+            }
+        elif eradiate.mode().has_flags("ANY_RGB"):
```
leroyvn, line 474:
```suggestion
        elif eradiate.mode().has_flags(ModeFlags.ANY_RGB):
```

eradiate/scenes/measure/_core.py
```diff
@@ -443,8 +459,29 @@ def to_dataset(self, aggregate_spps: bool = False) -> xr.Dataset:
...
+            spectral_coord_label = eradiate.mode().spectral_coord_label
+            spectral_coord_metadata = {
+                "long_name": "wavelength",
+                "units": symbol(ucc.get("wavelength")),
+            }
+        elif eradiate.mode().has_flags("ANY_CKD"):
+            spectral_coord_label = eradiate.mode().spectral_coord_label
+            spectral_coord_metadata = {
+                "long_name": "bindex",
+                "units": symbol(ucc.get("wavelength")),  # I don't know what to put here
```
leroyvn, line 472:
```suggestion
```
It doesn't have units so you can just drop this field.

eradiate/scenes/measure/_core.py
```diff
@@ -465,36 +502,78 @@ def to_dataset(self, aggregate_spps: bool = False) -> xr.Dataset:
...
+                        {"long_name": "color channel"},
                     ),
-                    "spp": (
-                        [spectral_coord_label, "sensor_id"],
-                        spps,
-                        {"long_name": "sample count"},
+                    spectral_coord_label: (
+                        spectral_coord_label,
+                        spectral_index,
+                        self._to_dataset_spectral_coord_metadata(),
```
leroyvn, line 532: This dimension should be scalar: do you actually need it at all?

eradiate/scenes/measure/_core.py
```diff
@@ -465,36 +502,78 @@ def to_dataset(self, aggregate_spps: bool = False) -> xr.Dataset:
...
+                    ),
+                    "sensor_id": (
+                        "sensor_id",
+                        sensor_ids,
+                        {"long_name": "sensor ID"},
+                    ),
+                },
+            )
```
leroyvn, line 576: I'll simplify/streamline that as part of the upcoming post-processing upgrade. In the meantime, we'll keep this conditional.

eradiate/scenes/measure/_core.py
```diff
@@ -585,7 +664,8 @@ def _to_dataset_spectral_coord_metadata() -> t.Dict:
 
         elif eradiate.mode().has_flags(ModeFlags.ANY_CKD):
             return {"long_name": "bindex"}
-
+        elif eradiate.mode().has_flags(ModeFlags.ANY_RGB):
+            return {"long_name": "red, green, blue"}
         else:
             raise UnsupportedModeError(supported=("monochromatic", "ckd"))
```
leroyvn, line 670:
```suggestion
            raise UnsupportedModeError(supported=("monochromatic", "ckd", "rgb"))
```

eradiate/scenes/measure/_core.py
```diff
@@ -672,29 +752,49 @@ def _to_dataset_helper_data_values(
...
-                # array: this works with mono variants but will fail otherwise
+            for i_spectral, spectral_coord in enumerate(spectral_coords):
+                for i_sensor, sensor_id in enumerate(sensor_ids):
+                    # Note: This doesn't handle heterogeneous sensor film sizes
+                    # (i.e. cases in which sensors have different film sizes).
+                    # To add support for it, blitting is probably a good approach
+                    # https://stackoverflow.com/questions/28676187/numpy-blit-copy-part-of-an-array-to-another-one-with-a-different-size
+                    data[i_spectral, i_sensor, ..., 0] = raw[spectral_coord]["values"][
+                        sensor_id
+                    ][..., 0]
```
leroyvn, line 775:
```suggestion
                    ][..., 0]
                    # This latter indexing selects only one channel in the raw data
                    # array: this works with mono variants but will fail otherwise

```
Please keep the comment, it's useful to understand what this complicated indexing does.

eradiate/scenes/measure/_core.py
```diff
@@ -672,29 +752,49 @@ def _to_dataset_helper_data_values(
...
+            for i_spectral, spectral_coord in enumerate(spectral_coords):
+                for i_sensor, sensor_id in enumerate(sensor_ids):
+                    # Note: This doesn't handle heterogeneous sensor film sizes
+                    # (i.e. cases in which sensors have different film sizes).
+                    # To add support for it, blitting is probably a good approach
+                    # https://stackoverflow.com/questions/28676187/numpy-blit-copy-part-of-an-array-to-another-one-with-a-different-size
+                    data[i_spectral, i_sensor] = raw[spectral_coord]["values"][
+                        sensor_id
+                    ]
+        else:
```
leroyvn, line 796:
```suggestion

        else:
```

eradiate/scenes/measure/_radiancemeterarray.py
```diff
@@ -66,6 +69,10 @@ def _target_origin_validator(self, attribute, value):
                 f"directions.shape = {self.directions.shape}"
             )
 
+    def __attrs_post_init__(self):
```
leroyvn, line 72:
```suggestion
    def __attrs_pre_init__(self):
```

eradiate/scenes/measure/_radiancemeter.py
```diff
@@ -67,6 +70,10 @@ def _target_origin_validator(self, attribute, value):
     def film_resolution(self) -> t.Tuple[int, int]:
         return (1, 1)
 
+    def __attrs_post_init__(self):
```
leroyvn, line 73:
```suggestion
    def __attrs_pre_init__(self):
```
No need to bother constructing the object if it's not allowed.

eradiate/scenes/spectra/_rgb.py
```diff
@@ -0,0 +1,77 @@
...
+from ... import unit_context_kernel as uck
+from ...attrs import documented, parse_docs
+from ...contexts import KernelDictContext, SpectralContext
+from ...exceptions import UnsupportedModeError
+
+
+@spectrum_factory.register(type_id="rgb")
+@parse_docs
+@attr.s
+class RGBSpectrum(Spectrum):
```
leroyvn, line 20: We should not have the option to introduce a spectrum usable only in one mode: instead, the existing spectrum types should emit a kernel dict compatible with the RGB variant, so that an existing scene can be dropped in a render setup without change.

eradiate/scenes/spectra/_solar_irradiance.py
```diff
@@ -116,6 +119,17 @@ def eval_mono(self, w: pint.Quantity) -> pint.Quantity:
 
         return irradiance
 
+    def eval_rgb(self) -> pint.Quantity:
+        w_units = ureg(self.data.ssi.w.attrs["units"])
+        wavelengths = ([600, 500, 400] * ureg.nm).m_as(w_units)
+        irradiance = to_quantity(self.data.ssi.interp(w=wavelengths, method="linear"))
```
leroyvn, line 125: This has a high chance of producing irrelevant data due to the fairly high-frequency features of the solar irradiance spectrum. While implementing an appropriate method to compute the RGB values accurately is straightforward, I understand that it's a bit of work and can be part of a separate issue/PR. However, I'd rather have this guy raise or, at least, issue a warning.

leroyvn (commented):
Hi @schunkes, I had a quick look and this looks overall in line with what we discussed. A couple of general comments:

- Mode-specific spectrum classes should be avoided. This means that the `RGBSpectrum` class should either work in all modes or not exist at all.
- The bindex has no units (note the difference with "dimensionless") and can therefore not have a `units` metadata field.
- Volumes actually have an RGB interface: when you build a volume file, it will be interpreted as RGB or monochromatic based on its shape. This can however be part of a dedicated PR, you can raise a `NotImplementedError` for now.

eradiate/scenes/spectra/_rgb.py
schunkes: At this point I am not sure if this was introduced by me or if this is an artifact from an old version of Eradiate :thinking: Anyway, I'll remove it.

schunkes:
>     * Volumes actually have an RGB interface: when you build a volume file, it will be interpreted as RGB or monochromatic based on its shape. This can however be part of a dedicated PR, you can raise a `NotImplementedError` for now.

Maybe I was not using the right words. The "sigma_a" and "sigma_s" spectra in our atmosphere classes produce errors, when I set them to RGB values. Rather instantiating a kernel object from our atmosphere classes raises an error, saying that these spectra have the wrong type. If this should actually be possible, I will take another look.

eradiate/scenes/measure/_core.py
leroyvn: Fair point (although I'm considering switching to well-specified dynamic dimensions, but that's to be left for the post-processing pipeline refactoring).

eradiate/scenes/spectra/_solar_irradiance.py
schunkes: What approach would you suggest for the RGB values? Or rather, do you have a reference document or something?

leroyvn:
@schunkes I think we should close this: the addition of the basic tone mapping operator makes this feature no longer required for visualisation. Do you agree?

PR #134, opened 2021-10-18
Rescale concentration within CKD mode

nollety:
# Description

Updated CKD data sets have a `H2O` and a `O3` coordinate that allow to select/interpolate the absorption coefficient cumulative-probability quadrature data set at different H2O/O3 concentration.
This pull request updates the components of eradiate that interact with these data sets so that they are compatible with the presence of these new `H2O` and `O3` coordinates.

⚠️  Requires #136 to be merged beforehand.

⚠️ Before merging, upload updated data sets to FTP server.

# Checklist

- [x] ckd data sets for the `10nm_test` bin set
- [x] ckd data sets for the `10nm` bin set

### DIFF
--- a/eradiate/radprops/rad_profile.py
+++ b/eradiate/radprops/rad_profile.py
@@ -28,6 +28,10 @@ from ..contexts import SpectralContext
 from ..data.absorption_spectra import Absorber, Engine, find_dataset
 from ..exceptions import UnsupportedModeError
 from ..thermoprops import afgl1986, us76
+from ..thermoprops.util import (
+    compute_column_mass_density,
+    compute_column_number_density,
+)
 from ..units import to_quantity
 from ..units import unit_context_config as ucc
 from ..units import unit_registry as ureg
@@ -820,8 +824,8 @@ class US76ApproxRadProfile(RadProfile):
         ).to("dimensionless")
 
     def eval_albedo_ckd(self, *bindexes: Bindex) -> pint.Quantity:
-        sigma_s = self.eval_sigma_s_ckd(bindexes=bindexes)
-        sigma_t = self.eval_sigma_t_ckd(bindexes=bindexes)
+        sigma_s = self.eval_sigma_s_ckd(*bindexes)
+        sigma_t = self.eval_sigma_t_ckd(*bindexes)
         return np.divide(
             sigma_s, sigma_t, where=sigma_t != 0.0, out=np.zeros_like(sigma_s)
         ).to("dimensionless")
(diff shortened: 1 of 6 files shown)


eradiate/radprops/rad_profile.py
```diff
@@ -1083,11 +1087,24 @@ def eval_sigma_a_ckd(self, *bindexes: Bindex) -> pint.Quantity:
...
+
         z = to_quantity(self.thermoprops.z_layer).m_as(ds.z.units)
         return ureg.Quantity(
             [
                 ds.k.sel(bd=(bindex.bin.id, bindex.index))
-                .interp(z=z, kwargs=dict(fill_value=0.0))
+                .interp(
+                    z=z,
+                    H2O=h2o_concentration.m_as(ds.H2O.units),
+                    O3=o3_concentration.m_as(ds.O3.units),
```
leroyvn, line 1105: I'd rather favour the dict syntax to access dataset members [since it seems to be the most basic way to do so](http://xarray.pydata.org/en/stable/user-guide/data-structures.html#dataset-contents).

```suggestion
                    H2O=h2o_concentration.m_as(ds["H2O"].units),
                    O3=o3_concentration.m_as(ds["O3"].units),
```

eradiate/thermoprops/afgl1986.py
```diff
@@ -25,7 +28,7 @@ def make_profile(
     model_id : {"us_standard", "midlatitude_summer", "midlatitude_winter", "subarctic_summer", "subarctic_winter", "tropical"}, default: "us_standard"
         Model identifier.
 
-    levels : quantity or array
+    levels : quantity, optional
```
leroyvn, line 31: If both scalars and arrays are allowed, it would be nice to mention it.

eradiate/thermoprops/afgl1986.py
```diff
@@ -92,7 +95,19 @@ def make_profile(
     concentration units, refer to the documentation of
     :func:`~eradiate.thermoprops.util.compute_scaling_factors`.
     """
+    if eradiate.mode().has_flags(ModeFlags.ANY_CKD):
+        if model_id != "us_standard":
+            raise NotImplementedError(
+                "In CKD mode, only the 'us_standard' model is supported."
+            )
+        if concentrations is not None:
+            if any([molecules not in ["H2O", "O3"] for molecules in concentrations]):
```
leroyvn, line 104: A nicer and efficient way to do so it to use sets. I think something like this should do:
```python
molecules = set(concentrations.keys()) if concentrations else set()

if molecules - {"H2O", "O3"}:
    raise ...
```

eradiate/thermoprops/tests/test_afgl1986.py
```diff
@@ -22,7 +22,7 @@
         "us_standard",
     ],
 )
-def test_make_profile(model_id: str) -> None:
+def test_make_profile(mode_mono, model_id: str) -> None:
```
leroyvn, line 25: Please don't add type hints to test functions.

leroyvn (changes requested):
Hi @nollety, thanks for this PR. It's briefer than I had was anticipating, it's nice! I have a few comments. Also, please be sure to rebase your branch on the data submodule so as to not remove the texture added for the `CentralPatchSurface`.

eradiate/thermoprops/afgl1986.py
nollety: I changed the docstrings too fast yesterday, I'll clarify them.

eradiate/radprops/rad_profile.py
nollety: I've gotten used to access data set members using the properties instead of the dict syntax because if I want to refactor things later, and also because the code is more compact, but I guess I could give the dict syntax another try 😃 In this particular case, the dict syntax is probably a better choice because the coordinate names match the names of the molecules that are in the thermo-physical properties data set.

eradiate/thermoprops/tests/test_afgl1986.py
nollety: I'll remove them!

PR #141, opened 2021-11-03
Switch to src layout

leroyvn:
# Description

This PR switches sources to the so-called "src layout". See [pytest docs](https://docs.pytest.org/en/6.2.x/goodpractices.html) for the motivation. In addition to moving files to the appropriate location, I made the following changes:

- Added `__init__.py` to every test subdirectory.
- Removed the disambiguating prefixes from filenames and test function names.
- Updated test report to the new source layout.

@nollety @schunkes please check whether this impacts your workflow. The source layout change will require that you fix your PyCharm configuration if you use it. You will also have to reinstall the Eradiate python package (the simplest way to do so is to rebuild your Conda env).

# Checklist


tests/kernel/test_gridvolume.py
```diff
@@ -2,17 +2,8 @@
...
 
-from eradiate.contexts import SpectralContext
 from eradiate.kernel.gridvolume import read_binary_grid3d, write_binary_grid3d
-from eradiate.radprops.rad_profile import US76ApproxRadProfile
-
-
-@pytest.fixture
-def test_radprops(mode_mono):
-    spectral_ctx = SpectralContext.new(wavelength=550.0)
-    return US76ApproxRadProfile().to_dataset(spectral_ctx=spectral_ctx)
```
nollety, line 15: I assume what you removed here was not used

tests/test_presolver.py
```diff
@@ -64,13 +64,14 @@ def test_resolve(presolver):
     presolver.clear()
 
     # We add a path
-    base = os.path.join(os.getenv("ERADIATE_DIR"), "eradiate")
+    base = os.getenv("ERADIATE_DIR")
     presolver.append(base)
+    print(presolver)
```
nollety, line 69: Just out of curiosity, why is this print statement needed here?

nollety (approved):
@leroyvn Great, I like this layout, it makes a lot of sense in my opinion! I hardly think this will have any (negative) impact on my workflow.

tests/test_presolver.py
leroyvn: It's not, it's a mistake. Thanks for catching it!

tests/kernel/test_gridvolume.py
leroyvn: Indeed, I checked for usage and got no hit.

leroyvn:
Seems like everything went well, let's merge!

PR #144, opened 2021-11-08
Updated the _inversebeta function 

schunkes:
A factor of two pi was missing in the `_inversebeta` function and the code utilizing it, was wrongly converting the results from degree to radians, when they were actually given in radians.

# Description

With this bugfix the code will finally produce leaf orientations that are correct.

# Checklist

### DIFF
--- a/src/eradiate/scenes/biosphere/_leaf_cloud.py
+++ b/src/eradiate/scenes/biosphere/_leaf_cloud.py
@@ -8,6 +8,8 @@ import attr
 import numpy as np
 import pint
 import pinttr
+import scipy as sp
+import scipy.special
 
 from ._core import CanopyElement, biosphere_factory
 from ..core import SceneElement
@@ -20,18 +22,28 @@ from ...units import unit_context_kernel as uck
 from ...units import unit_registry as ureg
 
 
-def _inversebeta(mu, nu, rng):
+def _sample_lad(mu, nu, rng):
     """
-    Approximates the inverse beta distribution as given in
-    :cite:`Ross1991MonteCarloMethods` (appendix 1).
+    Generate an angle sample from the Leaf angle distribution function according to
+    :cite:`GoelStrebel1984`, using the rejection method.
     """
+
     while True:
         rands = rng.random(2)
-        s1 = np.power(rands[0], 1.0 / mu)
-        s2 = np.power(rands[1], 1.0 / nu)
-        s = s1 + s2
-        if s <= 1:
-            return s1 / s
+        theta_candidate = rands[0] * np.pi / 2.0
+        gs_lad = (
+            2.0
+            / np.pi
+            * sp.special.gamma(mu + nu)
+            / (sp.special.gamma(mu) * sp.special.gamma(mu))
+            * pow((1 - (2 * theta_candidate) / np.pi), mu - 1)
+            * pow((2 * theta_candidate) / np.pi, nu - 1)
+        )
+
+        # scaling factor for the rejection method set to 2.0 to encompass the
+        # entire distribution
+        if rands[1] * 2.0 <= gs_lad:
+            return theta_candidate
 
 
 @ureg.wraps(ureg.m, (None, ureg.m, ureg.m, None))
@@ -169,8 +181,8 @@ def _leaf_cloud_orientations(n_leaves, mu, nu, rng):
     """Compute leaf orientations."""
     orientations = np.empty((n_leaves, 3))
     for i in range(np.shape(orientations)[0]):
-        theta = np.rad2deg(_inversebeta(mu, nu, rng))
-        phi = rng.random() * 360.0
+        theta = _sample_lad(mu, nu, rng)
+        phi = rng.random() * 2.0 * np.pi
 
         orientations[i, :] = [
             np.sin(theta) * np.cos(phi),
--- a/tests/scenes/biosphere/test_leaf_cloud.py
+++ b/tests/scenes/biosphere/test_leaf_cloud.py
@@ -1,6 +1,7 @@
 import os
 import tempfile
 
+import attr
 import numpy as np
 import pytest
 
@@ -8,7 +9,7 @@ from eradiate import unit_registry as ureg
 from eradiate.contexts import KernelDictContext
 from eradiate.scenes.biosphere._leaf_cloud import (
     LeafCloud,
-    _inversebeta,
+    _sample_lad,
     _leaf_cloud_orientations,
     _leaf_cloud_positions_cuboid,
     _leaf_cloud_positions_cuboid_avoid_overlap,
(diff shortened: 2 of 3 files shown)


src/eradiate/scenes/biosphere/_leaf_cloud.py
```diff
@@ -169,8 +169,8 @@ def _leaf_cloud_orientations(n_leaves, mu, nu, rng):
     """Compute leaf orientations."""
     orientations = np.empty((n_leaves, 3))
     for i in range(np.shape(orientations)[0]):
-        theta = np.rad2deg(_inversebeta(mu, nu, rng))
-        phi = rng.random() * 360.0
+        theta = _inversebeta(mu, nu, rng)
+        phi = rng.random() * 2 * np.pi
```
leroyvn, line 173:
```suggestion
        phi = rng.random() * 2.0 * np.pi
```

src/eradiate/scenes/biosphere/_leaf_cloud.py
```diff
@@ -31,7 +31,7 @@ def _inversebeta(mu, nu, rng):
         s2 = np.power(rands[1], 1.0 / nu)
         s = s1 + s2
         if s <= 1:
-            return s1 / s
+            return s2 / s * np.pi / 2.0
```
leroyvn, line 34: This is not a 2π factor: I assume the approximate was also wrong? If this departs from what's written in the reference you cite in the docstring, can you also update it?

tests/scenes/biosphere/test_leaf_cloud.py
```diff
@@ -42,7 +42,7 @@ def tempfile_leaves():
 
 def test_inversebeta(rng):
     """Unit tests for :func:`_inversebeta`."""
-    assert np.allclose(_inversebeta(1, 1, rng), 0.4178246)
+    assert np.allclose(_inversebeta(1, 1, rng), 0.914478941)
```
leroyvn, line 45: Basing this test on a RNG seems fragile to me (in the unlikely event of a RNG update, it will break while it should not). I'd rather have something more robust, properly checking that the `_inversebeta()` function really does what it is supposed to (*i.e.* approximate a more complicated thing).

tests/scenes/biosphere/test_leaf_cloud.py
```diff
@@ -80,7 +80,7 @@ def test_leaf_cloud_orientations(rng):
     """Unit tests for :func:`_leaf_cloud_orientations`."""
     orientations = _leaf_cloud_orientations(1, 1, 1, rng)
     assert orientations.shape == (1, 3)
-    assert np.allclose(orientations, [0.36553973, 0.85464537, 0.36873025])
+    assert np.allclose(orientations, [0.2323117, -0.75741865, 0.61020346])
```
leroyvn, line 83: Same here: this just tests for regression of the RNG + function block, not the function itself. We can discuss how to improve that if you wish.

leroyvn (changes requested):
Hi @schunkes, this a very welcome fix. However I'd like to seize the opportunity to improve the testing of this code (it's mostly about rescoping the tests). I guess you determined  the reference values using code which could be integrated in this test suite?

src/eradiate/scenes/biosphere/_leaf_cloud.py
schunkes: To be honest, I do not understand how this code follows from the cited source anyways. This is just copied from raytran. If we really want something we can cite and refer to, we should remove this and implement the distribution function by Goel and Strebel directly.

tests/scenes/biosphere/test_leaf_cloud.py
schunkes: I can implement the distribution from goel and strebel and do a statistical comparison of my two implementations, if that makes sense... But we don't really have an external reference for these statistical results which would exclude programming mistakes I could be making... :/

leroyvn: I think the best approach is to hard-code the samples you want to use for testing—just like what we do when we test sampling code in Mitsuba. This way you don't have any risk of regression due to the RNG being updated.

src/eradiate/scenes/biosphere/_leaf_cloud.py
leroyvn: Okay, so how about comparing the reference (Goel and Strebel) and this, both precision- and performance-wise? Since performance matters when it comes to generating those leaf clouds, I can see how an approximate would make sense.

tests/scenes/biosphere/test_leaf_cloud.py
leroyvn: As for testing the distribution itself, it's a broader-scope upgrade: Doing this without designing a proper statistical test framework would be wasteful IMO. So I'd advise against for now.

src/eradiate/scenes/biosphere/_leaf_cloud.py
schunkes: the performance of sampling the Goel and Strebel function using the rejection method is faster, I already checked that (however it takes single digit number of seconds to generate 100k samples in both cases, so this seems very uncritical to me)

for the precision: well there we need a statistical test. but we could instead just drop the `_inversebeta` function and just sample the original distribution and be done with it.

leroyvn: Going for the Goel & Strebel function is better IMO, especially if it's what RAMI instructs—is it?

tests/scenes/biosphere/test_leaf_cloud.py
```diff
@@ -24,6 +24,29 @@ def rng():
...
+                    returnlist.append(self.b)
+                else:
+                    returnlist.append(self.a)
+
+            if len(returnlist) == 1:
+                return returnlist[0]
+            else:
+                return returnlist
+
+    yield RngMock
```
leroyvn, line 47: How about generalising this with a list?
```suggestion
@pytest.fixture(scope="function")
def rng_mock():
    @attr.s
    class RngMock:
        values = attr.ib(converter=np.atleast_1d)

        def random(self, n=1):
            selector = np.array([i % len(self.values) for i in range(n)])
            return self.values[selector].squeeze()

    yield RngMock
```
This way you can initialise it with any sequence you like.

leroyvn (changes requested):
Hi @schunkes, this seems good to me apart from a couple of minor details. Once they're addressed, you can clean up the commit history and we'll merge!

docs/conf.py
```diff
@@ -128,7 +128,7 @@ def find_version(*file_paths):
     "reference_url": {
         "eradiate": None,  # The module you locally document uses None
     },
-    "plot_gallery": False,  # Disabled until we move away from RTD or package the kernel
+    "plot_gallery": True,  # Disabled until we move away from RTD or package the kernel
```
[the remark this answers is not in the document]
schunkes: argh, I thought I explicitly did not commit this!!!

PR #145, opened 2021-11-08
Added test of consistency between ckd and mono modes

nollety:
# Description

Add a test that assert that monochromatic results are consistent with CKD results.

We run a one-dimensional experiment with the following settings:

* lambertian surface with reflectance of [0.0, 0.5, 1.0] (parametrised)
* directional illumination with `zenith=50.0 deg` and `azimuth=0.0 deg`
* AFGL (1986) `us_standard` atmosphere (`has_scattering=True`, `has_absorbing=True`)
* `spp=1e4`
* multi-distant measures with 11 viewing zenith angles linearly spaced in [-75, 75]°.

The experiment is run in `mono_double` and `ckd_double` mode with the following specific settings
* `mono_double`:
  * `wavelengths = np.linspace(545.0, 555.0, 1001) * ureg.nm`
* `ckd_double`:
  * `bin_set = "10nm"`
  * `bins = "550"`

Before comparing the results, the monochromatic results are averaged.
The comparaison targets the BRF in the principal plane, for angles inferior to 75° in absolute values.

⚠️ Note that the test will be skipped if the monochromatic absorption cross section data sets for H2O are not present.

The test has also been marked as slow at the moment because it takes around 7 minutes to complete. However, I plan to try to reduce the number of viewing zenith angles and samples per pixel to see if a reasonable executing time is within reach.

Of course, the current spectral coverage (545 to 555 nm) of that test is pretty poor and we shall remedy that.
I'll investigate other spectral regions, but that means that the total executing time of such tests will increase and we will probably end up marking all these tests as slow, which is not necessarily bad I assume.
What do you think? @leroyvn 

# Checklist

- [x] print figures
- [x] more spectral regions
- [x] change surface reflectance
- [x] reduce `n_vza` and `spp`
- [x] number of wavelenght points in mono mode
- [x] add doc-strings

### DIFF
--- a/tests/_system/test_onedim_ckd_vs_mono.py
+++ b/tests/_system/test_onedim_ckd_vs_mono.py
@@ -1,0 +1,379 @@
+"""Test cases asserting the results consistency between the mono and ckd modes."""
+import os
+
+import matplotlib.pyplot as plt
+import numpy as np
+import pytest
+
+import eradiate
+from eradiate import unit_registry as ureg, unit_context_kernel as uck
+
+eradiate_dir = os.environ["ERADIATE_DIR"]
+output_dir = os.path.join(eradiate_dir, "test_report", "generated")
+
+
+def ensure_output_dir(path):
+    if not os.path.isdir(path):
+        os.makedirs(path)
+
+
+def init_mono_experiment(wavelengths, spp, reflectance, zeniths):
+    """
+    Initialise the monochromatic experiment.
+    """
+    eradiate.set_mode("mono_double")
+
+    return eradiate.experiments.OneDimExperiment(
+        measures=[
+            eradiate.scenes.measure.MultiDistantMeasure.from_viewing_angles(
+                spectral_cfg=eradiate.scenes.measure.MeasureSpectralConfig.new(
+                    wavelengths=wavelengths
+                ),
+                zeniths=zeniths,
+                azimuths=0.0 * ureg.deg,
+                spp=spp,
+            )
+        ],
+        illumination=eradiate.scenes.illumination.DirectionalIllumination(
+            zenith=50.0 * ureg.deg, azimuth=0.0 * ureg.deg
+        ),
+        atmosphere=eradiate.scenes.atmosphere.MolecularAtmosphere.afgl_1986(
+            id="molecules",
+            has_absorption=True,
+            has_scattering=True,
+            absorption_data_sets={},
+        ),
+        surface=eradiate.scenes.surface.LambertianSurface(
+            reflectance=reflectance,
+        ),
+    )
+
+
+def init_ckd_experiment(bin_set, bins, spp, reflectance, zeniths):
+    """
+    Initialise CKD experiment.
+    """
+    eradiate.set_mode("ckd_double")
+
+    return eradiate.experiments.OneDimExperiment(
+        measures=[
+            eradiate.scenes.measure.MultiDistantMeasure.from_viewing_angles(
+                spectral_cfg=eradiate.scenes.measure.MeasureSpectralConfig.new(
+                    bin_set=bin_set,
+                    bins=bins,
+                ),
+                zeniths=zeniths,
+                azimuths=0.0 * ureg.deg,
+                spp=spp,
+            )
+        ],
+        illumination=eradiate.scenes.illumination.DirectionalIllumination(
+            zenith=50.0 * ureg.deg, azimuth=0.0 * ureg.deg
+        ),
+        atmosphere=eradiate.scenes.atmosphere.MolecularAtmosphere.afgl_1986(
+            id="molecules",
+            has_absorption=True,
+            has_scattering=True,
+        ),
+        surface=eradiate.scenes.surface.LambertianSurface(
+            reflectance=reflectance,
+        ),
+    )
+
+
+def make_figure(
+    ckd_results, mono_results_averaged, fname_plot, wavelength_bin, reflectance
+):
+    fig = plt.figure(figsize=(8, 5))
+    brf_mono_averaged = mono_results_averaged.brf.squeeze()
+    brf_ckd = ckd_results.brf.squeeze()
+    brf_mono_averaged.plot(x="vza", ls="dotted", marker=".")
+    brf_ckd.plot(x="vza", ls="dotted", marker=".")
+
+    # Compute relative absolute difference and annotate plot with min, max and
+    # mean values
+    # NB: RAD(x, y) = abs(x - y) / x
+    rad = np.divide(
+        np.abs(brf_mono_averaged.values - brf_ckd.values),
+        brf_mono_averaged.values,
+        where=brf_mono_averaged.values != 0.0,
+        out=np.full_like(brf_mono_averaged.values, fill_value=np.nan),
+    )
+    rmad = np.mean(rad[~np.isnan(rad)])
+    minrad = np.min(rad[~np.isnan(rad)])
+    maxrad = np.max(rad[~np.isnan(rad)])
+    text = (
+        f"mean(RAD) >= {round(100 * rmad, 2)} %\n"
+        f"min(RAD) >= {round(100 * minrad, 2)} %\n"
+        f"max(RAD) >= {round(100 * maxrad, 2)} %"
+    )
+    plt.text(
+        0.5,
+        0.9,
+        text,
+        ha="center",
+        va="center",
+        transform=plt.gca().transAxes,
+    )
+
+    plt.legend(["mono_double (averaged)", "ckd_double"])
+    plt.title(f"Wavelength bin: {wavelength_bin}, reflectance = {reflectance}")
+    plt.tight_layout()
+    fig.savefig(fname_plot, dpi=200)
+    plt.close()
+
+
+@pytest.mark.parametrize("reflectance", [0.0, 0.5, 1.0])
+@pytest.mark.skipif_data_not_found("absorption_spectrum", "H2O-spectra-18000_18100")
+@pytest.mark.skipif_data_not_found("absorption_spectrum", "H2O-spectra-18100_18200")
+@pytest.mark.skipif_data_not_found("absorption_spectrum", "H2O-spectra-18200_18300")
+@pytest.mark.skipif_data_not_found("absorption_spectrum", "H2O-spectra-18300_18400")
+@pytest.mark.slow
+def test_550(reflectance):
+    r"""
+    Results consistency between `mono_double` and `ckd_double` modes
+    ================================================================
+
+    This test checks the results consistency between mono_double and ckd_double
+    modes in the [545, 555] nm wavelength bin.
+
+    Rationale
+    ---------
+
+    * Sensor: Distant measure covering a plane (11 angular points,
+      10000 sample per pixel) and targeting (0, 0, 0).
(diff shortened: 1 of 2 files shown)


nollety:
Below is the BRF comparaison illustrated (note that the monochromatic results are averaged before compared):

test_onedim_ckd_vs_mono

tests/_system/test_onedim_ckd_vs_mono.py
```diff
@@ -0,0 +1,178 @@
...
+    ckd_results = ckd_exp.results["measure"]
+
+    # Make figure
+    fig = plt.figure(figsize=(8, 5))
+    filename = f"test_onedim_ckd_vs_mono.png"
+    ensure_output_dir(os.path.join(output_dir, "plots"))
+    fname_plot = os.path.join(output_dir, "plots", filename)
+
+    mono_results_averaged.brf.squeeze().where(mono_results_averaged.brf.vza < 75).where(
+        mono_results_averaged.brf.vza > -75
```
leroyvn, line 154: I think the following is more elegant and should work (please double-check):
```suggestion
    mono_results_averaged.brf.squeeze().where(
        np.logical_and(
            mono_results_averaged.brf.vza < 75, mono_results_averaged.brf.vza > -75
        )
    )
```

tests/_system/test_onedim_ckd_vs_mono.py
```diff
@@ -0,0 +1,178 @@
...
+    plt.title("")
+    plt.tight_layout()
+    fig.savefig(fname_plot, dpi=200)
+    plt.close()
+
+    # Assert averaged mono results are consistent with ckd results
+    mono_brf = (
+        mono_results_averaged.brf.squeeze()
+        .where(mono_results_averaged.brf.vza < 75)
+        .where(mono_results_averaged.brf.vza > -75)
```
leroyvn, line 169: ditto

tests/_system/test_onedim_ckd_vs_mono.py
```diff
@@ -0,0 +1,178 @@
...
+        .values
+    )
+    ckd_brf = (
+        ckd_results.brf.squeeze()
+        .where(ckd_results.brf.vza < 75)
+        .where(ckd_results.brf.vza > -75)
```
leroyvn, line 175: ditto

leroyvn (commented):
> We run a one-dimensional experiment with the following settings:
>
> * lambertian surface with reflectance of 0.0 (I will change that to a non-zero values, possibly parametrised the test to check with several values)
> * directional illumination with zenith=50.0 deg and azimuth=0.0 deg
> * spp=1e5
> * 40 viewing zenith angles (probably too many)

* You can likely reduce the VZA parametrisation to {-75, -60, -45, -30, -15, 0, 15, 30, 45, 60, 75} (11 values).
* Surface reflectance could be {0, 0.5, 1}.
* What are your atmosphere params?

> The experiment is run in `mono_double` and `ckd` mode with the following specific settings

For consistency, please use the `ckd_double` mode.

> * `mono_double`: `wavelengths=np.linspace(545.0, 555.0, 2) * ureg.nm`
> * `ckd`: `bin_set="10nm", bins="550"`

I'm surprised that this yields correct results at all: I guess this means that absorption plays a diminutive role, since two wavelengths in mono mode are insufficient so as to capture absorption features correctly. IMO you don't even have to bother running two monochromatic simulations and averaging them: just simulate once at 550 nm.

> Note that the test will be skipped if the monochromatic absorption cross section data sets for H2O are not present.

Could we ship a minimal dataset for testing purposes or is this too much data?

> The test has also been marked as slow at the moment because it takes 40 seconds to complete. However, I plan to try to reduce the number of viewing zenith angles and samples per pixel to see if a reasonable executing time is within reach.

Good.

> Of course, the current spectral coverage (545 to 555 nm) of that test is pretty poor and we shall remedy that. I'll investigate other spectral regions, but that means that the total executing time of such tests will increase and we will probably end up marking all these tests as slow, which is not necessarily bad I assume. What do you think?

Slow tests are fine, but they still have to mean something. In its current form, is this test really checking that the CKD mode yields correct results? Sort of, but only if you consider that it's about testing that the basic machinery works as expected.

A proper assessment of the accuracy of our CKD method is a more ambitious test because it requires running a complete spectral simulation on a band—way out of the scope of the test suite.

nollety:
>     * You can likely reduce the VZA parametrisation to {-75, -60, -45, -30, -15, 0, 15, 30, 45, 60, 75} (11 values).
> 

Done!

>     * Surface reflectance could be {0, 0.5, 1}.

Done!

>     * What are your atmosphere params?

I've updated the PR description with the atmosphere parameters.
 
> > The experiment is run in `mono_double` and `ckd` mode with the following specific settings
> 
> For consistency, please use the `ckd_double` mode.

Right, this is done now.

> > * `mono_double`: `wavelengths=np.linspace(545.0, 555.0, 2) * ureg.nm`
> > * `ckd`: `bin_set="10nm", bins="550"`
> 
> I'm surprised that this yields correct results at all: I guess this means that absorption plays a diminutive role, since two wavelengths in mono mode are insufficient so as to capture absorption features correctly. IMO you don't even have to bother running two monochromatic simulations and averaging them: just simulate once at 550 nm.

This is a copy-paste mistake, I've run the test with different number of wavelengths in monochromatic mode and ended up using only 2 when experimenting with the scene parameters.

> Could we ship a minimal dataset for testing purposes or is this too much data?

I think there is room for the required datasets on our FTP server, but not on `eradiate-data`.
 
> A proper assessment of the accuracy of our CKD method is a more ambitious test because it requires running a complete spectral simulation on a band—way out of the scope of the test suite.

I changed the number of wavelengths in monochromatic mode to 1001 which correspond to a 0.01 nm resolution, i.e. 10 times less than the native datasets resolution. I think that if we work in [549.5, 550.5] nm with the bin "550" from the "1nm" bin set (once the datasets are ready, ETA: beginning next week), we could achieve that.

tests/_system/test_onedim_ckd_vs_mono.py
nollety: With the vza parametrisation that you suggested, this is not necessary anymore :)

nollety:
I've duplicated the test for the [1045, 1055] nm wavelength bin, just to confirm that things work well there too (and it does). I don't believe it's worth it to add any more wavelength bin for the time being.

nollety:
We can always add more tests of the kind later.

nollety:
Below are the figures that the tests output (RAD stands for "relative absolute difference"):

test_onedim_ckd_vs_mono_550_0 0

test_onedim_ckd_vs_mono_550_0 5

test_onedim_ckd_vs_mono_550_1 0

test_onedim_ckd_vs_mono_1050_0 0

test_onedim_ckd_vs_mono_1050_0 5

test_onedim_ckd_vs_mono_1050_1 0

tests/_system/test_onedim_ckd_vs_mono.py
```diff
@@ -0,0 +1,310 @@
...
+
+def make_figure(ckd_results, mono_results_averaged, fname_plot, wavelength_bin):
+    fig = plt.figure(figsize=(8, 5))
+    brf_mono_averaged = mono_results_averaged.brf.squeeze()
+    brf_ckd = ckd_results.brf.squeeze()
+    brf_mono_averaged.plot(x="vza", ls="dotted", marker=".")
+    brf_ckd.plot(x="vza", ls="dotted", marker=".")
+
+    # Compute relative absolute difference and annotate plot with min, max and
+    # mean values
```
leroyvn, line 92: I find the "relative absolute difference" nomenclature to be quite confusing in our context ("absolute bias" is actually a signed quantity in usual EO terminology). However, since this is correct vocabulary, let's clarify this:
```suggestion
    # Compute relative absolute difference and annotate plot with min, max and
    # mean values
    # NB: RAD(x, y) = abs(x - y) / x
```
Please also add similar detail in the text docstring.

leroyvn (changes requested):
Hi @nollety, seems good to me apart from a few details. When these are taken care of, we can merge!

tests/_system/test_onedim_ckd_vs_mono.py
```diff
@@ -0,0 +1,310 @@
...
+        ckd_results=ckd_results,
+        mono_results_averaged=mono_results_averaged,
+        fname_plot=fname_plot,
+        wavelength_bin="[1045, 1055] nm",
+    )
+
+    # Assert averaged mono results are consistent with ckd results
+    mono_brf = mono_results_averaged.brf.squeeze().values
+    ckd_brf = ckd_results.brf.squeeze().values
+    assert np.allclose(mono_brf, ckd_brf, rtol=5e-2, equal_nan=True)
```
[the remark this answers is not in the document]
nollety: This is probably not required anymore. I was expecting nan values as a result of applying `where(vza < 75)` to the BRF data arrays ; the values corresponding to vza >= 75° would be set to nan, but these values were out of the scope of the comparison, hence the `equal_nan=True`. I'll remove this setting.

tests/_system/test_onedim_ckd_vs_mono.py
nollety: I understand your confusion. I derived the definition of "relative absolute difference" from the "mean absolute difference" definition from [wikipedia](https://en.wikipedia.org/wiki/Mean_absolute_difference#Relative_mean_absolute_difference) and it sounds a little weird as is 😃.
Where did you find the definition of absolute bias?
It looks like it would be nice to settle this vocabulary throughout eradiate since we use it a lot, don't you think?
As defined, the "relative absolute difference" can be negative, when `x` is negative (unlikely given that the quantities that we usually compare are defined positive).

tests/_system/test_onedim_ckd_vs_mono.py
```diff
@@ -0,0 +1,375 @@
...
+
+    where
+
+    * :math:`q` is some spectral quantity (e.g. the BRF)
+    * :math:`\lambda` denotes the wavelength
+    * :math:`[\lambda_{\mathrm{min}},\, \lambda_{\mathrm{max}}]` is the
+      wavelength bin
+
+    Results
+    -------
```
leroyvn, line 189: I suggest also annotating the plots
```suggestion
    -------
    
    Note: :math:`\mathrm{RAD}(x, y) = \left| \dfrac{x - y}{x} \right|`

```

tests/_system/test_onedim_ckd_vs_mono.py
```diff
@@ -0,0 +1,375 @@
...
+
```
leroyvn, line 320: Ditto:
```suggestion
    
    Note: :math:`\mathrm{RAD}(x, y) = \left| \dfrac{x - y}{x} \right|`

```

leroyvn (commented):
Hi @nollety, let's also annotate the plots (the LaTeX might require a very quick check). Then let's merge!

nollety:
For the record, these tests are very slow:

* tests run time **without** these tests: 2 minutes 10 seconds
* tests run time **including** these tests: ~23 minutes

nollety:
@leroyvn I've added the annotation you suggested and checked the LaTeX rendering, this is ready to be merged! 😃

leroyvn:
Thanks! Let's go :)

PR #149, opened 2021-11-18
Solar irradiance spectrum: Add datetime-based scaling

leroyvn:
# Description

This PR add support for datetime-based Solar irradiance spectrum scaling. Changes are as follows:

- `SolarIrradianceSpectrum`'s `scale` parameter can now be set to a datetime value, possibly using a string converted using the dateutil package. Scaling is computed based on the Earth-Sun distance, computed using astropy and assuming that the irradiance spectrum data is given for a reference Earth-Sun distance of 1 AU. Corresponding tests have been also added.
- Requirements were updated and docs inventory files for astropy and dateutil are now referenced.

Example of scaling usage:
```python
illumination = DirectionalIllumination(
    zenith=30.0,
    azimuth=0.0,
    spectrum=SolarIrradianceSpectrum(scale="2021-11-18")
)
```

# Checklist


nollety:
That's great!

Just a comment:
One of the solar irradiance spectrum data sets in our collection -- ``solid_2017`` -- already has a time coordinate that can be used to fetch the solar irradiance spectrum on a particular day:

```python
import eradiate

ds = eradiate.data.open(
    category="solar_irradiance_spectrum",
    id="solid_2017"
)
ds.sel(t="2014-11-18")
```

This is not the same as scaling the whole solar irradiance spectrum by one factor, because the Earth-Sun distance not the only factor that is affecting the solar irradiance spectrum observed at the top of Earth's atmosphere.
So I don't think it's a good idea to try to forward the value of ``scale`` in ``Dataset.sel(t=scale)`` but the variations of the solar irradiance spectrum with time captured by a data set such as ``solid_2017`` would most probably be worth including.
Something to consider for the future perhaps.

leroyvn:
Hi @nollety, I disregarded the SOLID spectrum (the time-indexed dataset is not supported IIRC). Whether distance-based scaling is relevant for a time-indexed dataset depends on whether the spectrum is normalised to a constant reference Earth-Sun distance—do you know if this is true for SOLID?

Anyway, for time-dependent spectra, we might have to update the `SolarIrradianceSpectrum` interface as follows:

- `scale`: only accepts a floating point value;
- `datetime`: new field, used to apply distance scaling to time-independent data and to select the time data point in time-dependent data.

For time-dependent data, the aforementioned condition for distance scaling still applies. This is just a first idea, I'm not convinced yet that this kind of interface will produce consistent behaviour between time-independent and time-dependent data.

leroyvn:
@nollety I changed the interface to leave the option of using a time coordinate. I'll squash and merge if it looks better to you.

```python
illumination = DirectionalIllumination(
    zenith=30.0,
    azimuth=0.0,
    spectrum=SolarIrradianceSpectrum(datetime="2021-11-18")
)
```

nollety:
This is perfect! 😃

PR #99, opened 2021-06-24
Test report refactoring

leroyvn:
# Description

This PR greatly improves the look and feel of the test report. Changes are as follows:

* The generating code was refactored to eliminate duplicates.
* A dropdown-based report output specific to HTML was added.
* The list-based report code was cleaned up and its output fixed.
* The theme was changed to Furo (features a navbar to the right-hand side).

This introduces the following dependencies:

* [sphinx-panels](https://sphinx-panels.readthedocs.io) (already a dependency of the Eradiate docs)
* [Furo](https://pradyunsg.me/furo/) (not included with Eradiate)

The new Furo req won't be added to our dependency list since this piece of code should eventually become an external tool.

**Selected screenshots:**

Screenshot 2021-06-24 at 17 12 48

Screenshot 2021-06-24 at 17 13 12

@schunkes can you please check if everything is fine?

# Checklist

- [x] The code follows the relevant coding guidelines
- [x] The code generates no new warnings
- [x] The code is appropriately documented
- [x] The code is tested to prove its function
- [x] The feature branch is rebased on the current state of the `main` branch
- [x] I give permission that the Eradiate project may redistribute my contributions under the terms of its license

PR #154, opened 2021-11-29
Extracting test report infrastructure

schunkes:
# Description

I removed the test report generation from the main Eradiate repo.
Additionally I introduced a new environment variable in the eradiate config:

ERADIATE_SYSTEMTEST_OUTPUT_DIR is set to `$ERADIATE_DIR/test_report/generated` by default, to provide a location for the system tests' plots. The test report generator replaces that directory with one in its file tree, to make the plots available for the test report.

# Checklist


tests/_system/test_onedim_ckd_vs_mono.py
```diff
@@ -6,10 +6,10 @@
 import pytest
 
 import eradiate
-from eradiate import unit_registry as ureg, unit_context_kernel as uck
+from eradiate import unit_context_kernel as uck
```
schunkes, line 9: This refactoring was done automatically by black, I suppose.

leroyvn: By isort.

schunkes: Anyway, nothing wrong with this.

src/eradiate/_config.py
```diff
@@ -87,6 +87,12 @@ def _data_path_validator(self, var, paths):
         help="Path to the Eradiate download directory.",
     )
 
+    systemtest_output_dir = environ.var(
+        default="$ERADIATE_DIR/test_report/generated",
+        converter=lambda x: pathlib.Path(os.path.expandvars(x)).absolute(),
+        help="Path to the Eradiate system test plot outputs.",
+    )
+
```
leroyvn, line 95: Test artifact output has little to do with Eradiate per se. I'd prefer that this configuration would be scoped to tests only. I'm sure pytest has infrastructure for passing configuration items upon launching the test session --- something like what you did for JSON artifacts. Can you investigate that?

leroyvn (commented):
Hi @schunkes, before I perform a thorough code review, there is a fundamental issue I'd like to address. For now I'll convert this PR to a draft, it's not ready for merge.

src/eradiate/_config.py
leroyvn: [This](https://stackoverflow.com/questions/40880259/how-to-pass-arguments-in-pytest-by-command-line) seems like a good way to start.

schunkes:
So how about you take another look at this code. Now the path for the systemtest plots is given by a pytest parametrization and the eradiate_test_report sets this to a value in its file tree, so that it can create the test report.

tests/conftest.py
```diff
@@ -83,6 +83,24 @@ def fixture(request):
...
 
+# ------------------------------------------------------------------------------
+#                Adding a customizable output dir for plots
+# ------------------------------------------------------------------------------
+
+
+def pytest_addoption(parser):
+    parser.addoption(
+        "--systemtest_outdir", action="store", default="./test_report/generated/plots"
+    )
```
leroyvn, line 94: Very nice! Let's broaden the scope of this and call that fixture `artefacts_dir`: this way, it's clear that all artefacts are to be exported to this directory, not only system test-related plots.

I suggest to set the default to `build/tests/artefacts`. Side question: what happens with this default if you call pytest in another directory that `$ERADIATE_DIR`?

tests/conftest.py
```diff
@@ -83,6 +83,24 @@ def fixture(request):
...
+
+
+def pytest_generate_tests(metafunc):
+    # This is called for every test. Only get/set command line arguments
+    # if the argument is specified in the list of test "fixturenames".
+    option_value = metafunc.config.option.systemtest_outdir
+    if "systemtest_outdir" in metafunc.fixturenames and option_value is not None:
```
leroyvn, line 101: Can it be `None` given that you set a default?

tests/conftest.py
```diff
@@ -83,6 +83,24 @@ def fixture(request):
...
+        metafunc.parametrize("systemtest_outdir", [option_value])
```
leroyvn, line 102: Maybe pre-processing this path and ensuring that the directory exists (or creating it if it doesn't) would be a good idea?

leroyvn (changes requested):
Hi @schunkes, I like the solution you came up with, I made suggestions for generalisation and improvements.

I'd appreciate if you could clarify the scope of this PR (notably through its title) because it's, in the end, more about addressing eradiate/eradiate#99 than about adding a regression testing framework (the latter deserves to be transferred to a new PR IMO).

tests/conftest.py
schunkes: I updated this part. The default is now $ERADIATE_DIR as long as that is set and otherwise defaults to the current working directory. But to be honest, if you try to run the tests without the eradiate conda environment, which sets that, the tests won't run anyways.

tests/conftest.py
```diff
@@ -83,6 +85,29 @@ def fixture(request):
     generate_fixture_group(name, variants)
 del generate_fixture_group
 
+# ------------------------------------------------------------------------------
+#                Adding a customizable output dir for plots
```
leroyvn, line 89:
```suggestion
#              Adding a customizable output dir for artefacts
```

tests/conftest.py
```diff
@@ -133,3 +158,11 @@ def pytest_configure(config):
...
         "cannot be found in paths resolved by Eradiate.",
     )
+
+    config.addinivalue_line(
+        "markers",
+        "regression: regression tests that are used in CI "
+        "and test reports. Tests that marked as regression"
+        "tests are only run if the regression marker is"
+        "set explicitly.",
+    )
```
leroyvn, line 168: This should be moved to a PR dedicated to the addition of the regression testing framework.

tests/conftest.py
leroyvn: Sounds good to me.

leroyvn (changes requested):
Hi @schunkes, this looks good. There are a couple of minor details to take care of, then you can clean up the history and we'll merge this!

PR #158, opened 2021-12-01
Make `test_onedim_phase` more reliable

leroyvn:
# Description

This PR makes the `test_onedim_phase` test case more reliable. Changes are as follows:

- Switched to double precision
- Use default kernel length units (also passes when using km, should we also parametrise the test with this?)
- Require precision is now 0.01%
- Minor plotting tweaks

# Checklist


nollety (approved):
Great, thanks for improving this test!

PR #160, opened 2021-12-02
Sensors in the atmosphere

schunkes:
# Description

This PR adds functionality which lets Experiment classes that support atmospheres add an external medium to the sensors, which are placed inside the atmosphere's volume.

# Checklist

### DIFF
--- a/src/eradiate/contexts.py
+++ b/src/eradiate/contexts.py
@@ -305,6 +305,14 @@ class KernelDictContext(Context):
         default="True",
     )
 
+    atmosphere_medium_id: t.Optional[str] = documented(
+        attr.ib(default=None, converter=attr.converters.optional(str)),
+        doc="If relevant, a kernel medium ID to reference by other kernel components.",
+        type="str or None",
+        init_type="str, optional",
+        default="None",
+    )
+
     override_scene_width: t.Optional[pint.Quantity] = documented(
         pinttr.ib(
             default=None,
--- a/src/eradiate/experiments/_onedim.py
+++ b/src/eradiate/experiments/_onedim.py
@@ -26,7 +26,7 @@ def measure_inside_atmosphere(atmosphere, measure, ctx):
     bbox = atmosphere.eval_bbox(ctx)
 
     if isinstance(measure, MultiRadiancemeterMeasure):
-        inside = [bbox.contains(origin) for origin in measure.origins]
+        inside = bbox.contains(measure.origins)
         if all(inside):
             return True
         elif not any(inside):
--- a/src/eradiate/experiments/_onedim.py
+++ b/src/eradiate/experiments/_onedim.py
@@ -10,10 +10,36 @@ from ..exceptions import OverriddenValueWarning
 from ..scenes.atmosphere import Atmosphere, HomogeneousAtmosphere, atmosphere_factory
 from ..scenes.core import KernelDict
 from ..scenes.integrators import Integrator, VolPathIntegrator, integrator_factory
-from ..scenes.measure import Measure, TargetPoint
+from ..scenes.measure import Measure, MultiRadiancemeterMeasure, TargetPoint
 from ..scenes.measure._core import MeasureFlags
 from ..scenes.surface import LambertianSurface, Surface, surface_factory
 from ..units import unit_context_config as ucc
+
+
+def measure_inside_atmosphere(atmosphere, measure, ctx):
+    """
+    Evaluate whether a sensor is placed within an atmosphere.
+
+    Raises a ValueError if called with a :class:`MultiRadiancemeterMeasure` with
+    origins both inside and outside of the atmosphere.
+    """
+    bbox = atmosphere.eval_bbox(ctx)
+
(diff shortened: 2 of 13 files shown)


src/eradiate/experiments/_rami4atm.py
```diff
@@ -43,6 +44,12 @@ class Rami4ATMExperiment(EarthObservationExperiment):
     * if neither atmosphere nor canopy are defined, the target is set to
       [0, 0, 0].
 
+    If a sensor is placed inside the atmosphere's volume, the experiment will attempt to specify
+    the surrounding medium of that sensor. If a :class:`MultiRadiancemeterMeasure` with subsensors
+    which are partially inside and outside of the atmosphere is created, the experiment will raise
+    a ValueError. In this case please use two separate measures and group the subsensors which are
+    inside and outside of the atmosphere.
+
```
leroyvn, line 52: Same comment as for the 1D module.

leroyvn (changes requested):
Hi @schunkes, I made a bunch of comments and suggestions for a cleaner implementation. I see overall two important issues which I tried to address:

1. The code is lengthy and hard to understand, but a bounding box abstraction will make it more compact, testable and legible.
2. This implementation relies on an instance method which, by nature, is not meant for reuse unless you subclass (not doable in our case). My suggestions address this by making the duplicate component a pure function.

src/eradiate/experiments/_onedim.py
```diff
@@ -93,14 +99,79 @@ def kernel_dict(self, ctx: KernelDictContext) -> KernelDict:
...
+                self.surface,
+                self.illumination,
+                *self.measures,
+                self.integrator,
+                ctx=ctx
+            )
 
         return result
 
+    def _is_measure_inside_atm(self, measure, ctx):
```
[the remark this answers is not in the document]
schunkes: I did add a method to the Atmosphere class. I agree that this kind of method makes sense and could be useful in the future as well.

src/eradiate/contexts.py
```diff
@@ -305,6 +305,15 @@ class KernelDictContext(Context):
         default="True",
     )
 
+    atmosphere_medium_id: t.Optional[str] = documented(
+        attr.ib(default=None, converter=attr.converters.optional(str)),
+        doc="If relevant, value which must be used as the surrounding medium for sensors, which"
+        "are positioned inside the atmosphere volume.",
```
leroyvn, line 311: Let's also broaden the scope of this doc:
```suggestion
        doc="If relevant, a kernel medium ID to reference by other kernel components.",
```

src/eradiate/experiments/_onedim.py
```diff
@@ -10,12 +10,38 @@
...
 from ..scenes.measure._core import MeasureFlags
 from ..scenes.surface import LambertianSurface, Surface, surface_factory
 from ..units import unit_context_config as ucc
 
 
+def measure_inside_atmosphere(atmosphere, measure, ctx):
+    """
+    Evaluate whether a sensor is placed within an atmosphere.
+
+    Raises a ValueError if calld with a :class:`MultiRadiancemeterMeasure` with
```
leroyvn, line 23:
```suggestion
    Raises a ValueError if called with a :class:`.MultiRadiancemeterMeasure` with
```

src/eradiate/experiments/_onedim.py
```diff
@@ -10,12 +10,38 @@
...
+    origins that are inside as well as outside of the atmosphere.
```
leroyvn, line 24:
```suggestion
    origins both inside and outside of the atmosphere.
```

src/eradiate/experiments/_onedim.py
```diff
@@ -10,12 +10,38 @@
...
+    """
+    bbox = atmosphere.eval_bbox(ctx)
+
+    if isinstance(measure, MultiRadiancemeterMeasure):
+        if all(bbox.contains(origin) for origin in measure.origins):
+            return True
+        elif not any(bbox.contains(origin) for origin in measure.origins):
+            return False
```
leroyvn, line 32: Still with the idea of not evaluating twice:
```suggestion
        bbox_contains = [bbox.contains(origin) for origin in measure.origins]
        if all(bbox_contains):
            return True
        elif not any(bbox_contains):
            return False
```

src/eradiate/experiments/_onedim.py
```diff
@@ -10,12 +10,38 @@
...
+        else:
+            raise ValueError(
+                "Inconsistent placement of Multiradiancemeter origins. "
+                "Either all sensors must be placed inside the participating "
+                "medium or all must be placed outside."
```
leroyvn, line 37: How about a briefer
```suggestion
                "Inconsistent placement of MultiRadiancemeterMeasure origins. "
                "All origins must lie either inside or outside of the atmosphere."
```

leroyvn (commented):
Hi @schunkes, this looks good to me—I just had a couple of comments, small details. When this is taken care of, we can squash + merge!

leroyvn:
I made a few more tiny updates, let's merge. Thanks @schunkes!

PR #163, opened 2022-01-06
Spherical grid volume plugin

schunkes:
# Description

This PR introduces a mitsuba plugin that maps the `grid` plugin to spherical coordinates. This allows for the creation of a spherical shell volume, which in turn enables the simulation of a spatially varying atmosphere around a spherical earth!

This plugin does not have tests yet. Also there is an issue, when adding a `rotation` to the volume's transformation matrix.
It is unknown whether this is actually an issue, since only rotationally symmetric atmospheres are currently planned.

# Checklist

### DIFF
--- a/src/plugins/src/volumes/CMakeLists.txt
+++ b/src/plugins/src/volumes/CMakeLists.txt
@@ -1,3 +1,3 @@
 set(MTS_PLUGIN_PREFIX "volumes")
 
-add_plugin(grid_spherical grid_spherical.cpp)
+add_plugin(sphericalgridvolume sphericalgrid.cpp)
--- a/src/plugins/src/volumes/CMakeLists.txt
+++ b/src/plugins/src/volumes/CMakeLists.txt
@@ -1,0 +1,3 @@
+set(MTS_PLUGIN_PREFIX "volumes")
+
+add_plugin(grid_spherical grid_spherical.cpp)
--- a/src/plugins/src/volumes/grid_spherical.cpp
+++ b/src/plugins/src/volumes/grid_spherical.cpp
@@ -28,10 +28,25 @@ Grid-based volume data source in spherical coordinates(:monosp:`sphericalgridvol
    - Nested gridvolume object that holds the actual data organized in spherical coordinates
      Method calls are pre processed if necessary and then forwarded to this object.
 
- * - inner_bsphere_radius
+ * - rmin
    - |float|
-   - Specifies the relative radius of the inner bounding sphere of the spherical shell
-     that delimits this volume. This must be within [0, 1.0]. 
+   - Radius for the inner limit of the spherical shell. Note that this radius
+     is given in relative units to the [0,1] unit sphere.
+
+ * - rmax
+   - |float|
+   - Radius for the outer limit of the spherical shell. Note that this radius
+     is given in relative units to the [0,1] unit sphere.
+
+ * - fillmin
+   - |float|
+   - Fill value to return for points which lie inside the bounding sphere of
+     the volume but have a radial component smaller than `rmin`.
+
+ * - fillmax
+   - |float|
+   - Fill value to return for points which lie inside the bounding sphere of
+     the volume but have a radial component greater than `rmax`.
 
 This class implements a gridvolume in spherical coordinates. To achieve that, it holds a nested
 gridvolume object and preprocesses method calls accordingly before forwarding them to the nested
--- a/src/plugins/src/volumes/grid_spherical.cpp
+++ b/src/plugins/src/volumes/grid_spherical.cpp
@@ -16,9 +16,9 @@ enum class FilterType { Nearest, Trilinear };
 enum class WrapMode { Repeat, Mirror, Clamp };
 
 /**!
-.. _volume-gridvolume_spherical:
+.. _volume-sphericalgridvolume:
 
-Grid-based volume data source in spherical coordinates(:monosp:`gridvolume_spherical`)
+Grid-based volume data source in spherical coordinates(:monosp:`sphericalgridvolume`)
 ----------------------------------------------------------
 
 .. pluginparameters::
@@ -42,35 +42,45 @@ cover the [0,1] range.
 */
 
 template <typename Float, typename Spectrum>
-class GridVolumeSpherical final : public Volume<Float, Spectrum> {
+class SphericalGridVolume final : public Volume<Float, Spectrum> {
 public:
-    MTS_IMPORT_BASE(Volume, m_to_local, update_bbox)
+    MTS_IMPORT_BASE(Volume, m_to_local, m_bbox)
     MTS_IMPORT_TYPES(Volume, VolumeGrid)
 
-    GridVolumeSpherical(const Properties &props) : Base(props) {
+    SphericalGridVolume(const Properties &props) : Base(props) {
         m_gridvol = props.volume<Volume>("gridvolume", 1.f);
-        m_inner_bsphere_radius_relative = props.get<ScalarFloat>("inner_bsphere_radius", 0.f);
+
+        m_rmin = props.get<ScalarFloat>("rmin", 0.f);
+        m_rmax = props.get<ScalarFloat>("rmax", 1.f);
+        m_rmin_rel = m_rmin / m_rmax;
+
+        m_fillmin = props.get<ScalarFloat>("fillmin", 0.f);
+        m_fillmax = props.get<ScalarFloat>("fillmax", 0.f);
+
+        m_to_local = props.get<ScalarTransform4f>("to_world", ScalarTransform4f()).inverse();
+        update_bbox_sphere();
     }
 
     UnpolarizedSpectrum eval(const Interaction3f &it, Mask active) const override {
-        auto p = m_to_local * it.p;
-
-        // transform from the [0:1, 0:1, 0:1] cube to a  [-1:1, -1:1, -1:1] cube
-        ScalarTransform4f to_center = ScalarTransform4f::translate(ScalarPoint3f(-1, -1, -1)) * ScalarTransform4f::scale(2.f);
-        p = to_center * p;
+        Point3f p = m_to_local * it.p;
+        Float p_magnitude = ek::norm(p);
 
         Point3f p_spherical = Point3f(
-            (ek::norm(p) - m_inner_bsphere_radius_relative) / (1.f - m_inner_bsphere_radius_relative),
-            ek::acos(p.z() / ek::norm(p)) / ek::Pi<ScalarFloat>,
-            ek::atan2(p.y(), p.x()) / ek::Pi<ScalarFloat> + 1.f
+            (p_magnitude - m_rmin) / (m_rmax - m_rmin),
+            ek::acos(p.z() / p_magnitude) * ek::InvPi<ScalarFloat>,
+            ek::atan2(p.y(), p.x()) * ek::InvPi<ScalarFloat> + 1.f
         );
 
         Interaction3f it_spherical = it;
         it_spherical.p = p_spherical;
 
-        Mask within_spherical_shell = (m_inner_bsphere_radius_relative <= ek::norm(p) && ek::norm(p) <= 1.0);
-
-        return ek::select(within_spherical_shell, m_gridvol->eval(it_spherical, active), 0.f);
+        return ek::select(p_magnitude < m_rmin,
+            m_fillmin,
+            ek::select(p_magnitude > m_rmax,
+                m_fillmax,
+                m_gridvol->eval(it_spherical, active)
+            )
+        );
     }
 
     ScalarFloat max() const override { return m_gridvol->max(); }
@@ -80,11 +90,18 @@ public:
     MTS_DECLARE_CLASS()
 
 protected:
-    ScalarFloat m_inner_bsphere_radius_relative;
+    ScalarFloat m_center, m_rmin, m_rmax, m_rmin_rel, m_fillmin, m_fillmax;
     ref<Volume> m_gridvol;
+
+    void update_bbox_sphere() {
+        ScalarTransform4f to_world = m_to_local.inverse();
+        ScalarPoint3f a = to_world * ScalarPoint3f(-1.f, -1.f, -1.f);
+        ScalarPoint3f b = to_world * ScalarPoint3f(1.f, 1.f, 1.f);
+        m_bbox = ScalarBoundingBox3f(a, b);
+    }
 };
 
-MTS_IMPLEMENT_CLASS_VARIANT(GridVolumeSpherical, Volume)
-MTS_EXPORT_PLUGIN(GridVolumeSpherical, "GridVolumeSpherical texture")
+MTS_IMPLEMENT_CLASS_VARIANT(SphericalGridVolume, Volume)
+MTS_EXPORT_PLUGIN(SphericalGridVolume, "SphericalGridVolume texture")
 
 NAMESPACE_END(mitsuba)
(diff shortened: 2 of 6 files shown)


leroyvn:
Hi @schunkes, here is a first feedback.

Design-impactful comments:

- For consistency with the naming convention used for the other volume texture plugins in Mitsuba, I think this plugin should be named "spherical [coordinates] grid volume". This leads to the following changes:
  - Rename plugin `sphericalgridvolume`
  - Rename class `SphericalGridVolume`
- If I understand correctly, your implementation uses an internal spherical coordinate system within the unit (*i.e.* [-1, 1]) cube. This should reflect in how you set the `to_local` transform (which, in practice, you update repeatedly upon calling `eval()` using the `to_center` matrix). The idea is to first compute the local spherical coordinates of `it.p`, then to infer the nested texture coordinates (this is where the handling of `rmin` and `rmax` should occur, see next comment).
- The shell extent control interface is not intuitive enough IMO. You can simply have two parameters `rmin` and `rmax` (both in [0, 1], `rmin` default to 0, `rmax` default to 1, with `rmin` < `rmax`) which map 0 and 1 on the radius axis of the nested `gridvolume` to the internal spherical coordinate system.
- Out-of-bound fill values could be user-defined parameters `fillmin` and `fillmax` (default to 0).

Optional ideas:

- Raise if `to_local` is set to a non-default value in the nested plugin.
- Warn (or raise?) if nested plugin is not a `gridvolume`.

Programming stuff:

- Refrain from using `auto` in variable declarations (I know you picked some lines from the Mitsuba code, but Wenzel was pretty clear on this and he has a point).
- Pre-compute frequently used values for efficiency. Notably,`p_magnitude = ek::norm(p)` can be pre-computed in the function scope. See above for the utility of `to_center`.
- For performance: `/ ek::Pi<ScalarFloat>` →`* ek::InvPi<ScalarFloat>`

schunkes:
A few comments on yours: I don't really see the point of having `r_max` and `fillmax` parameters.
About `r_max`: I'd prefer that to be always equal to 1. Its extent would then be set by the overall scale of the `SphericalGridVolume`.
About `fillmax`: Maybe you need to explain the use case for this option. I mean, is there a clear need for a volume that varies only in a certain range but is constant, but non zero, outside of that range? For the inside I can kind of understand the point a bit more...

I think the `r_min` - `r_max` interface would be a good choice for the Eradiate scene element, that defines this spherical atmosphere. 

About the `to_local` and `to_center` transform: 
`to_local` is the inverse of `to_world` and is used to set the bbox that surrounds the Volume. This is therefore necessarily set such that it maps the desired worldspace coordinates to the [0, 1]³ cube. 
However a spherical coordinate system covers the [-1, 1]³ cube, which is why I have to transform the point again, once it is transformed to local coordinates.
I can alternatively just add the transformation from `to_center` directly to `m_to_local` and save the multiplication in `eval`, but I'd prefer to have that invisible to the user.

All the performance comments you had are of course completely valid :)

leroyvn:
> I don't really see the point of having r_max and fillmax parameters.
About r_max: I'd prefer that to be always equal to 1. Its extent would then be set by the overall scale of the SphericalGridVolume.

First point: clarity and expressiveness. Just compare 
```cpp
Point3f(
    (ek::norm(p) - m_inner_bsphere_radius_relative) / (1.f - m_inner_bsphere_radius_relative),
    ...
);
```
with 
```cpp
Point3f(
    (r - rmin) / (rmax - rmin),
    ...
);
```

Second point: low implementation cost, avoids patching in the future if somebody finds it inconvenient. Relying on a transform to set `rmax` can be very difficult (the world `rmin` is also scaled in the process, so it's more computation for the user). Defaults cover sensible use cases, and exotic usage is not particularly encouraged, but if a user want to do it, they can, without us having to do further development.

> About `fillmax`: Maybe you need to explain the use case for this option. I mean, is there a clear need for a volume that varies only in a certain range but is constant, but non zero, outside of that range? For the inside I can kind of understand the point a bit more...

For completeness. Again, implementing this is very cheap, defaults cover the intended use case, and you can build a hollow sphere using `fillmax = 1` and another value inside if you wish (seems very useless for now, of course).

> `to_local` is the inverse of `to_world` and is used to set the bbox that surrounds the Volume. This is therefore necessarily set such that it maps the desired worldspace coordinates to the [0, 1]³ cube.
> However a spherical coordinate system covers the [-1, 1]³ cube, which is why I have to transform the point again, once it is transformed to local coordinates.

Seems quite complicated to me. Can't you override `Volume::update_bbox()` and map to [-1, 1]³ instead? If not, it would at least be good to add a comment explaining the rationale so that your future self or another maintainer can understand that quickly.

schunkes:
Ah well clarity/readability is a good point, I won't argue with that.
And I guess I understand the point of defining the whole thing through two radii. It's more uniform and easier to think about :thinking: 
However, if we want to use `rmax` for scaling, we can't really have a spherical volume outside of `rmax` anymore. But your usecase could be handled by setting `rmax` = `rmin`  and filing the inside, given by `fillmin` with a constant value :thinking: 

About the last point: I can't override `Volume::update_bbox()` unfortunately. I'll document the code properly. :+1:

leroyvn:
> However, if we want to use rmax for scaling, we can't really have a spherical volume outside of rmax anymore. But your usecase could be handled by setting `rmax` = `rmin` and filing the inside, given by fillmin with a constant value 🤔

Those params are not about scaling, they are about defining how the local `r` coordinate is mapped to the radius coordinate in the nested volume. A scaling would basically change the size of the reference cube, and `rmin` and `rmax` define the limits of the spherical shell in the reference cube. The use case I'm mentioning would be handled by setting `rmax` < 1, `fillmax` = *x* and using a  nested `constvolume` with `value` < *x*.

> About the last point: I can't override Volume::update_bbox() unfortunately. I'll document the code properly. 👍

You can't override it and have it called by the constructor of `Volume` indeed. You can still overwrite `m_bbox` in your constructor, though (I just tried it).

src/plugins/src/volumes/grid_spherical.cpp
```diff
@@ -0,0 +1,90 @@
...
+template <typename Float, typename Spectrum>
+class GridVolumeSpherical final : public Volume<Float, Spectrum> {
+public:
+    MTS_IMPORT_BASE(Volume, m_to_local, update_bbox)
+    MTS_IMPORT_TYPES(Volume, VolumeGrid)
+
+    GridVolumeSpherical(const Properties &props) : Base(props) {
+        m_gridvol = props.volume<Volume>("gridvolume", 1.f);
+        m_inner_bsphere_radius_relative = props.get<ScalarFloat>("inner_bsphere_radius", 0.f);
+    }
```
leroyvn, line 53: Constructor with full bounding box control:

```suggestion
    MTS_IMPORT_BASE(Volume, m_to_local)
    MTS_IMPORT_TYPES(Volume, VolumeGrid)

    GridVolumeSpherical(const Properties &props) {
        m_to_local = props.get<ScalarTransform4f>("to_world", ScalarTransform4f()).inverse();
        update_bbox();
        m_gridvol = props.volume<Volume>("gridvolume", 1.f);
        m_inner_bsphere_radius_relative = props.get<ScalarFloat>("inner_bsphere_radius", 0.f);
    }
```
Note that the call to the parent constructor is removed.

And further down, possibly as a public member:

```cpp
void update_bbox() {
    ScalarTransform4f to_world = m_to_local.inverse();
    ScalarPoint3f a = to_world * ScalarPoint3f(-1.f, -1.f, -1.f);
    ScalarPoint3f b = to_world * ScalarPoint3f(1.f, 1.f, 1.f);
    m_bbox = ScalarBoundingBox3f(a, b);
}

**Comment by schunkes** [comment]
> A scaling would basically change the size of the reference cube, and rmin and rmax define the limits of the spherical shell in the reference cube.

But in this case we would again have computational effort for the user, because they have to figure out how the limits on the spherical shell in world coordinates, map to the [0, 1] range. :thinking: 

Manipulating the bbox myself is a good idea :)

**Comment by leroyvn** [comment]
> But in this case we would again have computational effort for the user, because they have to figure out how the limits on the spherical shell in world coordinates, map to the [0, 1] range. 🤔

Yes, but it is directly inferred from the scaling set in the transform matrix. Let's say I want [0, 1] to be mapped to [1, 2] with a relevant enclosing volume of size 3:

1. I can set `rmax = 1`, `rmin = 0.5`, and `to_world = scale(2)`; or
2. I can set `rmax = 2/3`, `rmin = 1/3` and `to_world = scale(3)`.

Solution 2 reflects better what happens in world coordinates IMO.

**Comment by schunkes** [comment]
I am not sure I understand, what **relevant enclosing volume** means :thinking:
Clearly in both cases the spherical shell between 1 and 2 units radially is covered by this, but in one case you have extra volume on the outside that you don't need :man_shrugging:

**Comment by leroyvn** [comment]
> I am not sure I understand, what relevant enclosing volume means 🤔

I mean that physically, my radius coordinate goes from 0 to 3.

> Clearly in both cases the spherical shell between 1 and 2 units radially is covered by this, but in one case you have extra volume on the outside that you don't need 🤷‍♂️

You don't, volume texture coordinates are not "volume" per se. You can actually attach the aforementioned volume definition to an arbitrary stencil (a cube, a sphere, a rabbit-shaped mesh) which, itself, will define this "unneeded volume".

**Comment by schunkes** [comment]
Ahh! Well okay.

**Comment by schunkes** [comment]
Alright, I implemented your suggestions for the interface. I just wasted some time being confused by the transformations until I scrapped it all and it worked :man_facepalming: Anyway, this is much simplified now. I'll add tests on monday.

**Comment by leroyvn** [comment]
> PS: The call to the Base constructor is still there, because it has to. When I removed it, the compiler complained.

Indeed, seems like I overlooked that!

**Comment by schunkes** [comment]
I added a few tests but they only work properly with the patched mitsuba2-next version. I did that locally on my machine for now, but I expect the RGL guys to either fix it quickly or accept a PR from me, it is a very minor change after all.

**Comment by leroyvn** [comment]
I may be wrong, but it doesn't look like a very minor change to me: it has a huge impact on the upstream data, doesn't it? From my perspective, such a huge bug should be caught quickly given how intensively the volume API is use by the RGL team for their research.

Data ordering, by the way, should be impacted by the data generation procedure. I just patched ours to use the `VolumeGrid` interface, which the now-standard way to write data files. Would you mind trying producing your data files with it to see if this has an impact? (see `gridvolume.py` for examples)

**Comment by schunkes** [comment]
Ah but this is not actually a problem of the binary file. That one works as inteded. It is actually a bug in the way, the indices for evaulating the file are generated.

**Comment by schunkes** [comment]
If you look at the code in this file: `ext/mitsuba2/resources/data/tests/scenes/participating_media/create_volume_data.py` the data arrays are filled in the order `[z, y, x, channel]` I have the impression that this is intentional. The bug I described to them might not have come up for them, because they always set the number of grid points for all three dimensions to the same value.

**Comment by schunkes** [comment]
Also: I reran my scenes with the updated grid generation method in `gridvolume.py` and the results are exactly the same.

File: src/plugins/src/volumes/grid_spherical.cpp [hunk 2]
```diff
@@ -0,0 +1,125 @@
...
+        m_fillmin = props.get<ScalarFloat>("fillmin", 0.f);
+        m_fillmax = props.get<ScalarFloat>("fillmax", 0.f);
+
+        m_to_local = props.get<ScalarTransform4f>("to_world", ScalarTransform4f()).inverse();
+        update_bbox_sphere();
+    }
+
+    UnpolarizedSpectrum eval(const Interaction3f &it, Mask active) const override {
+        Point3f p = m_to_local * it.p;
+        Float p_magnitude = ek::norm(p);
```
leroyvn, line 85: For clarity, how about calling this one `r`?

src/plugins/src/volumes/grid_spherical.cpp
```diff
@@ -0,0 +1,125 @@
...
+
+        Point3f p_spherical = Point3f(
+            (p_magnitude - m_rmin) / (m_rmax - m_rmin),
+            ek::acos(p.z() / p_magnitude) * ek::InvPi<ScalarFloat>,
+            ek::atan2(p.y(), p.x()) * ek::InvTwoPi<ScalarFloat> + 0.5
```
leroyvn, line 90:
```suggestion
            ek::atan2(p.y(), p.x()) * ek::InvTwoPi<ScalarFloat> + .5f
```

src/plugins/src/volumes/CMakeLists.txt
```diff
@@ -0,0 +1,3 @@
+set(MTS_PLUGIN_PREFIX "volumes")
+
+add_plugin(grid_spherical grid_spherical.cpp)
```
leroyvn, line 3: For consistency with the other Mitsuba plugins, I suggest:
```suggestion
add_plugin(sphericalgridvolume sphericalgrid.cpp)
```
(just like in the docstring)

src/plugins/src/volumes/grid_spherical.cpp
```diff
@@ -0,0 +1,125 @@
...
+
+NAMESPACE_BEGIN(mitsuba)
+
+enum class FilterType { Nearest, Trilinear };
+enum class WrapMode { Repeat, Mirror, Clamp };
+
+/**!
+.. _volume-sphericalgridvolume:
+
+Grid-based volume data source in spherical coordinates(:monosp:`sphericalgridvolume`)
```
leroyvn, line 21:
```suggestion
Grid-based volume data source in spherical coordinates (:monosp:`sphericalgridvolume`)
```

tests/test_plugins/test_volumes/test_grid_spherical.py
```diff
@@ -0,0 +1,114 @@
...
+    it_outside_max = Interaction3f()
+    it_outside_max.p = p_outside_max
+    it_outside_min = Interaction3f()
+    it_outside_min.p = p_outside_min
+    it_inside = Interaction3f()
+    it_inside.p = p_inside
+
+    assert volume.eval(it_outside_max) == 1.0
+    assert volume.eval(it_outside_min) == 3.0
+    assert volume.eval(it_inside) == 2.0
```
leroyvn, line 74: I suggest to refactor this and use a for loop.

tests/test_plugins/test_volumes/test_grid_spherical.py
```diff
@@ -0,0 +1,114 @@
...
+    results = []
+    for angle in angles:
+        p = [0.5*np.cos(np.deg2rad(angle)), 0.5*np.sin(np.deg2rad(angle)), 0.0]
+        print(f"Angle: {angle}, Point: {p}")
+        it = Interaction3f()
+        it.p = p
+        results.append(volume.eval(it))
+
+    # This order looks strange, but the output of enokis atan2 function
+    # does not map the [0: 2pi] range to [0:1] but is shifted by pi
```
leroyvn, line 113: So it maps [-π, π], if I understand correctly?

tests/test_plugins/test_volumes/test_grid_spherical.py
```diff
@@ -0,0 +1,114 @@
...
+                "filter_type": "nearest"
+            }
+        }
+    )
+
+    # have one point in each sector of the spherical volume
+    angles = [5, 65, 125, 185, 245, 305]
+    results = []
+    for angle in angles:
+        p = [0.5*np.cos(np.deg2rad(angle)), 0.5*np.sin(np.deg2rad(angle)), 0.0]
```
leroyvn, line 106: Black pass please ;)

tests/test_plugins/test_volumes/test_grid_spherical.py
```diff
@@ -0,0 +1,114 @@
...
+        print(f"Angle: {angle}, Point: {p}")
```
leroyvn, line 107: Please remove or comment

leroyvn (changes requested):
Hi @schunkes, we're almost done :) I just have a couple of minor comments. Once you're done, please set your PR as ready for review and we'll merge this!

leroyvn:
Thanks @schunkes, let's merge this!

PR #165, opened 2022-01-14
Upgrade Docker build process to work with mitsuba2-next on a local dev env

wint3ria:
# Description

This PR Allows the Docker build to be performed on a local machine with the private mitsuba kernel for eradiate:next

## Limitations
Due to the current private status of the Eradiate kernel on the next branch, these changes do not allow to build the images in github actions or to distribute them through docker Hub. This is intended.

## Github actions changes
The docker github actions now triggers only when there is a tag in the form v*.*.* or build_*. Eradiate developers with sufficient permissions may also trigger the build manually on github.

# Checklist

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -6,11 +6,7 @@ name: Docker build
 on:
   # Triggers the workflow on push or pull request events but only for the main branch
   push:
-    tags: [ v*.*.* ]
-
-  # Trigger every day at 23:42
-  schedule:
-    - cron: '42 23 * * *'
+    tags: [ 'v*.*.*', 'build_*' ]
 
   # Allows you to run this workflow manually from the Actions tab
   workflow_dispatch:
--- a/docker/eradiate-kernel-builder/Dockerfile
+++ b/docker/eradiate-kernel-builder/Dockerfile
@@ -22,10 +22,8 @@ ENV CXX=clang++-9
 COPY . /sources/eradiate
 WORKDIR /sources/eradiate
 RUN make conda-init
-RUN git submodule update --init --recursive
 
-RUN mkdir -p /build/eradiate-kernel                                                                                       \
-    && cd /build/eradiate-kernel                                                                                          \
-    && cmake -GNinja -DPYTHON_EXECUTABLE=$(python3 -c "import sys; print(sys.executable)") /sources/eradiate/ext/mitsuba2 \
-    && ninja                                                                                                              \
-    && tar -cvf eradiate-kernel-dist_${ERADIATE_KERNEL_VERSION}.tar ./dist
+WORKDIR /build/eradiate-kernel
+RUN cmake -GNinja -DPYTHON_EXECUTABLE=$(python3 -c "import sys; print(sys.executable)") /sources/eradiate/ext/mitsuba2
+RUN ninja
+
--- a/docker/eradiate-kernel-builder/Dockerfile
+++ b/docker/eradiate-kernel-builder/Dockerfile
@@ -7,7 +7,6 @@ ARG ERADIATE_KERNEL_VERSION
 ARG NUM_CORES=4
 
 RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y   \
-    cmake                                           \
     git                                             \
     ninja-build                                     \
     clang-9                                         \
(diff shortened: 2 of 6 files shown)


docker/eradiate-kernel-builder/Dockerfile
```diff
@@ -22,10 +22,8 @@ ENV CXX=clang++-9
...
-RUN git submodule update --init --recursive
 
-RUN mkdir -p /build/eradiate-kernel                                                                                       \
-    && cd /build/eradiate-kernel                                                                                          \
-    && cmake -GNinja -DPYTHON_EXECUTABLE=$(python3 -c "import sys; print(sys.executable)") /sources/eradiate/ext/mitsuba2 \
-    && ninja                                                                                                              \
-    && tar -cvf eradiate-kernel-dist_${ERADIATE_KERNEL_VERSION}.tar ./dist
+WORKDIR /build/eradiate-kernel
+RUN cmake -GNinja -DPYTHON_EXECUTABLE=$(python3 -c "import sys; print(sys.executable)") /sources/eradiate/ext/mitsuba2
+RUN ninja
```
leroyvn, line 28: The new build process requires you to use the CMake presets to compile the appropriate variants and disable Embree: https://eradiate.readthedocs.io/en/next/rst/getting_started/install.html#compiling-the-kernel

Also, you no longer need to pass the `PYTHON_EXECUTABLE` variable, pybind11 is smarter than before :)

leroyvn (commented):
Hi @wint3ria, I see one major issue here; you just have to align with the latest build instructions, shouldn't be much trouble.

docker/eradiate-kernel-builder/Dockerfile
wint3ria: This should be fixed now

wint3ria:
I added an environment variable to configure the build parallelism. The default is 4 cores

Users can build everything inside the eradiate repository the following way:

`VERSION=myLocalVersionName NUM_CORES=16 docker-compose -f docker-build.yml build`

wint3ria:
Compilation parallelism is now documented in the Docker guide

.github/workflows/docker.yml
```diff
@@ -6,11 +6,7 @@ name: Docker build
 on:
   # Triggers the workflow on push or pull request events but only for the main branch
   push:
-    tags: [ v*.*.* ]
```
schunkes, line 9: This is more of a question about docker: Does this `tags` entry mean, that only pushes that create tags trigger this build, or is it additional?

schunkes (approved):

leroyvn:
Looks good to me. @wint3ria if you could squash these commits, then we'll merge.

.github/workflows/docker.yml
wint3ria: This line means that any tag using a v*.*.* form (following a globing expansion method) will trigger the docker workflow on github actions. I changed this line to include the tags build_*. We may add other forms if your work on the CI requires it ofc.

We don't get builds for any push (for now). This was intended because I thought it would be overkill. Keep in mind that I try to have the docker images "production ready" (it's a best effort, since eradiate itself is still under dev). So the docker build may be a bit slow to do it in preamble of a CI that you want to trigger at each push. We could reuse parts of the docker work to create a cheaper CI executor.

Another solution is to schedule daily builds and CI runs. This depends on the workflow you want to implement with the CI. We can also think of other triggers that could make sense in the CI. For instance, a build that occurs on any push on specific branches? Use it mostly for PR? We could have a testing branch closely following the main branch, or directly use the main or next branch for this purpose? This really depends on what you want to achieve with the CI. What do you think also @leroyvn ?

I suggest to discuss this on a CI related PR. This one is more focused on the Docker build itself, which happen to be integrated in github actions

wint3ria:
After more testing of the docker images, I found out that the compilation was not using a correct mitsuba variant. This was due to the use of an outdated cmake version (mostly). I am currently correcting this issue, and will squash when my tests pass

wint3ria:
I mention here that I still have 10 tests failing and 9 tests skipped inside the Docker container. I do not think it is related to the installation however. I think the images are usable :)

leroyvn:
> I mention here that I still have 10 tests failing and 9 tests skipped inside the Docker container. I do not think it is related to the installation however. I think the images are ready :)

These issues are batman-related, nothing to do with Docker indeed.

PR #167, opened 2022-01-20
HeterogeneousAtmosphere: Pass the cache_dir parameter to the BlendPha…

### LINKED ISSUE #166: Investigate high /tmp usage on Linux system
It seems that Eradiate does not cleanup its temporary files. It is a problem for long lasting computation as files accumulate quite quickly in `/tmp`. I had more than ~100GB in a night of computations.

https://github.com/eradiate/eradiate/blob/f83994cc8337e65a52eab29231af9f831b96186d/src/eradiate/scenes/atmosphere/_core.py#L367-L371

The update function overwrites the current cache directory location without removing the previous one on disk.
I can't see anywhere else in the codebase a line removing the cachedir.

Eradiate uses the tempfile module to handle the creation of temporary dicts.

https://github.com/eradiate/eradiate/blob/f83994cc8337e65a52eab29231af9f831b96186d/src/eradiate/scenes/atmosphere/_core.py#L4

https://github.com/eradiate/eradiate/blob/f83994cc8337e65a52eab29231af9f831b96186d/src/eradiate/scenes/atmosphere/_core.py#L340-L349

This module does not provide [any mechanism for cleaning up the temporary files.](https://docs.python.org/3/library/tempfile.html#tempfile.mkdtemp)

Did I miss a cleanup mechanism somewhere?
Would it be sufficient and reliable to add a call to `shutil.rmtree` in this `update` function?

wint3ria:
# Description

close #166

Add the cache_dir parameter to the BlendPhaseFunction in the HeterogeneousAtmosphere class.
The constructor was creating a new tmp folder at each call to `kernel_phase`, thus filling the `/tmp` directory.

# Checklist

### DIFF
--- a/src/eradiate/scenes/atmosphere/_heterogeneous.py
+++ b/src/eradiate/scenes/atmosphere/_heterogeneous.py
@@ -397,6 +397,7 @@ class HeterogeneousAtmosphere(AbstractHeterogeneousAtmosphere):
 
             # Construct a blended phase function based on those weighting values
             phase = BlendPhaseFunction(
+                cache_dir=self.cache_dir,
                 components=[component.phase for component in components],
                 weights=sigma_ss,
             )


leroyvn:
Thanks for this fix @wint3ria!

PR #168, opened 2022-01-25
data: Complete rewrite

leroyvn:
This PR refactors the data serving system to allow for more flexibility, improve reliability and reduce the amount of maintenance. Changes are as follows:

* The `data` module is completely redesigned. A `DataStore` interface is now responsible for serving files from a specific location based on relative paths passed to its `fetch()` method. Two initial implementations come with this refactoring, in addition to a composer.
* The `DirectoryDataStore` serves data directly stored in a directory. It is designed to be used with the `resources/data` submodule and implements checks based on a registry: unregistered files cannot be fetched through the data store interface. A tool is shipped to make registry update as simple as possible from a user's standpoint.
* The `OnlineDataStore` serves data stored on the web using the HTTP(S) protocol. The online store can ship a registry to allow for integrity checks, but it also allows serving unregistered, but reachable files. Registered files are meant to be more stable than unregistered files, and it is strongly recommended to only use registered files in tests.
* A `MultiDataStore` composer aggregates a list of `DataStore` instances and queries them successively upon request until one returns a resource. A default data store, accessed as `eradiate.data.data_store`, is configured automatically upon importing Eradiate: it points to the `resources/data` submodule and the Eradiate online store. The URL pointing to the latter can be controlled through the `config.data_store_url` configuration variable and the corresponding environment variable.
* A new `DataError` exception is added and signals when a resource cannot be fetched.
* A new `ertdata` tool supersedes the `ertdownload` script. Its main use is to manage the remote data cache. It also contains a registry generator, which is defined as a pre-commit hook on the `resources/data` submodule.

The downstream code was updated to use this new interface. This sometimes required deeper changes and some refactoring was done on the way. Major changes are as follows:

* The `radprops` subpackage was modified to accomodate the removal of various enums. It was broken down into small submodules with a narrow scope. Some parts are still not updated and now raise a `NotImplementedError`: they will have to be fixed ASAP.
* The `SolarIrradianceSpectrum` class was modified and is now parametrised by data directly. This makes injecting custom spectra much easier.

@nollety I'd greatly appreciate if you could take a look and give me your opinion about the changes I made to the atmosphere, radprops and thermoprops components (I know some things are still broken, I think we'll have to fix them together).

# To do

- [x] Fix / refactor broken components
- [x] Check docstrings
- [x] Add documentation

# Related issues

* Closes eradiate/eradiate-issues#129
* Closes eradiate/eradiate-issues#127

# Checklist


tests/test_eradiate/_system/test_onedim_ckd_vs_mono.py
```diff
@@ -212,8 +212,7 @@ def test_550(reflectance, artefact_dir):
         "H2O-spectra-18300_18400",
     ]:
         test_tools.skipif_data_not_found(
-            "absorption_spectrum",
-            dataset_id,
+            f"spectra/absorption/{dataset_id}.nc",
```
nollety, line 215: Shouldn't that be `f"spectra/absorption/us76_u86_4/{dataset_id}.nc"`?

nollety (commented):
Hi @leroyvn How critical is this PR? It looks great, but I did not go through all of it. About the `radprops` package, did you just copy paste the different `RadProfile` classes in the new modules that you created, or did you also bring changes to these `RadProfile` classes? (it's hard to tell from the diff)

tests/test_eradiate/_system/test_onedim_ckd_vs_mono.py
leroyvn: Most probably, this is typically a part of the code I can't verify since I don't have the data.

leroyvn:
Hi @nollety, this PR is pretty critical: data facilities are to be rewritten for the final delivery. I indeed copied the original code without much change. Some adaptations were required, but nothing significant. The big issue, at this stage, it that the monochromatic evaluation of the AFGL 1986 profile cannot be tested: I don't have the required data.

tests/test_eradiate/_system/test_onedim_ckd_vs_mono.py
nollety: Ah right, well, we won't upload these datasets, they're too big (or perhaps a subset for testing?)

nollety:
I see, I'll try to be as quick as possible to review then

nollety:
The monochromatic absorption data sets for AFGL 1986 are so large (~700 GB) that it's not feasible to distribute them. Even if we distribute a subset of these datasets for testing, the user would not be able to define a AFGL 1986 atmosphere in monochromatic mode. Why don't we simply drop support for monochromatic mode until we can add radis to our dependencies?

tests/test_eradiate/_system/test_onedim_ckd_vs_mono.py
leroyvn: Yes, the new data workflow would make this possible. If those data are small enough, we might even consider shipping them through the data submodule.

leroyvn:
> Why don't we simply drop support for monochromatic mode until we can add radis to our dependencies?

To be discussed at a meeting.

tests/test_eradiate/_system/test_onedim_ckd_vs_mono.py
nollety: OK, I will look into that then

nollety:
About the changes to `eradiate/thermoprops`, this is all fine to me! :heavy_check_mark:

nollety:
About `eradiate/radprops`, everything seems fine to me as well. :heavy_check_mark: 
If I understand correctly, the only thing that is broken is the monochromatic mode for the AFGL 1986 atmospheres, when the user does not provide `absorption_data_sets` for all molecules.

nollety:
I cannot `git submodule update --recursive` on the branch `data_redesign`, did you push all your changes to `eradiate-data`?

nollety:
> I know some things are still broken, I think we'll have to fix them together

as part of this PR, or following it?

nollety:
I agree with the changes to `eradiate/scenes/atmosphere` of course! (this was trivial :laughing:)

nollety:
I'll now take a look at the new `data` module, I hope I'll be done before today's meeting

leroyvn:
Sorry, I forgot to push a commit on the data submodule. This is fixed. Replies below.

> If I understand correctly, the only thing that is broken is the monochromatic mode for the AFGL 1986 atmospheres, when the user does not provide `absorption_data_sets` for all molecules.

I can't tell for sure, I didn't thoroughly inspect the code myself.

> as part of this PR, or following it?

Could be right after.

leroyvn:
I think it's ready for merge now. We'll go through all the atmosphere-related components soon, so no real reason to further delay this.

PR #169, opened 2022-01-25
Change `$ERADIATE_DIR` to `$ERADIATE_SOURCE_DIR`

leroyvn:
# Description

This commit changes a configuration variable from `$ERADIATE_DIR` to `$ERADIATE_SOURCE_DIR`. Purposes are to:

* make it clear that this variable is supposed to point to the root of the source file tree;
* lay ground for making this variable optional (useful to detect whether a development or production setup is used).

I think I've been thorough and didn't miss anything.

@nollety @schunkes @wint3ria please review this PR, it will have a significant impact on your setup.

## Note: Why is this happening?

In a (hopefully) not so distant future, we'll ship Eradiate through PyPI. This means that people won't necessarily clone the repo to use it. Consequently, relying on the fact that we use a dev setup is not a viable option in the long term, in particular for the distribution of data contained in the `resources/data` submodule: without the full submodule, the `DirectoryDataStore` I wrote (see #168) doesn't work.

Fortunately, there is an alternative: if the data submodule is not available, we can just replace the `DirectoryDataStore` with an `OnlineDataStore` (or something very similar) pointing to data files on GitHub. Pooch actually has features to address this specific use case. But still, in order to do so, we need to be able to tell whether we are using a dev setup. Achieving this with an environment-controlled config variable is easy: if `$ERADIATE_DIR` is set, it means we are in dev mode; otherwise, we are in production mode. And just to make it clear that the purpose of this variable is to indicate that we are in dev mode, let's actually name it `$ERADIATE_SOURCE_DIR`!

# Checklist

### DIFF
--- a/docker/eradiate-jupyterlab/Dockerfile
+++ b/docker/eradiate-jupyterlab/Dockerfile
@@ -4,7 +4,7 @@ ARG BASE_IMAGE_VERSION
 FROM ${BASE_IMAGE}:${BASE_IMAGE_VERSION}
 
 ENV PORT=8888
-ENV ERADIATE_DIR=/sources/eradiate
+ENV ERADIATE_SOURCE_DIR=/sources/eradiate
 
 CMD jupyter lab --no-browser --allow-root --port=$PORT --ip='0.0.0.0' --allow_origin='*'
 EXPOSE 8888
--- a/docker/eradiate/Dockerfile
+++ b/docker/eradiate/Dockerfile
@@ -16,7 +16,7 @@ WORKDIR /sources/eradiate
 SHELL ["conda", "run", "-n", "docker", "/bin/bash", "-c"]
 
 ENV LD_LIBRARY_PATH=/mitsuba/ext/mitsuba2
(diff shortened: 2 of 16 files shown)


nollety (approved):
I only went through the diff but did not see anything wrong :)

schunkes (approved):
Looks good :) I checked the branch and there are no leftover ERADIATE_DIR instances.

wint3ria:
Sorry but I will not have much time to review this PR until ~4th of february. I removed myself from the reviewers to avoid blocking the process

leroyvn:
@wint3ria alright, just be aware that this will modify the Docker files. If you prefer, we can delay the merge of this PR, it's not urgent.

leroyvn:
@wint3ria if I merge this by Friday, will it impact your workflow negatively?

wint3ria:
It should be fine, I will need pull the latest changes at some point anyway :)

leroyvn:
Okay, so we'll go with it! Thanks :)

PR #170, opened 2022-01-31
`AFGL1986RadProfile`: Refactor concentration rescaling

leroyvn:
# Description

This PR refactors the concentration rescaling code to simplify the addition of new species. Supporting a new species now simply requires adding a new entry to the *species → quantity computation method* mapping.

I tried to improve code readability with [partials](https://docs.python.org/3/library/functools.html#functools.partial) and by reducing the unit conversion boilerplate. I hope this helps!

I'm not a super fan of the rescaled species detection a.t.m, but fixing that requires more refactoring (the underlying `compute_*()` functions should then be modified).

@nollety if this looks good to you, I think checking out if additional testing is required would be good. I think we already have unit tests, but I might have overlooked something.

# Checklist

### DIFF
--- a/src/eradiate/radprops/_afgl1986.py
+++ b/src/eradiate/radprops/_afgl1986.py
@@ -4,6 +4,7 @@ AFGL (1986) radiative profile definition.
 
 from __future__ import annotations
 
+import functools
 import typing as t
 
 import attr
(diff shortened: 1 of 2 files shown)


nollety (approved):
This looks great! I agree that the way rescaled species are detected is not the best at the moment; I've created [this issue](https://github.com/eradiate/eradiate-issues/issues/146) to address that later. With these changes, the code is already better and easier to read.

leroyvn:
Alright, then I'll merge it. Note that test code should be revisited (and split) as part of eradiate/eradiate-issues#143 .

PR #171, opened 2022-02-04
Regression tests

schunkes:
# Description

This PR introduces a framework to run complex render tests and compare them to reference results.

The `test_tools.py` module, contains methods to evaluate a test based on the RMSE and a chi2 test.
Especially the chi2 test function needs to be reviewed, since it is adapted from mitsuba2.
(`mitsuba.python.chi2`)

There can also be discussion about which metadata should be included in the [0][1] subplot in the plotting method.

[This](https://en.wikipedia.org/wiki/Pearson%27s_chi-squared_test) wikipedia article defines the chi-squared test as it is used here

# Checklist

### DIFF
--- a/src/eradiate/test_tools/regression.py
+++ b/src/eradiate/test_tools/regression.py
@@ -1,0 +1,178 @@
+import os.path
+
+from abc import ABC, abstractmethod
+import attr
+import numpy as np
+import matplotlib.pyplot as plt
+import xarray as xr
+
+from ..typing import PathLike
+
+
+def regression_test_plots(data: xr.Dataset, filename: PathLike, metric: tuple) -> None:
+    """
+    ...
+
+    Parameters
+    ----------
+    data : xarray.Dataset
+        Dataset containing the reference and test result
+
+    filename : path-like
+        Path to the output file for the plot
+
+    metric : tuple
+        A tuple of the form (metric name, value) to be added to the plots.
+    """
+    fig, axes = plt.subplots(2, 2, figsize=(8, 6), dpi=120)
+
+    ref = np.squeeze(data.reference)
+    result = np.squeeze(data.brf)
+    vza = data.vza
+    axes[0][0].plot(vza, ref, label="reference")
+    axes[0][0].plot(vza, result, label="result")
+    axes[0][0].set_title("Reference and test result")
+    handles, labels = axes[0][0].get_legend_handles_labels()
+
+    axes[1][0].plot(vza, result - ref)
+    axes[1][0].set_xlabel("VZA in degrees")
+    axes[1][0].set_ylabel("BRFpp")
+    axes[1][0].set_title("Absolute difference")
+
+    axes[1][1].plot(vza, (result - ref) / ref)
+    axes[1][1].set_title("Relative difference")
+
+    axes[0][1].set_axis_off()
+    axes[0][1].legend(handles=handles, labels=labels, loc="upper center")
+    axes[0][1].text(
+        0.5, 0.5, f"{metric[0]}: {metric[1]:.4}", horizontalalignment="center"
+    )
+
+    plt.tight_layout()
+    plt.savefig(filename)
+    plt.close()
+
+
+@attr.s
+class RenderTest(ABC):
+    """
+    Common interface for tests based on the comparison of a result array against
+    reference values.
+    """
+
+    value: xr.DataArray = attr.ib()
+    reference: xr.DataArray = attr.ib()
+    threshold: float = attr.ib(default=0.05)
+    archive_filename: PathLike = attr.ib(default=None)
+
+    @abstractmethod
+    def run(self) -> bool:
+        """
+        Execute the testing criterion and potentially create archive files.
+
+        Depending on the specialized implementation, a metric will be computed
+        from value and reference and compared to threshold.
+
+        If archive_filename is set, a xarray.Dataset and a summary plot will be
+        created. Otherwise, file creation will be omitted.
+
+        Returns
+        -------
+        bool
+            Result of the test criterion comparison.
+        """
+        pass
+
+
+@attr.s
+class RMSETest(RenderTest):
+    """
+    This class implements a simple test based on the root mean squared
+    error (RMSE) of a result array with respect to the reference data.
+    """
+
+    def run(self) -> bool:
+        value_np = self.value.brf.values
+        ref_np = self.reference.brf.values
+        if np.shape(value_np) != np.shape(self.reference):
+            raise ValueError(
+                f"Result and reference do not have the same shape!\n"
+                f"Got: {np.shape(value_np)}, {np.shape(ref_np)}"
+            )
+
+        result_flat = np.array(value_np).flatten()
+        ref_flat = np.array(ref_np).flatten()
+
+        rmse = np.linalg.norm(result_flat - ref_flat) / np.sqrt(len(ref_flat))
+
+        if self.archive_filename is not None:
+            archive_dataset = self.reference.rename({"brf": "reference"})
+            archive_dataset = archive_dataset.merge(self.value)
+            archive_dirname = os.path.dirname(self.archive_filename)
+            os.makedirs(archive_dirname, exist_ok=True)
+            archive_dataset.to_netcdf(self.archive_filename)
+
+            regression_test_plots(
+                archive_dataset,
+                "".join([self.archive_filename.splitext()[0], ".png"]),
+                ("rmse", rmse),
+            )
+
+        return rmse <= self.threshold
+
+
+@attr.s
+class Chi2Test(RenderTest):
+    """
+    This class implements a statistical test for the regression testing
+    campaign, based on Pearson's Chi-squared test.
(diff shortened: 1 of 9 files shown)


src/eradiate/test_tools.py
```diff
@@ -71,3 +75,142 @@ def missing_artefact(filename: PathLike) -> None:
...
+
+    result_flat = np.array(result).flatten()
+    ref_flat = np.array(reference).flatten()
+
+    rmse = np.linalg.norm(result_flat - ref_flat) / np.sqrt(len(ref_flat))
+
+    return rmse <= rmse_threshold, rmse
+
+
+def assert_chi2(reference, result, significance=0.05):
```
leroyvn, line 123: Maybe sort the arguments consistently with `assert_rmse()`?

src/eradiate/test_tools.py
```diff
@@ -71,3 +75,142 @@ def missing_artefact(filename: PathLike) -> None:
...
+
+    if p_value < significance:
+        passed = False
+    else:
+        passed = True
+
+    return passed, p_value
+
+
+def create_archive_files(path: PathLike, filename: str, data: xr.Dataset):
```
leroyvn, line 150: I generally find names such as `create_foo()` to be not very informative. But anyway, all this function does is call `to_netcdf()`, which, IMO, isn't really worth a dedicated function. The dir creation can be turned into a one-liner thanks to the `exist_ok` kwarg of [`os.makedirs()`](https://docs.python.org/3/library/os.html#os.makedirs).

src/eradiate/test_tools.py
```diff
@@ -71,3 +75,142 @@ def missing_artefact(filename: PathLike) -> None:
...
+        Result data to be stored in a NetCDF file
+    """
+
+    if not os.path.exists(path):
+        os.makedirs(path)
+
+    data.to_netcdf(os.path.join(path, filename))
+
+
+def create_regression_test_plots(data: xr.Dataset, filename: PathLike, metric: tuple):
```
leroyvn, line 174: As previously: maybe something more to-the-point, like (I'll also take the opportunity to fix type hints)
```suggestion
def regression_test_plots(data: xr.Dataset, filename: PathLike, metric: t.Tuple[str, float]):
```

src/eradiate/test_tools.py
```diff
@@ -71,3 +75,142 @@ def missing_artefact(filename: PathLike) -> None:
...
+    data : xarray.Dataset
+        Dataset containing the reference and test result
+
+    filename : path-like
+        Absolute path to the output file for the plot
+
+    metric : tuple
+        A tuple of the form (metric name, value) to be added to the plots.
+    """
+    eradiate.plot.set_style()
```
leroyvn, line 190: This is done by `conftest.py` and you don't want to force it at this level.
```suggestion
```

src/eradiate/test_tools.py
```diff
@@ -71,3 +75,142 @@ def missing_artefact(filename: PathLike) -> None:
...
+    axes[1][1].set_title("Relative difference")
+
+    axes[0][1].set_axis_off()
+    axes[0][1].legend(handles=handles, labels=labels, loc="upper center")
+    axes[0][1].text(
+        0.5, 0.5, f"{metric[0]}: {metric[1]:.4}", horizontalalignment="center"
+    )
+
+    plt.tight_layout()
+    plt.savefig(filename)
```
leroyvn, line 216: Don't forget to close that figure:
```suggestion
    plt.savefig(filename)
    plt.close(fig)
```

leroyvn (commented):
Hi @schunkes, I don't have much to say about your implementation of Pearson's chi squared test, it seems fine to me. You'll have to document it properly so that readers understand well that you're testing for goodness of fit.

I made a couple of small comments, more on implementation details. On the design side, and with the prospect of having a somewhat uniform interface for all those tests, I think having different return types for each `assert_*()` function can be impractical. Instead, turning those functions into small objects with a log buffer (containing messages and reported metric as a string, just like Mitsuba's `ChiSquaredTest` class) would likely be cleaner and allow for swapping tests more easily . For instance, you could have something like
```python
# Some test code
...
test = RMSETest(values, reference, threshold=1e-3)
assert test.run()
```
and `run()` would output log contents in case of failure.

If you happen to decide that you actually want to run a different kind of test, all you have to do is change the test line—the reporting will follow.

tests/test_eradiate/regression/romc/test_het04.py
```diff
@@ -0,0 +1,97 @@
...
+        integrator={"type": "path", "max_depth": -1},
+    )
+
+    exp.run()
+    result = exp.results["measure"]
+
+    reference_path = data_store.fetch("tests/canopies/HET04/het04_brfpp_ref.nc")
+    reference = xr.load_dataset(reference_path)
+
+    passed, rmse = assert_rmse(result.brf.values, reference.brf.values)
```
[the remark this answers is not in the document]
schunkes: I can't do that because I use the rmse value inside the archive and plot generation routine.

tests/test_eradiate/regression/romc/test_het01.py
```diff
@@ -0,0 +1,88 @@
...
+    reference = xr.load_dataset(reference_path)
+    archive_path = metadata.get("archive_path", None)
+
+    if archive_path:
+        datestring = session_timestamp.strftime(format="%Y%m%d")
+        timestring = session_timestamp.strftime(format="%H%M%S")
+        archive_filename = f"{datestring}-{timestring}-het01.nc"
+        archive_filename = os.path.join(archive_path, archive_filename)
+    else:
+        archive_filename = None
```
leroyvn, line 79:
```suggestion
archive_filename = os.path.join(
    archive_path, f"{session_timestamp:%Y%m%d-%H%M%S}-het01.nc"
) if archive_path else None
```
(to be confirmed)

See also

- https://www.python.org/dev/peps/pep-3101/#controlling-formatting-on-a-per-type-basis
- https://docs.python.org/3/library/datetime.html#datetime.date.__format__

src/eradiate/test_tools/regression.py
```diff
@@ -0,0 +1,232 @@
+import os.path
+
+import numpy as np
+import matplotlib.pyplot as plt
+import xarray as xr
+
+from eradiate.typing import PathLike
```
leroyvn, line 7: Please use relative imports in package code.

tests/test_eradiate/conftest.py
```diff
@@ -59,3 +59,10 @@ def ert_seed_state():
     from eradiate.rng import SeedState
 
     return SeedState(0)
+
+
+@pytest.fixture(scope="session")
+def session_timestamp():
+    from datetime import datetime
+
+    return datetime.now()
```
leroyvn, line 68: Maybe this should go one level up? Or do you want this fixture to be restricted to non-plugin tests?

leroyvn (changes requested):
Hi @schunkes, this is taking a nice shape. Some general feedback hereafter.

- Thanks for scoping regression testing-oriented components in a module.
- Your classes should be written with attrs (the rest of our codebase assumes so, and we have infrastructure to facilitate their documentation).
- Your test classes should inherit from a parent interface. At this point, it's just a simple `Test` ABC, with a single `run()` abstract method.
- Old code should be deleted. Much of the docs can be transferred to class field docs.
- Docstrings need some polishing.

src/eradiate/test_tools/regression.py
```diff
@@ -0,0 +1,160 @@
...
+from abc import ABC, abstractmethod
+import attr
+import numpy as np
+import matplotlib.pyplot as plt
+import xarray as xr
+
+from ..typing import PathLike
+
+
+def regression_test_plots(data: xr.Dataset, filename: PathLike, metric: tuple):
```
leroyvn, line 12:
```suggestion
def regression_test_plots(data: xr.Dataset, filename: PathLike, metric: tuple) -> None:
```

leroyvn (changes requested):
Hi @schunkes, we're almost there. I made a pass on the `regression` module and made specific comments, mostly related with documentation. I also see duplicate code between the two concrete test classes which maybe could be pulled up to the `RenderTest` class; we'll see about that later.

I think the reference data should be gathered in a dedicated directory of the `data`submodule rather than thematically sorted. This kind of storage is more flexible: multiple tests may use the same data, and tests may use data coming from different sources.

I trust there is a logic associating test names, artefact names and reference file names? Is there automation for this yet? This is probably something we'll have to think about implementing in a not-too-distant future.

schunkes:
About the data submodule: I'll update the structure tomorrow.

About file names: At the moment the pattern is rather loose and now that you mentioned it, it does actually not work too well for the archive files that are placed in a single output directory. I'll think about this and update that as well.

schunkes:
Alright, I reorganized the data submodule and updated the tests accordingly. Additionally I renamed the folders in the data repo to be all lower case, for consistency.

leroyvn:
Alright, that seems good now. I'll take a few moments to pimp the branch and PR a little, then I'll merge this. Thanks @schunkes!

leroyvn:
Let's merge this. It's missing a developer manual entry but we can add this a little later.

PR #180, opened 2022-03-08
ckd datasets update

nollety:
# Description

Resolves eradiate/eradiate-issues#102

# Changes

* updated ckd absorption data sets, with the following main features:
  * based on `HITRAN2020` (previously a mix between HITRAN2016 and HITRAN2020)
  * correct use of ozone photodissociation absorption data (previously incorrect, affecting the range `400-650nm` in particular)
  * new `CO2` dimension, giving the carbon dioxide volume mixing ratio at sea level (representative of the amount of carbon dioxide in the entire atmosphere)
  * absorption data correspond to the 10-molecule version of the AFGL (1986) atmospheres, namely `H2O`, `CO2`, `O3`, `N2O`, `CO`, `CH4`, `O2`, `NO`, `SO2`, `NO2` (previously 7 molecules only were included)
* you can use the AFGL (1986) `us_standard` and `midlatitude_summer` variants (previously only `us_standard`)
* you can rescale `H2O`, `CO2` and `O3` (previously only `H2O` and `O3`)

## Notes about the FTP server update

I uploaded the new ckd data sets (gzipped) on our ftp server here: https://eradiate.eu/data/store/unstable/ckd/absorption.
The total upload size is 1.2 GB.
In order to not break previously running eradiate versions, I did not overwrite the previous ckd data sets. Instead, I appended a version number to the new data set filenames.
`src/eradiate/radprops/_util_ckd.py.PATHS` was updated with the addition of these filenames.

## Notes about the data submodule update

I updated the AFGL (1986) thermophysical properties data sets to add an `identifier` attribute. This identifier attribute is aligned with how the corresponding ckd absorption data sets are named.

# Checklist

- [ ] The feature branch is rebased on the current state of the `next` branch

:warning: we must pull eradiate/eradiate-data#6 before pulling this

### DIFF
--- a/src/eradiate/radprops/_afgl1986.py
+++ b/src/eradiate/radprops/_afgl1986.py
@@ -22,6 +22,7 @@ from ..thermoprops import afgl_1986
 from ..thermoprops.util import (
     compute_column_mass_density,
     compute_column_number_density,
+    compute_volume_mixing_ratio_at_surface,
 )
 from ..units import to_quantity
 from ..units import unit_registry as ureg
@@ -102,11 +103,13 @@ class AFGL1986RadProfile(RadProfile):
         if not self.has_absorption:
             return ureg.Quantity(np.zeros(self.thermoprops.z_layer.size), "km^-1")
 
-        with _util_ckd.open_dataset(f"afgl_1986-us_standard-{bin_set_id}") as ds:
+        ds_id = f"{self.thermoprops.attrs['identifier']}-{bin_set_id}-v3"
+        with _util_ckd.open_dataset(ds_id) as ds:
             # This table maps species to the function used to compute
             # corresponding physical quantities used for concentration rescaling
             compute = {
                 "H2O": functools.partial(compute_column_mass_density, species="H2O"),
+                "CO2": functools.partial(compute_volume_mixing_ratio_at_surface, species="CO2"),
                 "O3": functools.partial(compute_column_number_density, species="O3"),
             }
 
--- a/src/eradiate/thermoprops/util.py
+++ b/src/eradiate/thermoprops/util.py
@@ -137,6 +137,26 @@ def compute_mass_density_at_surface(ds, species):
     return m * compute_number_density_at_surface(ds=ds, species=species)
 
 
+def compute_volume_mixing_ratio_at_surface(ds, species):
+    """Compute the volume mixing ratio at the surface of a given species in an
+    atmospheric profile.
+
+    Parameters
+    ----------
+    ds : Dataset
+        Atmosphere thermophysical properties data set.
+
+    species : str
+        Species.
+
+    Returns
+    -------
+    quantity
+        Volume mixing ratio at the surface.
+    """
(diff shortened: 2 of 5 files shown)


nollety:
At the moment, the PR does not pass the following regression tests:

* `tests/02_eradiate/03_regression/atmospheres/test_rpv_afgl1986.py`
* `tests/02_eradiate/03_regression/atmospheres/test_rpv_afgl1986_continental.py`

nollety:
Since these regression tests run at 550 nm and include an absorbing molecular atmosphere, they are affected by the "ozone bug" which this PR aims to resolve.

leroyvn:
> Since these regression tests run at 550 nm and include an absorbing molecular atmosphere, they are affected by the "ozone bug" which this PR aims to resolve.

Hi @nollety, this is the whole point of such tests. We'll check the new results and establish them as the reference.

nollety:
Below are plots where I compare the regression tests results (`old` denotes the reference results and `new` denotes the current results, with the new ckd data sets):

* `tests/02_eradiate/03_regression/atmospheres/test_rpv_afgl1986.py`:
test_rpv_afgl1986

* `tests/02_eradiate/03_regression/atmospheres/test_rpv_afgl1986_continental.py`:
test_rpv_afgl1986_continental

nollety:
@leroyvn, if you agree, this will update the reference results used in regression tests:
https://github.com/eradiate/eradiate-data/pull/6

src/eradiate/thermoprops/util.py
```diff
@@ -137,6 +137,26 @@ def compute_mass_density_at_surface(ds, species):
     return m * compute_number_density_at_surface(ds=ds, species=species)
 
 
+def compute_volume_mixing_ratio_at_surface(ds, species):
```
leroyvn, line 140: Type hints

src/eradiate/radprops/_afgl1986.py
```diff
@@ -22,6 +22,7 @@
 from ..thermoprops.util import (
     compute_column_mass_density,
     compute_column_number_density,
+    compute_volume_mixing_ratio_at_surface,
```
leroyvn, line 25: Side comment: I think this naming pattern ("compute...") is heavy and could be made more concise. After all, the fact that a computation occurs is made obvious, and this actually breaks the expressiveness of code using your functions. For example, compare
```python
x = compute_volume_mixing_ratio_at_surface(...)
```
and
```python
x = volume_mixing_ratio_at_surface(...)
```
This is much nicer to read IMO.

leroyvn (commented):
Hi @nollety, not too much to say on my end. I made a couple of comments.

The data directory will have to be cleaned up eventually. If you want to make current data permanent, I think it's worth considering a transfer to the `stable` data store.

leroyvn:
Sorry, I forgot to tick the right box: I approve the PR for merge after discussing points I raised.

leroyvn:
Thanks @nollety, let's merge that.

PR #186, opened 2022-03-11
Added a bounding box for proper scaling of the blend phasefunction

schunkes:
# Description

This PR fixes the "seam bug", by ensuring the bounding box for blend phase function objects is set to match the underlying atmosphere.

Please check if the bbox I define for the spherical shell geometry case is correct. I was not completely certain about this.

# Checklist

### DIFF
--- a/src/eradiate/scenes/atmosphere/_heterogeneous.py
+++ b/src/eradiate/scenes/atmosphere/_heterogeneous.py
@@ -14,8 +14,9 @@ import xarray as xr
 from ._core import AbstractHeterogeneousAtmosphere, atmosphere_factory
 from ._molecular_atmosphere import MolecularAtmosphere
 from ._particle_layer import ParticleLayer
-from ..core import KernelDict
+from ..core import BoundingBox, KernelDict
 from ..phase import BlendPhaseFunction
+from ..shapes import CuboidShape, SphereShape
 from ..._util import onedict_value
 from ...attrs import documented, parse_docs
 from ...contexts import KernelDictContext, SpectralContext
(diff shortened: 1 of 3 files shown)


src/eradiate/scenes/atmosphere/_heterogeneous.py
```diff
@@ -363,10 +364,21 @@ def kernel_phase(self, ctx: KernelDictContext) -> KernelDict:
...
             # Construct a blended phase function based on those weighting values
+            shape = self.eval_shape(ctx)
+            if isinstance(shape, CuboidShape):
+                shape_min = shape.center - shape.edges * 0.5
+                shape_min[2] = self.bottom
+                shape_max = shape.center + shape.edges * 0.5
+            elif isinstance(shape, SphereShape):
+                length_units = ucc.get("length")
+                shape_min = [0, 0, 0] * length_units
+                shape_max = [1, 1, shape.radius.m_as(length_units)] * length_units
```
leroyvn, line 375: I must admit: I haven't tested this with the spherical shell geometry. Are you certain that this will be correct? If not, I'd rather have this raise.

Other shapes should be explicitly unsupported. Our current way to do so is to raise a `RuntimeError`:
```suggestion

            elif isinstance(shape, SphereShape):
                length_units = ucc.get("length")
                shape_min = [0, 0, 0] * length_units
                shape_max = [1, 1, shape.radius.m_as(length_units)] * length_units
                
            else:
                raise RuntimeError  # Only cuboids and spheres are allowed currently
```

leroyvn (commented):
Hi @schunkes, thanks for investigating and solving this. I have a couple of comments/questions, only minor changes.

src/eradiate/scenes/atmosphere/_heterogeneous.py
schunkes: Other shapes are excluded by `eval_shape()`, I had a check here at first too, but it can actually not be anything else, I think.

leroyvn: Having this `else` here is still free and indicates that special care must be taken when adding a new shape, it's important that this unhandled case raises (if `eval_shape()` is fixed but not this method, then something can go silently bad).

schunkes:
I added a safety check :)

leroyvn (approved):
Good, let's go with that!

leroyvn:
Ah—just a detail: can you squash before we merge?

leroyvn:
Thanks :) Merging now!

PR #187, opened 2022-03-14
Compute King correction factor from Bates (1984) data

nollety:
# Description

Resolves eradiate/eradiate-issues#138 and eradiate/eradiate-issues#8

Computation of the Rayleigh scattering coefficient involves the King correction factor which depends on wavelength. So far this King correction factor was set to the constant value of 1,049 (dimensionless). This pull request suggests to compute the King correction factor by interpolating the data published by Bates (1984). Bates' data covers the range of wavelength from 200 to 1000 nanometers, therefore it is extrapolated in [1000, 2400] assuming a constant value. This assumption is rough, but since the corresponding scattering coefficient decreases at a rate proportional to the 4th power of the wavelength, it can be justified.
The figure below illustrates the old and new King correction factor spectra. The largest deviation occurs at the shortest wavelength, and is around 3%.

king_correction_factor

The figure below illustrate the corresponding scattering coefficient spectra. The difference between the two curves is barely visible. The largest deviation occurs around 280 nm and is approximately 0.8 %.

scattering_coefficient

## Changes

* remove obsolete function `kf`
* remove obsolete arguments `king_factor` and `depolarisation_ratio` of `compute_sigma_s_air`

# Checklist

### DIFF
--- a/src/eradiate/radprops/rayleigh.py
+++ b/src/eradiate/radprops/rayleigh.py
@@ -1,9 +1,12 @@
 """Functions to compute Rayleigh scattering in air."""
+import typing as t
 
 import numpy as np
 import pint
 from scipy.constants import physical_constants
 
+from .. import data
+from .._util import Singleton
 from ..units import unit_registry as ureg
 
 # Physical constants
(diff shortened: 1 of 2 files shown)


leroyvn (changes requested):
Hi @nollety, thanks for this PR. I have mixed feelings about opening the data upon loading the module: this will trigger a data store query even for people who actually don't use the data. Problem: in order to implement convenient lazy loading, we'd need some extra logic. It could be based on the `Singleton` helper metaclass:

```python
class _BATES_1984_DATA(metaclass=Singleton):
    # Lazy loader for Bates (1984) King correction data
    def __init__(self):
        self._data = None

    @property
    def data(self):
        if self._data is None:
            self._data = data.load_dataset("spectra/optics/bates_1984.nc")
        return self._data
```

and then
```python
BATES_1984_DATA = _BATES_1984_DATA().data
f_left = BATES_1984_DATA.f.values[0]
```

nollety:
Hi @leroyvn, thanks for your feedback, I applied your suggestions and squashed the commits. I double-checked: all tests passed.

PR #188, opened 2022-03-15
Refactor regression testing framework

schunkes:
# Description

So here is my implementation of the features we discussed for the regression test framework.

Closes eradiate/eradiate-issues/issues/148

I'd like your thoughts on the way I handled getting the name of the metric, by checking for the type of `self` in the `plot()` method.

# Checklist


leroyvn (changes requested):
Hi @schunkes, here's a first round of comments.

schunkes:
I implemented your feedback and added some tests for the regression test helpers

src/eradiate/test_tools/regression.py
```diff
@@ -54,36 +64,152 @@ def regression_test_plots(data: xr.Dataset, filename: PathLike, metric: tuple) -
...
+    def f(value):
+        if isinstance(value, (str, os.PathLike)):
+            try:
+                return xr.load_dataset(value)
+            except (FileNotFoundError, ValueError):
+                return None
+        elif isinstance(value, xr.Dataset):
+            return value
+
+    return f
```
leroyvn, line 92: You can implement this converter statically. Unless `None` is supposed to trigger specific behaviour, I think it's wiser to raise if the file can't be open. We'd end up with this:
```suggestion
def reference_converter(value):
    if isinstance(value, (str, os.PathLike)):
        return xr.load_dataset(value)
    else:
        return value
```

src/eradiate/test_tools/regression.py
```diff
@@ -54,36 +64,152 @@ def regression_test_plots(data: xr.Dataset, filename: PathLike, metric: tuple) -
...
 
-    value: xr.DataArray = attr.ib()
-    reference: xr.DataArray = attr.ib()
-    threshold: float = attr.ib(default=0.05)
-    archive_filename: PathLike = attr.ib(default=None)
+    value: xr.Dataset = attr.ib(validator=attr.validators.instance_of(xr.Dataset))
+    reference: xr.Dataset = attr.ib(default=None, converter=reference_converter())
+    threshold: float = attr.ib(kw_only=True)
+    archive_filename: PathLike = attr.ib(kw_only=True)
+    METRIC_NAME: str = attr.ib(default=None)
```
leroyvn, line 106: Here, you defined `METRIC_NAME` as a field (instance attribute). To define it as a class attribute, you must not use `attr.ib()`:
```suggestion
    METRIC_NAME = None

    value: xr.Dataset = attr.ib(validator=attr.validators.instance_of(xr.Dataset))
    reference: xr.Dataset = attr.ib(default=None, converter=reference_converter())
    threshold: float = attr.ib(kw_only=True)
    archive_filename: PathLike = attr.ib(kw_only=True)
```

src/eradiate/test_tools/regression.py
```diff
@@ -92,12 +218,12 @@ class RMSETest(RenderTest):
     error (RMSE) of a result array with respect to the reference data.
     """
 
-    def run(self) -> bool:
-        # Inherit docstring
+    METRIC_NAME: str = attr.ib(default="rmse")
```
leroyvn, line 221:
```suggestion
    METRIC_NAME = "rmse"
```

src/eradiate/test_tools/regression.py
```diff
@@ -135,12 +249,11 @@ class Chi2Test(RenderTest):
 
     # The algorithm is adapted from Mitsuba's testing framework.
 
-    def run(self) -> bool:
-        # Inherit docstring
-
-        from mitsuba.python.math import rlgamma
+    METRIC_NAME: str = attr.ib(default="p-value")
```
leroyvn, line 252:
```suggestion
    METRIC_NAME = "p-value"
```

src/eradiate/test_tools/regression.py
```diff
@@ -54,36 +64,152 @@ def regression_test_plots(data: xr.Dataset, filename: PathLike, metric: tuple) -
...
+            Create only a simple visualisation of the computation data
+        """
+        filename = "".join([os.path.splitext(self.archive_filename)[0], ".png"])
+        os.makedirs(os.path.dirname(filename), exist_ok=True)
+        print(f"Saving plot in {filename}")
+        if reference_only:
+            plt.figure(figsize=(8, 6))
+            plt.plot(np.squeeze(self.value.vza), np.squeeze(self.value.brf))
+            plt.xlabel("VZA in degrees")
+            plt.ylabel("BRFpp")
```
leroyvn, line 195:
```suggestion
            plt.ylabel("BRF in principal plane [—]")
```

src/eradiate/test_tools/regression.py
```diff
@@ -54,36 +64,152 @@ def regression_test_plots(data: xr.Dataset, filename: PathLike, metric: tuple) -
...
     Common interface for tests based on the comparison of a result array against
     reference values.
     """
 
-    value: xr.DataArray = attr.ib()
-    reference: xr.DataArray = attr.ib()
-    threshold: float = attr.ib(default=0.05)
-    archive_filename: PathLike = attr.ib(default=None)
+    value: xr.Dataset = attr.ib(validator=attr.validators.instance_of(xr.Dataset))
+    reference: xr.Dataset = attr.ib(default=None, converter=reference_converter())
```
leroyvn, line 103: Here, we can:
* use the updated converter;
* let `None` through using the `optional()` converter;
* apply a validator to be sure that the data passed is of the type we are expecting.

```suggestion
    reference: xr.Dataset = attr.ib(
        default=None, 
        converter=attr.converters.optional(reference_converter), 
        validator=attr.validators.optional(attr.validators.instance_of(xr.Dataset)),
     )
```

leroyvn (changes requested):
Hi @schunkes, this is looking good :) I suggested a couple of changes. It would also be great to add a few words about the recommended workflow in the documentation.

src/eradiate/test_tools/regression.py
schunkes: Hm, I 'm not entirely happy with this. It would make the process of getting a new reference quite annoying, because you'd have to modify every single test case and set the path for the reference to None, then run the tests and put the new file names in every file.

I'd suggest, we raise if the file cannot be read, because it is broken or can't be interpreted as a dataset. But if the file can't be found, we set to None and create a new reference. That way, if you want to create a new reference, you simply delete the old reference file from your local file system and run.
That way you only have to modify the tests themselves once.

schunkes:
About the docs: I couldn't find a section about tests yet, so should I just add one, probably under developer guide?

src/eradiate/test_tools/regression.py
leroyvn: Good point. In that case, the best would be IMO to issue a warning if the file can't be found. That would attract the attention of the user while not preventing the intended behaviour. But how that will articulate with the no-warning policy we currently have is yet to be investigated: filtering the warning would prevent it from making the test fail, but I'm not sure if it would then show up in the test output.

If the warning strategy can't be implemented, then a print would be the second choice.

schunkes:
Two questions about my docs contribution:
I kept writing *users* but I suppose I should write *contributors* in this section of the docs, right?
Second, I wrote that people might want to add the reference results for their regression tests to the data repo, but I was not sure which part of the docs to reference for the proper way to do tha. Can you help me out?

src/eradiate/test_tools/regression.py
leroyvn: Alright, so let's go with the print strategy. About the output being flooded with Mitsuba's progress bars: we could actually suppress all progress as part of `conftest.py`.

docs/rst/contributing.rst
```diff
@@ -402,6 +402,70 @@ and plots will be placed. If the directory does not exist, it will be created.
 The artefact directory defaults to ``./build/test_artefacts``, which is resolved
 relative to the current working directory.
 
+Adding new regression tests
+***************************
+
+Regression tests usually utilize non trivial test scenarios, which means that it
+is not advisable to assess their success on exact matching with a reference result.
+Instead, these tests are based on statistical metrics and we provide helper classes
+that handle computing these statistical metrics.
```
leroyvn, line 411: Following our live discussion, I suggest:

```
Regression tests use a comparison framework providing interfaces for statistical and other metric-based tests. Relevant components are listed in the API reference [:mod:`eradiate.test_tools`].
```

leroyvn (commented):
Following up on our discussion:

* Don't repeat in this manual what you already tell in docstrings (link to API reference and pimp the docstrings instead)
* Doc outline:
  1. Testing framework motivation
  2. Basic principles (run a test case; compare result with reference, comparison protocol implemented by one of the classes)
  3. Special behaviour: test fails because of wrong result (how to retrieve results, how to update the reference), missing reference (new test), archival feature

After some more thinking, I believe we should also change the base class's name from `RenderTest` to `RegressionTest`.

schunkes:
I rebased on the most recent `main`. Additonally I made all methods execept for `run()` private, since they are not supposed to be used directly.

docs/rst/contributing.rst
```diff
@@ -402,6 +402,69 @@ and plots will be placed. If the directory does not exist, it will be created.
...
+        value=result,
+        reference="path/to/the/data-file/reference.nc",
+        threshold=0.05,
+        archive_filename="/path/for/file/output.nc",
+    )
+
+After running a simulation on an Eradiate scene, you provide the resulting dataset as well as a path to
+the reference result to the helper class. Adding a threshold value, which may depend on the scenario and the chosen
+metric, and a path and filename for the outputs generated by the class the test is ready.
+To execute the test it exposes the ``run()`` method, which handles computing the
```
leroyvn, line 436: Maybe use a cross-reference here? Like ```:meth:`.RenderTest.run` ```

docs/rst/contributing.rst
```diff
@@ -402,6 +402,69 @@ and plots will be placed. If the directory does not exist, it will be created.
...
+
+If the test fails due to a significant difference between the reference and the result the output can help in analysis.
+The reference data and the result are stored in two NetCDF files under the path given in ``archive_filename``, which can
+be imported and used in python scripts for detailed analysis. Furthermore the test adds an overview plot made up of four
+parts: A direct visualisation of the result and reference data on the same axis, the absolute and relative difference between
+result and reference in their own axes and the numerical value of the chosen metric.
+
+In case this difference stems from a change made to Eradiate, which significantly alters the code's behaviour, the
+reference needs to be updated. In this case, replace the existing reference file in the data repository and create
+pull request for the maintainers to review and add.
```
leroyvn, line 461:
```suggestion
a pull request for the maintainers to review and add.
```

src/eradiate/test_tools/regression.py
```diff
@@ -54,50 +63,172 @@ def regression_test_plots(data: xr.Dataset, filename: PathLike, metric: tuple) -
...
+
+    def f(value):
+        if isinstance(value, (str, os.PathLike)):
+            if not os.path.isfile(value):
+                return None
+            return xr.load_dataset(value)
+        elif isinstance(value, xr.Dataset):
+            return value
+
+    return f
```
leroyvn, line 90: You still don't need to return a callable:
```suggestion
def reference_converter(value):
    if isinstance(value, (str, os.PathLike)):
        if not os.path.isfile(value):
            return None
        return xr.load_dataset(value)
    elif isinstance(value, xr.Dataset):
        return value
```
What happens if `value` is neither a path-like nor a dataset?

src/eradiate/test_tools/regression.py
```diff
@@ -54,50 +63,172 @@ def regression_test_plots(data: xr.Dataset, filename: PathLike, metric: tuple) -
...
     """
 
-    value: xr.DataArray = attr.ib()
-    reference: xr.DataArray = attr.ib()
-    threshold: float = attr.ib(default=0.05)
-    archive_filename: PathLike = attr.ib(default=None)
+    value: xr.Dataset = attr.ib(validator=attr.validators.instance_of(xr.Dataset))
+    reference: xr.Dataset = attr.ib(
+        default=None,
+        converter=attr.converters.optional(reference_converter()),
```
leroyvn, line 103: With the aforementioned changes:
```suggestion
        converter=attr.converters.optional(reference_converter),
```

tests/02_eradiate/01_unit/test_tools/test_regression.py
```diff
@@ -0,0 +1,172 @@
...
+import pytest
+import xarray as xr
+
+import eradiate
+
+eradiate.set_mode("mono")
+
+import mitsuba as mi
+
+import eradiate.test_tools.regression as tt
```
leroyvn, line 11: isort pass (you might also need to move the `eradiate.set_mode()` call down a bit)

tests/02_eradiate/01_unit/test_tools/test_regression.py
```diff
@@ -0,0 +1,172 @@
...
+            value=xr.Dataset(),
+            threshold=0.05,
+            reference=tempfile,
+        )
+
+
+def test_rmse_evaluate():
+    # test the computation of the RMSE value from given data.
+    # we give the dataset some wrong data fields to ensure the right
+    # we give the dataset some wrong data fields to ensure the right
```
leroyvn, line 89: Repeated line

leroyvn (changes requested):
Hi @schunkes, we're getting there and it's looking good. I made a few more comments. In addition, I'd appreciate if you could document the fields you defined for the `RegressionTest` line as they won't show up in docs otherwise. Note that class attributes should be documented as regular attributes, not using the `documented()` helper.

src/eradiate/test_tools/regression.py
```diff
@@ -54,50 +63,172 @@ def regression_test_plots(data: xr.Dataset, filename: PathLike, metric: tuple) -
...
+        validator=attr.validators.optional(attr.validators.instance_of(xr.Dataset)),
+    )
+    threshold: float = attr.ib(kw_only=True)
+    archive_filename: PathLike = attr.ib(kw_only=True)
+    METRIC_NAME = None
```
leroyvn, line 108: Class attribute: please move to top.

schunkes:
Alright, I updated the converter and I moved the imports in `test_regression.py`

src/eradiate/test_tools/regression.py
```diff
@@ -54,50 +65,191 @@ def regression_test_plots(data: xr.Dataset, filename: PathLike, metric: tuple) -
...
+        doc="Reference result. Can be specified as a dataset or a path to a file which can be loaded into a dataset.",
+        type=":class:`xarray.Dataset`",
+        init_type=":class:`xarray.Dataset` or path-like, optional",
+        default="None",
+    )
+    threshold: float = documented(
+        attr.ib(kw_only=True),
+        doc="Threshold for test evaluation",
+        type="bool",
+        init_type="bool",
```
leroyvn, line 120:
```suggestion
        type="float",
        init_type="float",
```

PR #201, opened 2022-04-04
Using abspath to fix the issue of os.makedirs being passed an empty s…

schunkes:
…tring as directory name.

# Description
Closes eradiate/eradiate-issues/issues/156

The issue was that the `RegressionTest` classes store the artefact path as a `pathlib.Path` object which strips off the leading `./`. Later we call `os.path.dirname` on the stripped Path object, which returns an empty path. This is why the call to `os.makedirs` failed.

# Checklist


leroyvn:
Looks good to me, let's merge this.

PR #207, opened 2022-04-14
Tests: Improve regression testing framework

leroyvn:
# Description

This commit improves the regression testing framework with the following changes:

* Artefact names are now assembled from a test name and an archive directory, both set in the test code.
* The reference data can now be passed as a data store path.
* Spectral results (*e.g.* with more than a single CKD bin) are now handled.
* Various type hint and docstring fixes.

This overall simplifies the test code and improves its robustness by handling cases where the reference is expected to be retrieved from the data store but is still missing (relevant when adding a new regression test case).

**To do**

- [x] Update testing framework tests
- [x] Update change log

# Checklist


schunkes (approved):

schunkes:
Looks good :)

PR #217, opened 2022-04-21
Added uninstall instructions

nollety:
# Description

I've added a section in the installation guide with instructions about how to uninstall Eradiate.

# Checklist

- [ ] ~~The code follows the relevant coding guidelines~~
- [ ] ~~The code generates no new warnings~~
- [ ] ~~The code is tested to prove its function~~


leroyvn (commented):
Hi @nollety, I made a different proposal which, I think, is more portable. What do you think about that?

nollety:
Hi @leroyvn, thanks for the feedback, I think your way is better indeed :+1

PR #225, opened 2022-05-13
Atmosphere: Scaling factor computation dispatch was ambiguous

schunkes:
# Description

The scaling factor for molecular rescaling in the atmospheres can be computed based on different kinds of inputs.
Depending on the unit of these input data, the computation differs. The `thermoprops.util` module calls the corresponding function by checking the units in the provided dataset. This is done based on a dict which maps the unit to a function. The entries in this dict were not unique.

I am not 100% sure the units i suggested here, are correct, because I am not familiar with the atmosphere physics. It does look like it should be this though.

The imports were automatically reordered by my `black` file watcher.

# Checklist

### DIFF
--- a/src/eradiate/thermoprops/util.py
+++ b/src/eradiate/thermoprops/util.py
@@ -3,8 +3,8 @@ Utility functions to manipulate atmosphere thermophysical properties data
 sets.
 """
 
-from datetime import datetime
 import typing as t
+from datetime import datetime
 
 import iapws
 import numpy as np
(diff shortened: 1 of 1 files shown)


leroyvn:
Hi @schunkes, let's wait for Yvan's review, I don't remember all the units which are supposed to be handled.

Prior to merge, it would also be nice if you could make this issue title more descriptive (no rush, this won't go any further before next Monday).

nollety (approved):
Dimensions for mass density are indeed `[mass] * [length]^-3`, thanks for spotting this typo!

PR #226, opened 2022-05-18
`TabulatedPhase`: better eval method

nollety:
# Description

Resolves eradiate/eradiate-issues#165

I improved how `TabulatedPhase.eval` maps the input phase function data to a regular grid of scattering angle cosine (`mu`) when the input data does not already map to a regular `mu` grid.

## Changes

I removed the hidden `_n_mu` class attribute and updated `eval_mono` in the following way:
* if the input phase function `data` maps to a `mu` regular grid, interpolate in wavelength and return the values
* else, compute regular `mu` grid based on the smallest `mu` step found in the input phase function data, and interpolate in wavelength (using `xarray.DataArray.interp`) and on that regular `mu` grid (using `numpy.interp` for performance) and return the values. For further details, refer to a discussion about the choice of this regular `mu` grid in the comments down below.

I added 2 unit tests for this new implementation (hence the data submodule update). 

If the user wants to control the number of points used along the `mu` axis to discretize a tabulated phase function, they would simply have to provide phase function data mapping directly to a regular `mu` grid, since in that case the input `DataArray` is left unchanged.

I added a converter for the `data` attribute, that ensures the `mu` coordinate is monotonically increasing.

I added a validator for the `data` attribute to check that the `mu` coordinate in the input data set goes from `-1.0` to `1.0`.

I added `TabulatedPhaseFunction` to the API Reference.

# Checklist

### DIFF
--- a/src/eradiate/scenes/phase/_tabulated.py
+++ b/src/eradiate/scenes/phase/_tabulated.py
@@ -14,38 +14,83 @@ from ...exceptions import UnsupportedModeError
 from ...units import unit_registry as ureg
 
 
+def _ensure_magnitude_array(q: pint.Quantity) -> pint.Quantity:
+    """Convert to Quantity whose magnitude is an array"""
+    if np.isscalar(q.magnitude):
+        return ureg.Quantity(np.atleast_1d(q.magnitude), q.units)
+    else:
+        return q
+
+
+def _convert_data(da: xr.DataArray) -> xr.DataArray:
+    """
+    If the coordinate mu is not strictly increasing, return a reindexed
+    copy of the DataArray where it is.
+    Else, return the input DataArray.
+    """
+    if np.all(da["mu"].values[1:] > da["mu"].mu.values[:-1]):
+        return da
+    elif np.all(da["mu"].values[1:] < da["mu"].mu.values[:-1]):
+        return da.reindex(mu=da.mu[::-1])
+    else:
+        return da.reindex(mu=np.sort(da.mu.values))
+
+
+def _validate_data(instance, attribute, value):
+    """
+    Bounds of mu coordinate must be [-1.0, 1.0].
+    """
+    mu_min = value["mu"].values.min()
+    mu_max = value["mu"].values.max()
+
+    if not (mu_min == -1.0 and mu_max == 1.0):
+        raise ValueError(
+            f"Coordinate 'mu' bounds must be -1.0, 1.0 (got {mu_min}, {mu_max})."
+        )
+
+
 @phase_function_factory.register(type_id="tab_phase")
 @parse_docs
 @attr.s
 class TabulatedPhaseFunction(PhaseFunction):
-    """
+    r"""
     Tabulated phase function [``tab_phase``].
 
-    A lookup table-based phase function. The `data` field is a
+    A lookup table-based phase function. The ``data`` field is a
     :class:`~xarray.DataArray` with wavelength and angular dimensions.
 
-    .. warning::
-       The phase function is evaluated on a regular grid of scattering angle
-       cosine values from -1 to 1 with a number of grid points set to 20001.
-       If the input `data` contains variations that cannot be resolved by
-       this regular grid, information about these variations will get lost
-       and the accuracy of simulations using the resulting
-       `TabulatedPhaseFunction` object might be impacted.
+    Notes
+    -----
+    The :math:`\mu` coordinate must cover the :math:`[-1, 1]` interval but
+    there is no constraints on the ordering of values nor on the spacing
+    between two consecutive values, i.e. non-regular :math:`\mu` grid are
+    supported. If the :math:`\mu` grid is non-regular and since the
+    underlying mitsuba ``tabphase`` plugin expects phase function values
+    on a regular :math:`\mu` grid, we compute a regular grid with a
+    :math:`\mu` step equal to the small :math:`\mu` step found in the input
+    `data` and interpolate the latter on that regular grid. This can
+    lead to large arrays (several 100 MB). If you want to control
+    the size of the :math:`\mu` grid yourself, you can simply provide a
+    :class:`~xarray.DataArray` object that already has a regular grid
+    :math:`\mu` coordinate ; in such a case, the :class:`~xarray.DataArray`
+    object is not further interpolated along the :math:`\mu` dimension.
+    For better performance, it is best to provide phase function data on a
+    regular (and sorted) :math:`\mu` grid, since neither conversion nor
+    interpolation is performed in that case.
     """
 
     data: xr.DataArray = documented(
         attr.ib(
-            validator=attr.validators.instance_of(xr.DataArray),
+            converter=_convert_data,
+            validator=[attr.validators.instance_of(xr.DataArray), _validate_data],
             kw_only=True,
         ),
         type="DataArray",
-        doc="Value table as a data array with wavelength (``w``) and "
-        "scattering angle cosine (``mu``) as coordinates. This parameter has "
-        "no default.",
+        doc="Value table as a data array with wavelength (``w``), scattering "
+        "angle cosine (``mu``), and scattering phase matrix row "
+        "(``i``) and column (``j``) indices (integer) as coordinates. "
+        "This parameter has no default.",
     )
-
-    # Number of points used to represent the phase function on the angular coordinate
-    _n_mu: int = attr.ib(default=20001, init=False, repr=False)
 
     def eval(self, spectral_ctx: SpectralContext) -> np.ndarray:
         r"""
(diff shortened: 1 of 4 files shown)


nollety:
I found no need to update the regression tests with these changes ; they still pass.

src/eradiate/scenes/phase/_tabulated.py
```diff
@@ -95,15 +107,46 @@ def eval_mono(self, w: pint.Quantity) -> np.ndarray:
...
+                )
+                .data
+            )
+
+        # data does not map to regular grid of scattering angle cosines
+        else:
+            # compute the smallest regular grid that does not loose information
+            # in the sense of Shannon theorem
+            dmu = np.abs(self.data.mu.diff(dim="mu")).values.min()
+            nmu = 2 * int(np.ceil(2.0 / dmu))
```
nollety, line 159: After discussion with @schunkes and @lucio-f, we could consider removing the outside factor 2 (from Shannon's theorem) which would result in lower memory footprint and faster interpolation

nollety:
I'll add a couple of tests

tests/02_eradiate/01_unit/scenes/phase/test_tabulated.py
```diff
@@ -8,17 +8,28 @@
...
+    return result.phase
 
 
-def test_tabulated_basic(modes_all, dataset):
+@pytest.fixture
+def irregular_mu() -> xr.DataArray:
+    """Phase function data with an irregular mu grid."""
+    result = xr.open_dataset(
+        path_resolver.resolve("tests/spectra/particles/random-irregular_mu.nc")
+    ).load()
```
leroyvn, line 25: Getting rid of the path resolver would be nice. Two possibilities:

* use `data.load_dataset()` (this data should be registered to the data store), or
* create it as a fixture, preferable if the data is used only in this test.

src/eradiate/scenes/phase/_tabulated.py
```diff
@@ -14,6 +14,41 @@
...
 
+def _magnitude_as_array(q: pint.Quantity) -> pint.Quantity:
+    """Convert to Quantity whose magnitude is an array"""
+    if isinstance(q.magnitude, np.ndarray):
+        return q
+    else:
+        return ureg.Quantity(np.array([q.magnitude]), q.units)
+
+
+def _convert_data(da: xr.DataArray) -> xr.DataArray:
```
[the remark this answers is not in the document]
nollety: Ooh, I was not aware such decorator exists ([https://github.com/python-attrs/attrs/pull/404](https://github.com/python-attrs/attrs/pull/404)) and I don't find in the attrs documentation, are you sure this is possible?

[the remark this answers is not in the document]
leroyvn: Oops, my bad! No it doesn't, you're right.

src/eradiate/scenes/phase/_tabulated.py
```diff
@@ -23,30 +58,34 @@ class TabulatedPhaseFunction(PhaseFunction):
...
+        "underlying mitsuba ``tabphase`` plugin expects phase function values "
+        "on a regular :math:`\mu` grid, we compute the smallest regular grid "
+        "without information loss in the sense of Shannon's theorem and "
+        "interpolate the data on that regular grid. Nevertheless, this can "
+        "lead to very large arrays (e.g. several GB). If you want to control "
+        "the size of the :math:`\mu` grid yourself, you can simply provide a "
+        ":class:`~xarray.DataArray` object that already has a regular grid "
+        ":math:`\mu` coordinate ; in such a case, the "
+        ":class:`~xarray.DataArray` object is not further interpolated along "
+        "the :math:`\mu` coordinate. This parameter has no default.",
```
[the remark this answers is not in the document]
nollety: I agree

src/eradiate/scenes/phase/_tabulated.py
```diff
@@ -95,15 +134,48 @@ def eval_mono(self, w: pint.Quantity) -> np.ndarray:
...
-                w=w.m_as(self.data.w.units),
-                mu=np.linspace(-1, 1, self._n_mu),
+        w_units = self.data.w.attrs["units"]
+
+        # data already maps to regular grid of scattering angle cosines
+        if np.allclose(
+            self.data.mu.values,
+            np.linspace(-1.0, 1.0, self.data.mu.values.size),
+            rtol=1e-9,
+        ):
```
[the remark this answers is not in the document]
nollety: Hmm, I did not think at all of this naive approach, indeed, much faster and much simpler!

src/eradiate/scenes/phase/_tabulated.py
nollety: Any thoughts about that? @leroyvn

src/eradiate/scenes/phase/_tabulated.py
```diff
@@ -95,15 +134,48 @@ def eval_mono(self, w: pint.Quantity) -> np.ndarray:
...
+            data_w_interpolated = self.data.sel(i=0, j=0).interp(
+                w=_w.m_as(w_units),
                 kwargs=dict(bounds_error=True),
             )
-            .data
-        )
+
+            # for performance, interpolation on mu dimension is performed with
+            # numpy.interp, which is found to be more than twice faster than
+            # xarray.DataArray.interp
```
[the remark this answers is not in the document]
leroyvn: Not really, it's just something I have in the back of my mind. Currently, the scene assembly is a bottleneck in some situations (like computing the TOA radiance on a 1D scene with an aerosol layer and a molecular atmosphere in CKD mode with a low sample number) and interpolation takes a lot of time. But it's not the only thing which slows us down.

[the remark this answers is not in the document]
nollety: Okay, I'll keep that in mind then^^

src/eradiate/scenes/phase/_tabulated.py
leroyvn: I'm not a signal processing expert so I don't have strong arguments to oppose: if you guys have proper arguments in favour of tuning down the number of grid points, go for it. Just keep that in mind when analysing results ;)

nollety:
Thanks for the feedback! @leroyvn

src/eradiate/scenes/phase/_tabulated.py
schunkes: Well Lucio made the point that the data is already sampled at a maximum frequency of dmu, so we don't really gain anything from resampling it at 2*dmu.

nollety: For the record, my reasoning is that the smallest mu step is likely to be found at the extreme right of the [-1, 1] `mu` interval since this is where the forward scattering peak is located. Therefore, if we buid the regular `mu` grid using:

```python
dmu_min = np.abs(self.data.mu.diff(dim="mu")).values.min()
nmu = int(np.ceil(2.0 / dmu_min)) + 1
mu = np.linspace(-1, 1, nmu)
```
the two far-right mu points in [-1, 1] will exactly match the `mu` points in the input data. From these points on, the data to the left will be interpolated, therefore we incur a loss accuracy. Arguably, the loss of accuracy is minimised where it is important to be accuracy, namely close to `mu=1.0`.

nollety: @schunkes I am not convinced that:

> dmu is the minimal mu step

implies that:

> 1 / dmu is the maximal frequency in the phase function signal

If you take the fft of a realistic particle phase function in the mu-space, you will find an infinite number of harmonics in my opinion

nollety: After more discussion with the whiteboard, we found that assuming the following:
* the input data is propertly discretised, namely the smallest mu-step is found where the phase function has the sharpest variations
* the sharpest feature occurs either at `mu=-1` or at `mu=1.0`, or if it is located somewhere in the middle of the interval, the input data is well enough sampled in the `mu` space
the smallest delta mu regularisation method is the best.

As illustrated by the figure below, in the presumably most common case where the phase function has sharpest variations around mu=1.0, interpolating on the regular grid prescribed by the Shannon's theorem or on the regular grid using the smallest mu-step lead to the same result:

phase_mu_interp

nollety: As a rule of thumb, the smallest the mu step in the input data, the lowest the interpolation errors.

We note that this regular grid tabulated phase function approach is not best suited for phase function with very sharp features, generally speaking, because of the overhead in memory and in interpolation time.

nollety: However, this approach is the only available to us at the moment...

nollety:
Alright, we are ready for another review round I think :)

leroyvn:
Thanks @nollety, we're good to go: I'll merge this.

PR #229, opened 2022-05-30
Add irregularly gridded tabulated phase function support

leroyvn:
# Description

This PR adds support for irregularly gridded phase functions. Angular regridding, required to reach appropriate accuracy with aerosol phase functions and implemented as part of #226, resulted in an important drop in performance due to the extra interpolation overhead. 

The `TabulatedPhaseFunction` kernel dictionary generation no longer triggers angular regridding. Instead, it checks whether the angular grid is regular or irregular and creates a different kernel dictionary for each case. See below for the complete list of changes:

- Added `tabphase_irregular` plugin.
- Modified `TabulatedPhaseFunction.eval_mono()` and `TabulatedPhaseFunction.kernel_dict()` to replace angular regridding with plugin selection.
- Minor refactoring of `TabulatedPhaseFunction` unit tests.

**Performance considerations**

[before]: https://user-images.githubusercontent.com/34740232/171014209-a8ee3fd0-02e5-4a68-ac6e-da59cc12804c.png "Before" 
[after]: https://user-images.githubusercontent.com/34740232/171014232-13ec1567-1acc-4f2f-8dca-9edee27c3fd6.png "After"

|   | Run time |
|---|---|
| Before (regridding) | ![][before] |
| After (no regridding) | ![][after] |

**Mini-benchmark**

Hereafter is a small script which compares performance with and without regridding. The following parameters were used:

* `n_samples = 1000000`, `n_runs = 1`, `n_reps = 1`: this compares the performance of both plugins with almost no pre-processing overhead. Little difference is observed.
* `n_samples = 0`, `n_runs = 1000`, `n_reps = 10`: this compares the performance of the two pre-processing steps. The regridding strategy performs significantly slower. It should be noted that regridding cost increases with the sharpness of the scattering lobe and the selected example is rather easy compared to what the aerosol phase function data looks like.

<details>
<summary>Click to expand</summary>

```python
import timeit
import numpy as np

import drjit as dr
import mitsuba as mi

mi.set_variant("scalar_mono_double")


def benchmark(func, n_runs=10000, n_reps=10):
    durations = timeit.Timer(func).repeat(repeat=n_reps, number=n_runs)
    results = {
        "durations": durations,
        "mean": np.mean(durations),
        "std": np.std(durations),
    }
    return results


def test_sample(plugin, n_samples):
    if not n_samples:
        return

    ctx = mi.PhaseFunctionContext(None)
    wi = [0, 0, 1]

    mei = dr.zero(mi.MediumInteraction3f, n_samples)
    mei.wi = wi
    mei.sh_frame = mi.Frame3f(mei.wi)
    mei.p = mi.Point3f(0, 0, 0)

    for sample in np.random.random((n_samples, 3)):
        plugin.sample(ctx, mei, sample[0], [sample[1], sample[2]])


def test_irregular(cos_thetas, values, n_samples=10000):
    p = mi.load_dict(
        {
            "type": "tabphase_irregular",
            "values": ",".join(map(str, values)),
            "nodes": ",".join(map(str, cos_thetas)),
        }
    )
    test_sample(p, n_samples)


def test_regular(cos_thetas, values, n_samples=10000):
    steps = cos_thetas[1:] - cos_thetas[:-1]
    cos_thetas_regular = np.linspace(-1.0, 1.0, int(np.ceil(2.0 / steps.min())))
    values_regular = np.interp(cos_thetas_regular, cos_thetas, values)
    p = mi.load_dict(
        {
            "type": "tabphase",
            "values": ",".join(map(str, values_regular)),
        }
    )
    test_sample(p, n_samples)


cos_thetas_irregular = np.array([-1.0, 0.999, 1.0])
values_irregular = np.array([0.5, 1.0, 1.5])
n_samples = 0
n_runs = 1000
n_reps = 10


for title, task in [
    ("Regular", test_regular),
    ("Irregular", test_irregular),
]:
    results = benchmark(
        lambda: task(cos_thetas_irregular, values_irregular, n_samples),
        n_runs=n_runs,
        n_reps=n_reps,
    )
    print(
        f"{title:10s}: {results['mean']:.3g} s ± {results['std']:.3g} ({n_reps} reps)"
    )
```
</details>

# Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,8 @@ when necessary—we also advise to not ignore `DeprecationWarning`s.
 * Transitioned dependency management to `pyproject.toml` following
   [PEP 621](https://peps.python.org/pep-0621/) ({ghpr}`203`).
 * Updated Mitsuba submodule ({ghpr}`228`).
+* Removed the angular regridding feature in `TabulatedPhaseFunction` and 
+  replaced with plugin selection ({ghpr}`229`).
 
 ## v0.22.3 (22 May 2022)
 
--- a/src/eradiate/scenes/phase/_tabulated.py
+++ b/src/eradiate/scenes/phase/_tabulated.py
@@ -61,26 +61,12 @@ class TabulatedPhaseFunction(PhaseFunction):
 
     Notes
     -----
-    The :math:`\mu` coordinate must cover the :math:`[-1, 1]` interval but
-    there is no constraint on value ordering or spacing. In particular,
-    irregular :math:`\mu` grids are supported.
+    * The :math:`\mu` coordinate must cover the :math:`[-1, 1]` interval but
+      there is no constraint on value ordering or spacing. In particular,
+      irregular :math:`\mu` grids are supported.
 
-    Since the underlying ``tabphase`` plugin expects phase function values on
-    a regular :math:`\mu` grid, this class will resample irregularly gridded
-    phase function data on a regular grid. The step of the resampling grid is
-    equal to the smallest :math:`\mu` step found in the input `data` field.
-
-    The resampling grid step selection policy ensures that the precision of
-    the input data is preserved, but the resampled data may be large (several
-    hundred MB) if no care is taken in the preparation of the input data.
-
-    To control the :math:`\mu` grid, you can simply provide a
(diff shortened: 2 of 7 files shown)


leroyvn:
@nollety what do you think?

PR #230, opened 2022-06-01
Change cache directory policy to a less aggressive behaviour

leroyvn:
# Description

This PR changes how cache directory creation is handled. Previous behaviour (cache dir creation upon scene element initialisation) would result in unused directories being created by heterogeneous atmosphere components, even though they would not write files themselves. When running many simulation, a large number of directories would be created and not cleaned up.

The new behaviour defers directory creation to the kernel dict generation step, at the very moment when files are written to disk. A drawback is that we lose a safety check previously preformed during initialisation: a protected cache directory will be detected only upon kernel dict creation. This also introduces a race condition, which requires that the user set the cache directory manually in a multiprocess environment.

# Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,7 +41,11 @@ when necessary—we also advise to not ignore `DeprecationWarning`s.
 * Updated all tests to use `eradiate.run()` instead of the deprecated 
   `Experiment.run()` method ({ghpr}`227`).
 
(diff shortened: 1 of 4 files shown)


leroyvn:
@schunkes does this address your issue?

leroyvn:
Closing in favour of #231.

PR #245, opened 2022-06-29
Define atmo-centimeter unit

nollety:
# Description

Define the atmo-centimeter unit.

Name: `*atmo_centimeter*`

Aliases:
* `"atm_cm"`
* `"centimeter_atmosphere"`
* `"centimeter_amagat"`

From: _Chapter 1 Vertical Structure of an Atmosphere. In International Geophysics,
22:1–45. Elsevier, 1978. https://doi.org/10.1016/S0074-6142(09)60038-3._

This unit is used to express column number density and corresponds to thousand Dobson units.

```python
from eradiate.units import ureg

ureg("milliatm_cm") == ureg("dobson_unit")  # True
ureg("atm_cm") == ureg("kilodu")  # True
```

The name of this unit originates in the fact the atmo-centimeter is defined so that the thickness of a layer of gas at standard temperature (273.15 K) and pressure (100 kPa, equivalent to 1 atm before 1982) with a column number density of 1 atmo-centimeter is 1 cm.
Arguably, the unit could be renamed baro-centimeter or similar...
A typical value for the ozone column number density is 300 Dobson units which is equivalent to 0.3 atmo-centimeter.
In other words, all the ozone in a column of air would form a 3 millimeter thick layer of pure ozone under standard temperature and pressure conditions.

# Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,8 +38,10 @@ when necessary—we also advise to not ignore `DeprecationWarning`s.
 %
 % ### Deprecations and removals
 %
-% ### Improvements and fixes
-%
+### Improvements and fixes
+
+* Added the atmo-centimeter, a.k. atm-cm, to the unit registry ({ghpr}`245`).
+
 % ### Documentation
 %
 % ### Internal changes


leroyvn:
Yes it does, this is a small but new feature.

PR #248, opened 2022-07-11
Fix paths to CKD data sets in `downloads.yml`

nollety:
# Description

Fix the paths to the CKD data sets that are actually used by the `AFGL1986RadProfile` constructor. 

# Checklist

### DIFF
--- a/resources/downloads.yml
+++ b/resources/downloads.yml
@@ -1,7 +1,7 @@
 # This file contains a list of frequently used files which you might want to
 # download.
-- ckd/absorption/1nm/afgl_1986-us_standard-1nm.nc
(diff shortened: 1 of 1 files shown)


nollety:
All tests passed, but as far as I understand, this issue was not detected by failed tests?

nollety:
Not sure if that deserves an entry in the changelog?

leroyvn:
No, this is not a proper bug: in most cases, the missing file will be downloaded silently and things will go through without trouble. This is a problem only when you're getting ready for offline work: in that case, an outdated download list will result in critical files missing from the cache and offline work becomes impossible.

leroyvn:
Thanks @nollety, this is merged.

PR #256, opened 2022-08-18
Substiantial speedup for large biosphere scenes

schunkes:
# Description

Profiling the execution of RAMI scenes, @wint3ria and me found that up to 30% of the time spent in a simulation is due to calls to `pint.Quantity.m_as()`. I investigated and found that the `eradiate.scenes.biosphere._core.InstancedCanopyElement.kernel_instances()` method was treating the instance positions in a very inefficient way.

When creating the dictionary entries for the instances, it would basically call `m_as()` a couple million times like this:

```python
for position in self.instance_positions:
    pos_mag = position.m_as(kernel_length)
```

Changing the code to call `m_as()` only once on the entire array of positions, reduces the time spent there to basically nothing:

```python
for position in self.instance_positions.m_as(kernel_length):
    pos_mag = position
```

# Checklist

### DIFF
--- a/src/eradiate/scenes/biosphere/_core.py
+++ b/src/eradiate/scenes/biosphere/_core.py
@@ -376,11 +376,9 @@ class InstancedCanopyElement(SceneElement):
             f"{self.canopy_element.id}_instance_{i}": {
                 "type": "instance",
                 "group": {"type": "ref", "id": self.canopy_element.id},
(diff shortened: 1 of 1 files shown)


leroyvn (approved):
Hi @schunkes, this is a very welcome fix. I'll merge this now.

PR #274, opened 2022-10-20
Add grid support to `MultipleDistantMeasure` 

leroyvn:
# Description

This PR adds support for gridded viewing angle layouts to `MultipleDistantMeasure`. Changes are as follows:

* `MultipleDistantMeasure` is reworked. The `hplane` and `directions` fields are replaced with a generic `direction_layout` field which holds a `Layout` instance. This field is dict-constructible, as always (I didn't use a Dessine-moi factory for this, it would be overkill IMO).
* `Layout` defines an interface to generate a sequence of viewing directions and angles. This has several advantages:
  * Adding a new implementation is very simple.
  * The layout type can be used for specific operations (this is however no longer needed, see below!).
  * There are currently 5 layouts: 
    * Directions: explicit specification of viewing directions as vectors.
    * Angles: explicit specification of viewing directions as angle pairs.
    * Hemisphere plane cut: as before, negative zenith values work naturally and no longer require fiddling with the output data during post-processing.
    * Azimuthal ring: as before.
    * Grid: directions are generated by making the Cartesian product of zenith and azimuth vectors.
* The `MultipleDistantMeasure.from_viewing_angles()` constructor is deprecated and replaced by specialised constructors, all accessible directly or from the dict API using the `construct` key.
* Post-processing is simpler: the assembly step no longer requires special treatment for the `hplane` layout.
* A helper function is defined to help users grid against (VZA, VAA) the output of a gridded `mdistant` measure. For an example, see the `unstack_gridded_helper.ipynb` file in [that gist](https://gist.github.com/leroyvn/6fac961c93631cbf34c0b0f6f738aef3).

This PR also contains small fixes and changes:

* Azimuth normalisation now also snaps close-to-zero negative values to 0 to avoid irrelevant offsetting of inaccurately positioned values.
* Tests are modified to use the `numpy.allclose()` function the way it is supposed to be, *i.e.* with the first argument as the reference.

# To do

* Code
  * [x] Decide whether to remove or deprecate `from_viewing_angles()`
  * [x] Add gridding helper function (see [this gist](https://gist.github.com/leroyvn/6fac961c93631cbf34c0b0f6f738aef3) for the basic principle)
  * [x] Check if test coverage is sufficient
* Docs
  * [x] Add docstrings
  * [x] Add `Layout` API docs
  * [x] Update tutorials
  * [x] ~~Add measure guide explaining how all this works (other PR maybe?)~~
  * [x] ~~Add measure tutorial (other PR maybe?)~~
  * [x] Add changelog entry

# Checklist

### DIFF
--- a/tests/02_eradiate/01_unit/scenes/measure/test_multi_distant.py
+++ b/tests/02_eradiate/01_unit/scenes/measure/test_multi_distant.py
@@ -4,7 +4,159 @@ import pytest
 from eradiate import unit_registry as ureg
 from eradiate.contexts import KernelDictContext
 from eradiate.frame import AzimuthConvention
-from eradiate.scenes.measure._multi_distant import MultiDistantMeasure
+from eradiate.scenes.measure._multi_distant import (
+    AngleLayout,
+    AzimuthRingLayout,
+    DirectionLayout,
+    GridLayout,
+    HemispherePlaneLayout,
+    MultiDistantMeasure,
+)
+
+# ------------------------------------------------------------------------------
+#                               Layout framework
+# ------------------------------------------------------------------------------
+
+
+def test_angle_layout(mode_mono):
+    """
+    Unit tests for the AngleLayout class
+    """
+    # Constructing without argument fails
+    with pytest.raises(TypeError):
+        AngleLayout()
+
+    # (2,) arrays are reshaped as (1, 2)
+    assert AngleLayout([0, 0] * ureg.deg).angles.shape == (1, 2)
+
+    # Regular construction pattern succeeds
+    layout = AngleLayout([[0, 0], [45, 0], [45, 90], [45, 180]] * ureg.deg)
+
+    # Directions are correctly computed and point outwards
+    assert np.allclose(
+        layout.directions,
+        [
+            [0, 0, 1],
+            [np.sqrt(2) / 2, 0, np.sqrt(2) / 2],
+            [0, np.sqrt(2) / 2, np.sqrt(2) / 2],
+            [-np.sqrt(2) / 2, 0, np.sqrt(2) / 2],
+        ],
+    )
+
+
+@pytest.mark.parametrize(
+    "convention, expected",
+    [
+        ("east_right", [[1, 0, 0], [0, 1, 0]]),
+        ("north_left", [[0, 1, 0], [1, 0, 0]]),
+    ],
+)
+def test_layout_azimuth_convention(mode_mono, convention, expected):
+    # Azimuth conventions are properly applied
+    layout = AngleLayout(
+        angles=[[90, 0], [90, 90]] * ureg.deg, azimuth_convention=convention
+    )
+    assert np.allclose(
+        expected, layout.directions
+    ), f"\nexpected {expected}\ngot      {layout.directions}"
+
+
+def test_direction_layout(mode_mono):
+    """
+    Unit tests for the DirectionLayout class
+    """
+    # Constructing without argument fails
+    with pytest.raises(TypeError):
+        print(DirectionLayout())
+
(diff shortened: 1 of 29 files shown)


tests/02_eradiate/01_unit/radprops/test_rayleigh_scattering.py
```diff
@@ -22,7 +22,7 @@ def test_sigma_s_air():
     # Compare to reference value computed from scattering cross section in
     # Bates (1984) Planetary and Space Science, Volume 32, No. 6.
     assert np.allclose(
-        compute_sigma_s_air(number_density=_LOSCHMIDT), expected, rtol=1e-2
+        expected, compute_sigma_s_air(number_density=_LOSCHMIDT), rtol=1e-2
```
schunkes, line 25: what  is going on wit the `allclose` statements? You mentioned a regression in DrJit, is this a mitigation for that?

leroyvn: [`allclose()` is asymmetric](https://numpy.org/doc/stable/reference/generated/numpy.allclose.html): one of the two arguments is the reference. Now, looking again at the formula, I'm realising that it's `b`: contrary to what I changed here, the initial formula was correct. There are howver other parts of the code where things are incorrect, I'll fix that (and revert the inappropriate changes I made).

This is unrelated with Dr.Jit or Mitsuba.

leroyvn: Done. Turns out the tests were a mess with that respect. This is fixed now (wherever the concept of a reference can be relevance, of course).

tests/02_eradiate/01_unit/test_frame.py
```diff
@@ -82,7 +85,7 @@ def test_transform_azimuth_from_east_right(to_convention, initial, normalize):
         to_convention=to_convention,
         normalize=normalize,
     )
-    assert np.allclose(result, expected)
+    assert np.allclose(result, expected), np.rad2deg(result)
```
schunkes, line 88: What is happening here?

tests/02_eradiate/01_unit/scenes/measure/test_multi_distant.py
```diff
@@ -66,47 +242,57 @@ def test_multi_distant_measure_from_viewing_angles(mode_mono):
     angles = np.reshape(np.stack((zeniths, azimuths), axis=-1), (-1, 1, 2)) * ureg.deg
 
     measure = MultiDistantMeasure.from_viewing_angles(zeniths, azimuths)
-    assert np.allclose(angles, measure.viewing_angles)
+    print(measure.direction_layout)
```
schunkes, line 245: should this still be here?

tests/02_eradiate/01_unit/test_frame.py
leroyvn: It's just a thing I do to make debugging simpler: the second argument of the assert statement gets printed upon failure, and it's sometimes much more comprehensive than the traceback. If the test succeeds, you see nothing. In that case, I just output the result vector converted to degrees to see better what was going on. I'll clean that up.

tests/02_eradiate/01_unit/scenes/measure/test_multi_distant.py
leroyvn: Nope

leroyvn:
I think this is ready to ship. I decided to postpone the tutorials, it's a lot of work and I have other priorities. The usage of the new construction API is however almost the same as the previous (the new `Layout` abstraction layer is entirely hidden from the user and it's not even part of the public API). The gist I added to the description demonstrates how to use the xarray post-processing helper.

PR #285, opened 2022-11-22
Experiment: Swap the spectral and measure loops

leroyvn:
# Description

This PR swaps the spectral and sensor loops.

The measure loop used is currently external to the spectral loop. This results in unnecessary kernel scene rebuilds. The new behaviour consists in computing radiative transfer for all selected sensors at each spectral loop iteration. This reduces the number of scene rebuilds for multi-measure experiments.

**Breaking change:** The spectral configuration of the first selected measure is used. If the user wants to process multiple measures with different spectral configurations, they must make one run per measure. This issue will be addressed in future updates.

# Checklist

### DIFF
--- a/src/eradiate/experiments/_core.py
+++ b/src/eradiate/experiments/_core.py
@@ -17,7 +17,7 @@ from .. import pipelines
 from .._config import ProgressLevel, config
 from .._mode import ModeFlags, supported_mode
 from ..attrs import documented, parse_docs
-from ..contexts import KernelDictContext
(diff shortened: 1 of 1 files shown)


wint3ria:
Should I use this patch to compute the canopies of the RAMI4ATM benchmark?

leroyvn:
It will speed up workloads using a multi-measure setup; however, this also breaks the `run()` API, so I'd advise to use it only if it's really useful since you'll have to refactor your calling code.

leroyvn:
Closing (superseded by #296).

PR #296, opened 2023-01-08
Scene update engine

leroyvn:
# Description

This PR refactors how kernel dictionaries are generated and re-engineers the spectral loop implementation to take advantage of Mitsuba's ability to retain a `Scene` object and update its parameters.

# Motivation

Eradiate currently rebuilds an entire Mitsuba scene at each iteration of the spectral loop. While this makes kernel dictionary generation simple, it may add significant unnecessary overhead to each iteration of the loop, *e.g.* by reconstructing all objects, in particular shapes. Instead, this PR transitions the spectral loop implementation to a system taking advantage of Mitsuba's ability to update selected parameters in an existing scene without rebuilding it. This requires a complete rewrite of the infrastructure used to generate scene dictionaries, as well as the addition of a mechanism to issue scene parameter updates during the spectral loop. This is the reason why this work was initiated.

# Summary of changes

Most of the changes in this PR implement the transition to a retained scene-based spectral loop. The rest is overdue maintenance which I finally took care of in the process.

- **Deferred kernel parameter evaluation.** This PR adds a standard interface to declare a parameter for use in a kernel scene dictionary or to be updated in the kernel scene after instantiation. Scene dictionary parameters are declared using a new `InitParameter` type; values use for an update are declared using the `UpdateParameter` type. Both are callables, but `UpdateParameter` also provides flags for filtering, as well as a means to look up its associated Mitsuba scene parameter once the scene is instantiated.
- **Kernel scene dictionary and update map templates.** Scene dictionaries are now declared as templates mixing pre-evaluated and deferred values, using a new `KernelDictTemplate` type. Update maps follow a similar approach with the `UpdateMapTemplate` type. Both must be rendered prior to usage with Mitsuba's `mi.load_dict()` or `SceneParameters.update()`. Template rendering requires an appropriately set `KernelDictContext` instance.
- **Updates to the `SceneElement` type hierarchy.** The new system renders the `kernel_dict()` methods obsolete. They are replaced by a set of properties (`template` and `params`) which respectively generate a Mitsuba scene dictionary template and a parameter update map template corresponding the considered scene element.
- **Object tree traversal system.** Scene dictionary and parameter update map templates are built through a new tree traversal system inspired from the one implemented by Mitsuba, with some extra logic. The current implementation is imperfect and could very likely benefit from upgrades or refactoring actions.
- **Rewrite of the spectral loop.** The spectral loop is entirely refactored and now utilises Mitsuba's update mechanism to perform minimal scene updates and reduce the amount of unnecessary computation. Scene updates are driven by kernel dictionary contexts, now formally emitted by the `Experiment` itself.
- **Batch computation of all sensors.** All sensors are rendered without a scene update during a spectral loop iteration.
- **Updates to the `Experiment` type hierarchy.** These classes are rewritten to implement the new update-based workflow. This resulted in several interface updates, some required, others for convenience. `Experiment`s now create an intermediate `Scene` object, from which a kernel dictionary template and parameter map are extracted. This intermediate scene can be adjusted depending on experiment configuration. It can also be used to inject arbitrary scene elements into the kernel dictionary.
- **Deep refactoring actions.** The following classes underwent extra refactoring, in addition to transitioning to the new kernel dictionary generation system:
  - `Atmosphere`: A common interface is defined for all heterogeneous atmosphere classes. In addition, all `HeterogeneousAtmosphere` subclasses can be assigned a forced vertical kernel-level grid.
  - `BlendedPhaseFunction`: Transition from a recursive to a loop-based algorithm.
  - `RadProfile`: Evaluation on arbitrary altitude grids is now permitted.
- **Bug fixes.** I took the opportunity to fix some pending issues when a fix was cheaper than a refactoring with a workaround.
  - Broken / discontinuous radiative property fields: the atmospheric component mixing policy was redesigned. The `HeterogeneousAtmosphere` container now evaluates all its subcomponents on its own grid instead of interpolating the components evaluated on their own altitude grid. This should result in smoother radiative property fields.
  - A bug where the texture transformation of the `BlendedPhaseFunction` class in spherical shell geometry was incorrect is fixed.
- **Dropped features.** I dropped some unused features to keep the code concise and optimise performance.
  - `Measure`: Sample count split is retired. I consequently removed the `AggregateSampleCount` pipeline step.
  - `RadProfile`: The `ArrayRadProfile` class is retired. Its use case range was too limited to justify a refactoring at this stage. We may bring it back later.
- **Radiometric kernel update.** This new version uses Mitsuba v3.2.1.

# Details

# To do

- [x] mitsuba-renderer/mitsuba3#480
- [x] Update Mitsuba ref to benefit from aforementioned bugfix (PR#480)
- [x] Update `biosphere` subpackage content
- [x] Update `Experiment` type hierarchy
- [x] Fix remaining bugs
- [ ] Update regression test references after a double check
- [x] Pass over API documentation
- [ ] Pass over user manual
- [ ] Pass over developer documentation

# Checklist


leroyvn:
We've been delaying the merge of this behemoth for too long, let's move on. Docs work can be done afterwards. At least everyone can now branch off from something that works.

PR #298, opened 2023-01-11
Refactor the `Mode` type

leroyvn:
# Description

This PR refactors the `Mode` class, simplifies the implementation and adds additional degrees of freedom in Mitsuba variant selection. Changes are as follows:

* The flag system was not really used and is replaced by multiple fields which use either boolean or flag values.
* The new `Mode` data class holds one field per degree of freedom in mode definition (5 at the moment).
* The `Mode.has_flags()` method is replaced by `Mode.check()`, which takes one argument per DoF field. Flag field values can be specified using enum values or strings.
* Minor code updates were done to upgrade old usage patterns.

# Checklist

### DIFF
--- a/src/eradiate/_mode.py
+++ b/src/eradiate/_mode.py
@@ -14,52 +14,51 @@ from .exceptions import UnsupportedModeError
 # ------------------------------------------------------------------------------
 
 
-class ModeFlags(enum.Flag):
+class Flag(enum.Flag):
     """
-    Flags defining Eradiate mode features.
+    Small extension to :class:`enum.Flag` that adds a ``convert()`` class method
+    constructor.
     """
 
-    # -- Mode feature flags ----------------------------------------------------
+    @classmethod
+    def convert(cls, value: t.Any) -> Flag:
+        """
+        Try to convert a value to a flag. Strings are capitalised and converted
+        to the corresponding enum member.
+        """
+        if isinstance(value, str):
+            return cls[value.upper()]
+        else:
+            return cls(value)
 
-    # Eradiate
-    ERT_MONO = enum.auto()  #: Mode performs monochromatic simulations
-    ERT_CKD = enum.auto()  #: Mode performs correlated-k spectral simulation
-    # ERT_RGB = enum.auto()  #: Mode performs renders RGB images
 
-    # Mitsuba
-    MI_SCALAR = enum.auto()  #: Mode maps to a scalar Mitsuba variant
-    # MI_LLVM = enum.auto()  #: Mode maps to a LLVM Mitsuba variant
-    # MI_AD = enum.auto()  #: Mode maps to an autodiff Mitsuba variant
-    MI_MONO = enum.auto()  #: Mode maps to a monochromatic Mitsuba variant
-    # MI_RGB = enum.auto()  #: Mode maps to an RGB Mitsuba variant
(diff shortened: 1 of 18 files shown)


src/eradiate/_mode.py
```diff
@@ -14,52 +14,51 @@
...
 
-class ModeFlags(enum.Flag):
+class Flag(enum.Flag):
     """
-    Flags defining Eradiate mode features.
+    Small extension to :class:`enum.Flag` that adds a ``convert()`` class method
+    constructor.
     """
 
-    # -- Mode feature flags ----------------------------------------------------
```
wint3ria, line 22: I like how the new flag classes removes the bitwise logic, its easier to read. It looks easier to maintain too

src/eradiate/_mode.py
```diff
@@ -68,33 +67,45 @@ class ModeFlags(enum.Flag):
 
 # Map associating a mode ID string to the corresponding class
 # (aliased in public API section)
-_mode_registry = {
```
wint3ria, line 71: From what I understand, the end user API does not change thanks to this mapping am I right?

src/eradiate/scenes/atmosphere/_particle_layer.py
```diff
@@ -441,93 +442,91 @@ def eval_radprops(
...
+                            long_name="scattering coefficient",
+                            units=symbol(sigma_s.units),
                         ),
                     ),
-                },
-            ).isel(w=0)
+                }
+            )
 
-        else:
```
wint3ria, line 529: If I understand correctly, incompatible configurations are detected when the eradiate.run function is called? If I'm correct, would it be feasible to have the classes constructor to raise? Detecting this kind of issue as soon as possible seems to be a good practice IMO, but if it may be complicated as I don't think it is foreseen at all. In any case implementing this would be the scope of another Issue/PR

wint3ria (approved):
SGTM. I only have a few questions for open discussion, no real concern or request

src/eradiate/scenes/atmosphere/_particle_layer.py
wint3ria, line 529: If I understand correctly, incompatible configurations are detected when the eradiate.run function is called? If I'm correct, would it be feasible to have the classes constructor to raise? Detecting this kind of issue as soon as possible seems to be a good practice IMO, but if it may be complicated as I don't think it is foreseen at all. In any case implementing this would be the scope of another Issue/PR

nollety (approved):

schunkes (approved):

src/eradiate/scenes/atmosphere/_particle_layer.py
leroyvn: Things are so because I originally to make scene element instantiation mode independent. In other words, you should be able to select the mode after creating the experiment. This, in practice, doesn't really work and the pattern is more to select a mode right after import; with that in mind, transferring lode checks to constructors makes sense.

src/eradiate/_mode.py
leroyvn: Indeed

src/eradiate/_mode.py
leroyvn: Yes, mode definitions are clearer and more streamlined. This refactoring was long overdue!

PR #302, opened 2023-01-31
Spot illumination

schunkes:
# Description

The SpotIllumination is rougly equivalent to a perspective camera. It is parametrized through a location, a target and a beam width.

# Checklist

### DIFF
--- a/src/eradiate/scenes/illumination/_spot.py
+++ b/src/eradiate/scenes/illumination/_spot.py
@@ -1,0 +1,233 @@
+from __future__ import annotations
+
+import typing as t
+import warnings
+from pathlib import Path
+
+import attrs
+import drjit as dr
+import mitsuba as mi
+import numpy as np
+import pint
+import pinttr
+
+from ._core import Illumination
+from ..core import NodeSceneElement
+from ..spectra import Spectrum, spectrum_factory
+from ... import validators
+from ...attrs import documented, parse_docs
+from ...units import unit_context_config as ucc
+from ...units import unit_context_kernel as uck
+from ...units import unit_registry as ureg
+from ...validators import has_quantity
+
+
+@parse_docs
+@attrs.define(eq=False, slots=False)
+class SpotIllumination(Illumination):
+    """
+    Spot illumination scene element [``spot``].
+
+    Eradiate ships a beam texture that implements a gaussian beam profile with three standard deviations
+    included in the total beam width. This texture is named `gaussian_3sigma.bmp` and can be used by
+    retrieving the path to the file using `eradiate.data.data_store.fetch()`.
+    """
+
+    origin: pint.Quantity = documented(
+        pinttr.field(
+            factory=lambda: [1, 1, 1] * ureg.m,
+            validator=[validators.has_len(3), pinttr.validators.has_compatible_units],
+            units=ucc.deferred("length"),
+        ),
+        doc="A 3-vector specifying the position of the spot.\n"
+        "\n"
+        "Unit-enabled field (default: ucc['length']).",
+        type="quantity",
+        init_type="array-like",
+        default="[1, 1, 1] m",
+    )
+
+    target: pint.Quantity = documented(
+        pinttr.field(
+            factory=lambda: [0, 0, 0] * ureg.m,
+            validator=[validators.has_len(3), pinttr.validators.has_compatible_units],
+            units=ucc.deferred("length"),
+        ),
+        doc="Point location targeted by the spot.\n"
+        "\n"
+        "Unit-enabled field (default: ucc['length']).",
+        type="quantity",
+        init_type="array-like",
+        default="[0, 0, 0] m",
+    )
+
+    @target.validator
+    @origin.validator
+    def _target_origin_validator(self, attribute, value):
+        if np.allclose(self.target, self.origin):
+            raise ValueError(
+                f"While initializing {attribute}: "
+                f"Origin and target must not be equal, "
+                f"got target = {self.target}, origin = {self.origin}"
+            )
+
+    up: np.ndarray = documented(
+        attrs.field(
+            converter=np.array,
+            validator=validators.has_len(3),
+        ),
+        doc="A 3-vector specifying the up direction of the spot.\n"
+        "This vector must be different from the spots's pointing direction,\n"
+        "which is given by ``target - origin``.",
+        type="array",
+        init_type="array-like",
+        default="[0, 0, 1]",
+    )
+
+    @up.default
+    def _up_factory(self):
+        direction = dr.normalize(
+            mi.ScalarVector3f(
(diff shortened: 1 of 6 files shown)


src/eradiate/scenes/illumination/_spot.py
```diff
@@ -0,0 +1,212 @@
...
+                "beam_width": self.beam_width.m_as(uck.get("angle")),
+                "cutoff_angle": self.beam_width.m_as(uck.get("angle")),
+                "intensity": self.intensity.kernel_dict(ctx=ctx)["spectrum"],
+            }
+        }
+
+        if self.beam_profile is not None:
+            retdict[self.id]["texture"] = {
+                "type": "bitmap",
+                "filename": self.beam_profile,
```
nollety, line 208:
```suggestion
                "filename": str(self.beam_profile),
```
otherwise we will get a `pathlib.PosixPath` in the dictionary, which mistuba does not like iirc.

src/eradiate/scenes/illumination/_spot.py
```diff
@@ -0,0 +1,212 @@
...
+        "Must be an intensity spectrum (in W/sr/nm or compatible unit). "
+        "Can be initialised with a dictionary processed by "
+        ":meth:`.SpectrumFactory.convert`.",
+        type=":class:`~eradiate.scenes.spectra.Spectrum`",
+        init_type=":class:`~eradiate.scenes.spectra.Spectrum` or dict or float",
+        default="1.0 ucc[intensity]",
+    )
+
+    beam_profile: Path = documented(
+        attrs.field(converter=Path),
```
nollety, line 121: It seems that this attribute is not required, but it has no default :thinking:

tests/02_eradiate/01_unit/scenes/illumination/test_spot.py
```diff
@@ -0,0 +1,49 @@
...
+        target=[0, 0, 0],
+        direction=[0, 0, -1],
+        spot_radius=spot_radius,
+        beam_divergence=beam_angle
+    )
+
+    assert np.allclose(d.origin.m_as(ureg.m), [0, 0, expected_distance])
+
+    assert d.kernel_dict(ctx).load() is not None
+
```
nollety, line 49: Perhaps, we could add a test where the `beam_profile` parameter is set. Since it is a file path, I guess things can easily go wrong there, so it'd be nice to confirm that this is working.

nollety (approved):
Looking forward to use a spot illumination in my scenes!

src/eradiate/scenes/illumination/_spot.py
```diff
@@ -0,0 +1,212 @@
...
+from ...validators import has_quantity
+
+
+@parse_docs
+@attrs.define
+class SpotIllumination(Illumination):
+    """
+    Spot illumination scene element [``spot``].
+
+    This light source is roughly equivalent to the :class:`.PerspectiveCamera` in that it
```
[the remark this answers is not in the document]
schunkes: That sounds nice :) I kinda copied the docstring from mitsuba :D

tests/02_eradiate/01_unit/scenes/illumination/test_spot.py
schunkes: Good point, but then I need a fixture that creates a temporary bitmap file :D

schunkes:
I took care of your feedback and added a testcase that creates a temporary bitmap file. I am just a bit worried that Vincent will force us to remove it. I am not sure why, but it smells like something he would not want.

src/eradiate/scenes/illumination/_spot.py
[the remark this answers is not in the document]
leroyvn: Gentlemen, I'm afraid neither is correct: the adjoint to the `perspective` sensor is the `projector` emitter, not the `spot`. I'd rather remove this part and reference the Mitsuba plugin docs for users interested in this illumination.

leroyvn (changes requested):
Hi @schunkes, finally found some time to have a look. In addition to my comments, I would add a more general one.

You branched off `main` but we won't merge to it—at least not in its current state. Instead, we'll merge into what is currently on `traverse`. No worries, though: it's only about porting some parts of the code to the new kernel interface (this will require to update some of your tests as well).

src/eradiate/scenes/illumination/_spot.py
```diff
@@ -0,0 +1,217 @@
...
+    beam_profile: t.Optional[Path] = documented(
+        attrs.field(
+            default=None,
+            converter=attrs.converters.optional(Path)
+        ),
+        type="Path",
+        init_type="path-like",
+        default=None,
+        doc="Path to the beam profile file, relative to the current working directory. "
+            "Must be a valid bitmap image file.",
```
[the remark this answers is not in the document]
schunkes: I don't like that idea. We'd either have to make arbitrary choices or have a dozen profiles in the data store.
- Gaussian 1-sigma
- Gaussian 2-sigma
- Gaussian 3-sigma
- Gaussian 0.5-sigma
...

And the first user to use it will of course want something we don't offer... I think the default should just be untextured.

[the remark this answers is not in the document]
leroyvn: * It's always nice to ship a bit of data to demonstrate features; with that in mind, the data store becomes the natural entry point.
* I'd be in favour of shipping a single default profile as an example.
* Having built-in profiles doesn't mean constraining users. We ship SRFs, but users who want to provide their own can as well and they have examples and documentation to prepare their data.
* I agree with leaving the default untextured.

schunkes:
I rebased this on traverse and made the corresponding changes.

leroyvn:
Hi @schunkes, you can rebase on `main` now, the `traverse` branch is merged.

schunkes:
> Hi @schunkes, you can rebase on `main` now, the `traverse` branch is merged.

Done :)

leroyvn:
Looks like there still are merge conflicts.

src/eradiate/scenes/illumination/_spot.py
```diff
@@ -0,0 +1,227 @@
...
+        target = pinttr.util.ensure_units(target, default_units=ucc.get("length"))
+        half_angle = (
+            pinttr.util.ensure_units(beam_width, default_units=ureg.radian) / 2.0
+        )
+        tan_divergence = np.tan(half_angle)
+        distance = spot_radius / tan_divergence
+
+        origin = target - (direction * distance)
+
+        _, up = mi.coordinate_system(direction)
```
leroyvn, line 199: See previous comment about the default `up` value: this should be moved to the constructor.

leroyvn (changes requested):
Hi @schunkes, this looks almost good to me. Two small things to change and we'll be good to go.

PR #311, opened 2023-03-29
Spectral (disptach) refactoring

nollety:
# Description

Resolve Clickup (private) issue: [https://app.clickup.com/t/388a4r4](https://app.clickup.com/t/388a4r4).

## Motivation

Eradiate supports two spectral modes —_monochromatic_ and _ckd_— at the moment and each method evaluating a spectral quantity must be implemented twice, once for each spectral mode. At runtime, depending on the active mode, the evaluation is dispatched to the relevant implementation, i.e. either the _monochromatic_ or _ckd_ implementation. The current implementation of this dispatch does not scale well with the number of spectral modes and can be improved with [singledispatchmethod](https://docs.python.org/3/library/functools.html#functools.singledispatchmethod) from the functools library.

We note that a multiple dispatch can be implemented with the [plum](https://github.com/beartype/plum) library (on PyPI but not on conda) and we keep this possibility in mind for future needs.

The `SpectralContext`—specifically its `CKDSpectralContext` subclass— data structure is not adequate to easily lookup/interpolate data in the new CKD absorption datasets whose integration is scheduled for the next milestone (see below for further details).

### `SpectralIndex`

This pull request suggests to replace the `SpectralContext` data structure by the equivalent `SpectralIndex` data structure (new :sparkles:), with as many subclasses as there are spectral modes in Eradiate. At the moment, Eradiate has two spectral modes, as indicated in the table below.

| spectral mode | corresponding spectral index subclass |
| :---: | :---: |
| mono | MonoSpectralIndex |
| ckd | CKDSpectralIndex |

The main difference between `SpectralContext` and `SpectralIndex` is in the subclass corresponding to the ckd spectral mode. `CKDSpectralContext` was defined by `bindex` and `bin_set` attributes where as `CKDSpectralIndex` is defined by a `w` (the bin center wavelength) and `g` (the quadrature _g_-point). This difference anticipates the integration of the new CKD absorption datasets which tabulates the absorption coefficient as a function of the cumulative distribution (_g_).

### Spectral dispatch

This pull request suggests that all `eval_*` methods are decorated with `singledispatchmethod` and take a `si` (**s**pectral **i**ndex) as first argument (:warning: the dispatch occurs only on the first argument). Based on the active spectral mode, these `eval_*` methods are called with a subclass of `SpectralIndex` that matches the spectral mode, as illustrated below for `UniformSpectrum`:

```python
from functools import singledispatchmethod

class UniformSpectrum:

  @singledispatchmethod  # must be the outer most decorator
  def eval(self, si: SpectralIndex):
    raise NotImplementedError

  @eval.register(MonoSpectralIndex)
  def _(self, si):
    return self.eval_mono(w=si.w)

  @eval.register(CKDSpectralIndex)
  def _(self, si):
    return self.eval_ckd(w=si.w, g=si.g)

  def eval_mono(self, w: pint.Quantity): 
    ...

  def eval_ckd(self, w: pint.Quantity, g: float):
    ...
```

We keep the actual methods implementation in separate methods —in this example `eval_mono` and `eval_ckd`— so that these methods can be accessed from the test/debug code.

Note that the class needs only to implement and register the evaluation specializations that it wants. For example, a class may not support evaluation in ckd mode. In that case, it would not define the corresponding method implementation:

```python
class UniformSpectrum:

  @singledispatchmethod  # must be the outer most decorator
  def eval(self, si: SpectralIndex):
    raise NotImplementedError

  @eval.register(MonoSpectralIndex)
  def _(self, si):
    return self.eval_mono(w=si.w)

  def eval_mono(self, w: pint.Quantity): 
    ...
```

If a `CKDSpectralIndex` is passed to the class's eval method, the latter raises a `NotImplementedError` as one would expects since the evaluation method is not implemented for the ckd spectral mode. 

#### Note on evaluation in ckd mode

In ckd mode, all spectral quantities except the molecular atmosphere absorption coefficient are evaluated at the bins' center wavelengths. Depending on the spectral variations of the said quantities (e.g. solar irradiance) and the bins' widths, the evaluated value for a bin may not represent the average value of the quantity in that bin. This is not an issue with the `eval_*` method. This simply means that the assumption of the CKD method that all spectra other than the absorption coefficient' are uniform over the bins' extents is not valid. There are at least two ways to go around that:
1. work with narrower bins, hopefully the spectral variations at that scale become negligible (if you must go so far that the bins are unreasonably narrow, it means that the CKD method is probably not an appropriate method in this case). This will probably lengthen the simulation time.
2. if you can, modify the spectral data so that the spectral variations are gone/reduced, e.g. rebin the solar irradiance spectrum by computing the average value over each CKD bin. This will probably add a source of uncertainty to your simulation.

### `Measure`

Some changes were also made on the `Measure` side, some of these changes addressing the following (private) issue: https://app.clickup.com/t/2trzg40.

The `MeasureSpectralConfig` family has been removed. The `Measure` attribute `spectral_cfg: MeasureSpectralConfig` has been replaced by a `srf: Spectrum` attribute (new :sparkles:) that holds the spectral response function (SRF) of the instrument. Two types of `Spectrum` are supported:
* `InterpolatedSpectrum` lets one define a spectral response function from data points (wavelength and response values). This use case is the typical high-level use case, e.g. satellite simulation.
* `MultiDeltaSpectrum` (new :sparkles:) lets one define a SRF made up of multiple delta-like response function. These delta-like response functions serve the purpose of selectionning a specific spectral point, i.e. a single wavelength in monochromatic mode, and a single bin in CKD mode. This use case is typically a test/debug use case.

### `Experiment`

This pull request also gives new duties to experiments. Given that the latter are the only instances with easy access to the combined information on:
* the spectral response function of each measure,
* the molecular atmosphere's absorption dataset, if present

they are in the best position to generate the set of spectral indices at which to run the experiment, which what does the new `Experiment.spectral_indices()`.

The behaviour of `spectral_indices()` relies on the `srf` attribute of each experiments' measures and on the `Experiment`'s `_spectral_set` attribute (new :sparkles:), exposed as a `spectral_set` property.

The rationale is the following:
* when an experiment is instanciated, its `default_spectral_set` parameter is set by the user or to a default value, `_spectral_set` is unset.
* in a post-init step, `_spectral_set` is set with the value returned by `Experiment.get_spectral_set`: the default implementation of these getters simply takes into account the SRF information to generate the set of wavelengths (resp., the set of bins). This is why  experiments that accept atmosphere scene elements are allowed to override these getters.

Finally, the wavelength set/bin set is able to generate spectral indices, and this is what happens when `Experiment.spectral_indices()` is called. 

Different pieces of information come in different order of priority when setting the set of wavelengths (resp., the set of bins), depending on whether an absorbing molecular atmosphere is present:

* if a **molecular** atmosphere is present and is is absorbing:
    1. the absorption dataset
    2. the default wavelength/bin set filtered by the SRF
* else:
    1. the default wavelength/bin set filtered by the SRF

Given that the user has control (easy) on the default wavelength/set, control (fairly easy) on the SRF, and control (relatively difficult) on the absorption dataset, it is possible for them with knowledge of the above priorities to directly determine what wavelength/bin set is going to be generated by the experiment.   

#### Wavelength/bin set filtering with SRF

The process of filtering wavelengths and bins with the SRF is handled by `Spectrum.select_in`.
Two kinds of spectra are supported to represent SRFs and the filtering process adapts to the spectrum kind:

* mode: `mono`
  * SRF kind: `MultiDeltaSpectrum`: selected wavelengths are the SRF wavelengths
  * SRF kind: `InterpolatedSpectrum`: selected wavelengths are all wavelengths from the wavelength set that fall within the wavelength interval where the SRF evaluates to a non-zero value.
* mode: `ckd`
  * SRF kind: `MultiDeltaSpectrum`: selected bins are the bins that include the SRF wavelengths
  * SRF kind: `InterpolatedSpectrum`: selected bins are all bins that overlaps with the wavelength interval where the SRF evaluates to a non-zero value.

The whole process of identifying the wavelength interval where a SRF evaluates to a non-zero value, as well as what wavelength values are included in what interval and what bins overlaps with what intervals was greatly simplified thanks to the [portion](https://anaconda.org/conda-forge/portion) library by Alexandre Decan.

### Pipelines

The `AggregateCKDQuad` pipeline step was modified to take a binset parameter. This lets the experiment pass the information about the quadrature rules to the pipeline step so that it can compute the quadrature.
Other pipelines underwent minor changes.

## Notes

* The present PR would serve as general documentation until dedicated pages are added to our docs (future work)
* A few improvements originally planned to be part of this PR have been identified as not crucial and, given the schedule, have been postponed —see this Clickup list (private): https://app.clickup.com/t/861mn6rwy.
* A few `# TODO: PR#311 hack` comments have been placed to mark hacky lines that are to be refactored when working on integrating the new CKD absorption datasets.
* ~~For some unknown reason, the test `tests/02_eradiate/02_system/test_basic.py` fails: this is under investigation~~

# Checklist


nollety:
@leroyvn I cannot seem to have `tests/02_eradiate/02_system/test_basic.py` to pass. It fails with for the `mono_single-1.0-constant-500000.0` case where the relative difference reach 0.11% whereas the test tolerance is set at 0.1%. I could not get the test to pass by changing the `spp` value (up to 1e7). If this is related to the seed state (as suggested by the fact that the test passes when run alone), how can I reset it? This test does not `eradiate.run(exp)` but call directly `mi_render` so I am not sure how to do that. I tried to add `ert_seed_state.reset()` before the call to `mi_render` but that did not have any effect.

leroyvn (commented):
Hi @nollety, this is a first round of review for this WIP. I checked the code, not the tests. Although I think you can go further with single dispatch and remove more redundant code / boilerplate, this looks very promising. We can address the most complicated points live.

When you update this branch, please don't squash commits immediately—do it only after we'll be done with reviews. Successive reviews will be easier on my end.

src/eradiate/contexts.py
```diff
@@ -288,17 +50,17 @@ class KernelDictContext(Context):
     instance.
     """
 
-    spectral_ctx: SpectralContext = documented(
+    si: SpectralIndex = documented(
```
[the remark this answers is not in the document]
nollety: When the ambiguity comes up, we can rename `si: SpectralIndex` to something else, of course

src/eradiate/experiments/_core.py
```diff
@@ -119,6 +119,77 @@ def results(self) -> dict[str, xr.Dataset]:
...
 
+    default_wset: WavelengthSet = documented(
+        attrs.field(
+            default=WavelengthSet.arange(280 * ureg.nm, 2401 * ureg.nm, 1 * ureg.nm),
+            validator=attrs.validators.instance_of(WavelengthSet),
+        ),
+        doc="Default wavelength set used by the experiment."
+        "This attribute is used in mono mode only."
+        "This attribute is not used in CKD mode.",
+        type="WavelengthSet.arange(280 * ureg.nm, 2401 * ureg.nm, 1 * ureg.nm)",
```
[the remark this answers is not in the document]
nollety: Good catch, thanks!

About the constants, I like the idea; I added a specific commit that defines these constants

src/eradiate/experiments/_core.py
```diff
@@ -119,6 +119,77 @@ def results(self) -> dict[str, xr.Dataset]:
         """
         return self._results
 
+    default_wset: WavelengthSet = documented(
```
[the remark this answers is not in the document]
nollety: I had not thought of that, and I find it a very good idea, thanks I'll implement it

src/eradiate/pipelines/_aggregate.py
```diff
@@ -100,20 +87,50 @@ def transform(self, x: t.Any) -> t.Any:
...
+            ng = _bins[_bin_index].quad.nodes.size  # TODO: generalize
+
             values_at_nodes = img.sel(bin=bin).values
 
             # Rationale: Avoid using xarray's indexing in this loop for
             # performance reasons (wrong data indexing method will result in
             # 10x+ speed reduction)
             for indexes in itertools.product(*[list(range(n)) for n in dims.values()]):
-                aggregated.values[(i_bin, *indexes)] = quad.integrate(
+                aggregated.values[(i_bin, *indexes)] = Quad.gauss_legendre(
```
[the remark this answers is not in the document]
nollety: No need to instanciate a new one, that code bit is inefficient - I fixed it, thanks

src/eradiate/scenes/measure/_core.py
```diff
@@ -556,17 +169,26 @@ class Measure(NodeSceneElement, ABC):
...
+        "``'sentinel_2a-msi-4'`` it tries to serve the corresponding dataset "
+        "from the Eradiate data store. By default, the *prepared* version of "
+        "the SRF is served unless it does not exist in which case the *raw* "
+        "version is served. To request that the raw version is served, append "
+        "``'-raw'`` to the keyword, e.g., ``'sentinel_2a-msi-4-raw'``. "
+        "Note that the prepared SRF provide a better speed versus accuracy "
+        "trade-off, but for the best accuracy, the raw SRF should be used. "
+        "Other types will be converted by :data:`.spectrum_factory`.",
+        type=".Spectrum",
+        init_type="Path or str or .Spectrum or dict or float",
```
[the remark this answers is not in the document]
nollety: No, this is a remnant of `UniformSpectrum` I think

src/eradiate/pipelines/_assemble.py
```diff
@@ -67,35 +74,30 @@ class AddIllumination(PipelineStep):
...
     def transform(self, x: t.Any) -> t.Any:
+        logger.debug("add_illumination: begin")
         k_irradiance_units = uck.get("irradiance")
         illumination = self.illumination
         measure = self.measure
         result = x.copy(deep=False)
-
         # Collect spectral coordinate values for verification purposes
-        wavelengths_dataset = to_quantity(x.w)
+        wavelengths_dataset = np.array(x.w.values, dtype=float) * ureg(x.w.units)
```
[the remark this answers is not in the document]
nollety: I dont remember clearly why; I tried with `to_quantity` and that seems to work so I came back to that

src/eradiate/pipelines/_gather.py
```diff
@@ -78,7 +83,16 @@ def transform(self, x: dict) -> xr.Dataset:
         sensor_datasets = []
 
         # Loop on spectral indexes
-        for spectral_index, result_dict in x.items():
+        for siah, result_dict in x.items():
+            if eradiate.mode().is_mono:
+                spectral_index = siah
+            elif eradiate.mode().is_ckd:
+                spectral_index = (
+                    str(int(siah[0])),
```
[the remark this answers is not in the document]
nollety: At this point in time, it is not expected to have non-integer bin centres; we support only the 1 m and 10 nm bin sets that we have used so far. I added a `# TODO: PR#311 hack` comment to remember us to refactor that in the upcoming absorption datasets integration work.

src/eradiate/quad.py
```diff
@@ -57,6 +57,14 @@ def _nodes_weights_validator(self, attribute, value):
                 f"and weights.shape = {self.weights.shape}"
             )
 
+    def __hash__(self) -> int:
```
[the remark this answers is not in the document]
nollety: I can't see why; this may goes back to a time when I wanted to have bins in a set for `BinSet` objects. Now these `BinSet` objects take a `list[Bin]` as parameter so that hash is useless. Removing it.

src/eradiate/radprops/_afgl1986.py
```diff
@@ -129,80 +167,119 @@ def _thermoprops_interp(self, zgrid: ZGrid) -> xr.Dataset:
...
+    def eval_albedo(
+        self,
+        si: SpectralIndex,
+        zgrid: ZGrid,
+    ) -> pint.Quantity:
+        raise NotImplementedError
+
+    @eval_albedo.register
+    def _(self, si: CKDSpectralIndex, zgrid: ZGrid) -> pint.Quantity:
+        return self.eval_albedo_ckd(si.w, si.g, zgrid)
```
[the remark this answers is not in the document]
nollety: This should not have remained here indeed, removing that boiler plate code.
Fixing type hints in a subsequent commit

src/eradiate/scenes/phase/_tabulated.py
```diff
@@ -146,34 +144,32 @@ def eval_mono(self, w: pint.Quantity) -> np.ndarray:
...
 
         Parameters
         ----------
-        *bindexes : :class:`.Bindex`
-            One or several CKD bindexes for which to evaluate the phase
-            function.
+        w : :class:`pint.Quantity`
+            Spectral bin center wavelength.
+
+        g: float
```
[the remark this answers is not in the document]
nollety: Gosh, I always forget about that, sorry

src/eradiate/scenes/spectra/_air_scattering_coefficient.py
```diff
@@ -46,63 +34,39 @@ def eval_mono(self, w: pint.Quantity) -> pint.Quantity:
...
         raise NotImplementedError
 
     @property
     def template(self) -> dict:
         # Inherit docstring
+
+        def eval(ctx: KernelDictContext) -> float:
+            supported_mode(spectral_mode=SpectralMode.MONO | SpectralMode.CKD)
+            kernel_units = uck.get("collision_coefficient")
+            return float(self.eval(ctx.si).m_as(kernel_units))
```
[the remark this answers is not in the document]
nollety: This is not required indeed

src/eradiate/scenes/spectra/_interpolated.py
```diff
@@ -244,8 +222,7 @@ def integral(self, wmin: pint.Quantity, wmax: pint.Quantity) -> pint.Quantity:
             pass
 
         # Evaluate spectrum at wavelengths
-        w.sort()
-        w = w * wavelength_units
+        w = np.array(w) * wavelength_units
```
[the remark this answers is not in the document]
nollety: I dont understand your comment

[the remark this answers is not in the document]
nollety: I removed the sort operation, because wavelengths values are required to be sorted

src/eradiate/scenes/spectra/_solar_irradiance.py
```diff
@@ -195,69 +195,38 @@ def eval_mono(self, w: pint.Quantity) -> pint.Quantity:
...
-            # -- Evaluate spectrum at wavelengths
-            interp = self.eval_mono(w)
-
-            # -- Average spectrum on bin extent
-            integral = np.trapz(interp, w)
-            result[i_bindex] = (integral / bin.width).m_as(quantity_units)
-
-        return result * quantity_units
+    def eval_ckd(self, w: pint.Quantity, g: float) -> pint.Quantity:
+        return self.eval_mono(w=w)
```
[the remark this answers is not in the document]
nollety: Yes, this is planned for later (see: https://app.clickup.com/t/861mn6rwy)

nollety:
Good news: the reason for these failing tests was that some scene element had `slots=True` which is not allowed since SceneElement has `slots=False`.

nollety:
I updated the PR description with respect to the implemented suggestions

src/eradiate/radprops/_afgl1986.py
```diff
@@ -87,6 +92,37 @@ class AFGL1986RadProfile(RadProfile):
         default="True",
     )
 
+    absorption_dataset: t.Optional[t.Union[PathLike, xr.Dataset]] = documented(
```
leroyvn, line 95:
```suggestion
    absorption_dataset: xr.Dataset | None = documented(
```

src/eradiate/radprops/_afgl1986.py
```diff
@@ -4,20 +4,23 @@
 
 from __future__ import annotations
 
-import functools
 import typing as t
+import warnings
+from functools import partial, singledispatchmethod
+from os import PathLike
```
leroyvn, line 10: If it's for typing, please use our `PathLike` construct.

src/eradiate/radprops/_afgl1986.py
```diff
@@ -87,6 +92,37 @@ class AFGL1986RadProfile(RadProfile):
...
+            converter=attrs.converters.optional(
+                absorption_dataset_convert
+            ),  # TODO: open dataset (not load)
+            validator=attrs.validators.optional(
+                attrs.validators.instance_of(xr.Dataset)
+            ),
+        ),
+        doc="Absorption coefficient dataset. If ``None``, the absorption "
+        "coefficient is set to zero.",
+        type=":class:`~xarray.Dataset`",
```
leroyvn, line 107:
```suggestion
        type=":class:`~xarray.Dataset` or None",
```

src/eradiate/radprops/_us76_approx.py
```diff
@@ -146,17 +146,20 @@ class US76ApproxRadProfile(RadProfile):
         default="True",
     )
 
-    absorption_data_set: str | None = documented(
+    absorption_dataset: PathLike | xr.Dataset | None = documented(
```
leroyvn, line 149: AFAIK you're not storing the path, event if you allow it in the constructor:
```suggestion
    absorption_dataset: xr.Dataset | None = documented(
```

src/eradiate/scenes/atmosphere/_core.py
```diff
@@ -129,6 +132,10 @@ def height(self) -> pint.Quantity:
         """
         return self.top - self.bottom
 
+    @property
+    def spectral_set(self) -> None | BinSet | WavelengthSet:
```
leroyvn, line 136: Maybe add a docstring?

src/eradiate/scenes/atmosphere/_homogeneous.py
```diff
@@ -142,80 +143,26 @@ def phase(self) -> PhaseFunction:
...
-        Return albedo.
+    def eval_albedo(self, si: SpectralIndex) -> pint.Quantity:
+        """Return albedo at given spectral index."""
+        return self.eval_sigma_s(si) / (self.eval_sigma_s(si) + self.eval_sigma_a(si))
 
-        Parameters
-        ----------
-        spectral_ctx : .SpectralContext
-            A spectral context data structure containing relevant spectral
-            parameters (*e.g.* wavelength in monochromatic mode).
```
leroyvn, line 158: Deleting all docstrings isn't a good idea IMO: the information about the returned type, in particular, is very useful.

src/eradiate/scenes/atmosphere/_molecular_atmosphere.py
```diff
@@ -98,21 +105,21 @@ def _switch_validator(self, attribute, value):
...
-        "set files whose paths are provided in the mapping will be used to "
-        "compute the absorption coefficient. If the mapping does not "
-        "include all species from the atmospheric "
-        "thermophysical profile, the default data sets will be used to "
-        "compute the absorption coefficient of the corresponding species.",
-        type="dict",
+        doc="Absorption coefficient dataset. If ``None``, the absorption "
+        "coefficient is set to zero.",
+        type=":class:`~xarray.Dataset`",
+        init_type="PathLike or :class:`~xarray.Dataset`",
```
leroyvn, line 121: Use Sphinx aliases for clarity in the init type field:
```suggestion
        init_type="path-like or Dataset`",
```

src/eradiate/scenes/atmosphere/_core.py
```diff
@@ -129,6 +132,10 @@ def height(self) -> pint.Quantity:
...
+        return None
```
leroyvn, line 137: You're providing a default implementation, but you also reimplement this default in some child classes. It would be safer IMO to make this property abstract.

src/eradiate/scenes/spectra/_air_scattering_coefficient.py
```diff
@@ -46,63 +34,37 @@ def eval_mono(self, w: pint.Quantity) -> pint.Quantity:
...
+
         return {
             "type": "uniform",
-            "value": InitParameter(
-                lambda ctx: float(
-                    self.eval(ctx.spectral_ctx).m_as(uck.get("collision_coefficient")),
-                )
-            ),
+            "value": InitParameter(evaluator=eval),
```
leroyvn, line 54: I thought you had rolled this back, what happened here?

src/eradiate/scenes/spectra/_core.py
```diff
@@ -141,40 +149,45 @@ def _quantity_validator(self, attribute, value):
...
 
         Returns
         -------
         value : quantity
-            Evaluated spectrum as a scalar.
+            Evaluated spectrum.
+
+        Notes
+        -----
+        This method dispatches evaluation to specialised methods depending
```
leroyvn, line 169: Oxford spelling
```suggestion
        This method dispatches evaluation to specialized methods depending
```

src/eradiate/scenes/spectra/_interpolated.py
```diff
@@ -254,23 +234,128 @@ def integral(self, wmin: pint.Quantity, wmax: pint.Quantity) -> pint.Quantity:
...
+            return value
+
         return {
             "type": "uniform",
-            "value": InitParameter(
-                lambda ctx: float(
-                    self.eval(ctx.spectral_ctx).m_as(uck.get(self.quantity))
-                )
-            ),
+            "value": InitParameter(evaluator=eval),
```
leroyvn, line 245: Ditto

src/eradiate/scenes/spectra/_interpolated.py
```diff
@@ -254,23 +234,128 @@ def integral(self, wmin: pint.Quantity, wmax: pint.Quantity) -> pint.Quantity:
...
+            return value
+
         return {
             "value": UpdateParameter(
-                lambda ctx: float(
-                    self.eval(ctx.spectral_ctx).m_as(uck.get(self.quantity))
-                ),
-                UpdateParameter.Flags.SPECTRAL,
+                evaluator=eval,
+                flags=UpdateParameter.Flags.SPECTRAL,
```
leroyvn, line 260: Ditto

src/eradiate/spectral/ckd.py
```diff
@@ -0,0 +1,269 @@
...
+                member_validator=attrs.validators.instance_of(Bin)
+            ),
+            repr=lambda x: "\n        "
+            + "\n        ".join(
+                [
+                    bin.pretty_repr()
+                    for bin in sorted(x, key=lambda b: b.wcenter.m_as("nm"))
+                ]
+            )
+            + "\n",
```
leroyvn, line 140: Please avoid multiline reprs, they break easily either the inline or the pretty repr of the object.

leroyvn (changes requested):
Hi @nollety, we're getting there. I had another round of review, this time it's more about details.

src/eradiate/experiments/_core.py
```diff
@@ -119,6 +132,53 @@ def results(self) -> dict[str, xr.Dataset]:
...
+        ":class:`.BinSet` or :class:`.WavelengthSet` instance.",
+        type=":class:`.BinSet` or :class:`.WavelengthSet`",
+        init_type=":class:`.BinSet` or :class:`.WavelengthSet` or :data:`AUTO`",
+        default=":data:`AUTO`",
+    )
+
+    @default_spectral_set.validator
+    def _default_spectral_set_validator(self, attribute, value):
+        if not isinstance(value, (WavelengthSet, BinSet)):
+            raise ValueError(f"invalid spectral set: {value}")
```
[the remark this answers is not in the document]
nollety: That's true

src/eradiate/experiments/_canopy_atmosphere.py
```diff
@@ -173,9 +183,20 @@ def integrator(self) -> Integrator:
...
+
+        self._normalize_spectral()
         self._normalize_atmosphere()
         self._normalize_measures()
 
+    def get_spectral_set(self) -> dict[int, BinSet | WavelengthSet]:
+        # Overriding parent implementation because the atmosphere
+        # might be molecular, hence provide authoritative information
+        # about the spectral set
+        return get_spectral_set_impl(exp=self)
```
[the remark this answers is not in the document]
nollety: This `spectral_set` method would be used only once, namely in `Experiment.get_spectral_set`, which in turn is only called by `Experiment._normalize_spectral()`, so I moved the implementation to `_normalize_spectral()`.

src/eradiate/radprops/_afgl1986.py
nollety: Outdated

src/eradiate/scenes/atmosphere/_core.py
nollety: Of course

src/eradiate/scenes/atmosphere/_homogeneous.py
nollety: I found these docstrings somehow redundant, they all over the place and they are exactly the same. I thought there was a way to write the docstrings once (in `Atmosphere`) and then inherit docstrings in subclasses, isn't that possible?
Otherwise, I'll gladly restore the docstrings, all that was still work in progress

src/eradiate/scenes/atmosphere/_core.py
nollety: Good point

src/eradiate/scenes/spectra/_air_scattering_coefficient.py
nollety: I thought I had this rolled back too, perhaps I forgot to commit those changes

nollety:
@leroyvn Let me know when I can squash the commits, so I can rebase on latest main

src/eradiate/scenes/atmosphere/_homogeneous.py
leroyvn: These methods haven't made it to the `Atmosphere` interface yet, have they?

src/eradiate/pipelines/_assemble.py
```diff
@@ -81,23 +89,19 @@ def eval_illumination_spectrum(
             # Local helper function to help with illumination spectrum evaluation
 
             spectrum: Spectrum = getattr(illumination, field_name)
+            wavelengths = to_quantity(x.w)
 
             if eradiate.mode().is_mono:
-                # Very important: sort spectral coordinate
-                wavelengths = np.sort(measure.spectral_cfg.wavelengths)
+                # Very important: sort spectral coordinate  # TODO: is it?
```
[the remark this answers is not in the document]
nollety: I think this is not necessary because it is garuanteed by the current SRF-compatible `Spectrum` subclasses:
* In `InterpolatedSpectrum`, `wavelengths` validator will raise if values are not monotonically increasing
* In `MultiDeltaSpectrum`, `wavelengths` converter will sort them

src/eradiate/scenes/atmosphere/_homogeneous.py
nollety: No, you are right.
They are in `AbstractHeterogeneousAtmosphere` and `HomogeneousAtmosphere`, the difference (as far as I can see) between the `eval_*` methods in these two is the presence of the `zgrid` parameter.

src/eradiate/scenes/spectra/_core.py
```diff
@@ -217,3 +241,21 @@ def integral(self, wmin: pint.Quantity, wmax: pint.Quantity) -> pint.Quantity:
             Computed integral value.
         """
         raise NotImplementedError
+
+    @singledispatchmethod
+    def select_in(self, spectral_set) -> Spectrum:
```
leroyvn, line 246: Docstring is missing

src/eradiate/scenes/spectra/_core.py
```diff
@@ -217,3 +241,21 @@ def integral(self, wmin: pint.Quantity, wmax: pint.Quantity) -> pint.Quantity:
...
+
+    @select_in.register(WavelengthSet)
+    def _(self, spectral_set) -> Spectrum:
+        return self.select_in_wavelength_set(spectral_set)
+
+    @select_in.register(BinSet)
+    def _(self, spectral_set) -> Spectrum:
+        return self.select_in_bin_set(spectral_set)
+
+    def select_in_wavelength_set(self, wset) -> Spectrum:
```
leroyvn, line 257: Docstring is missing (please mention that the default implementation raises `NotImplementedError`)

src/eradiate/scenes/spectra/_core.py
```diff
@@ -217,3 +241,21 @@ def integral(self, wmin: pint.Quantity, wmax: pint.Quantity) -> pint.Quantity:
...
+        raise NotImplementedError
+
+    def select_in_bin_set(self, binset) -> Spectrum:
```
leroyvn, line 260: Ditto

leroyvn (commented):
I made a few more minor comments. I think a squash is then appropriate.

src/eradiate/spectral/ckd.py
```diff
@@ -0,0 +1,320 @@
...
+          nanometers.
+        * the absorption dataset has a ``bin_set`` attribute with the bin
+          width in nanometers.
+        """
+        if quad is None:
+            quad = Quad.gauss_legendre(8)
+
+        # TODO: PR#311 hack (next 2 lines)
+        wavelengths = np.unique(np.array(dataset.bin.values, dtype=float)) * ureg.nm
+        width = ureg(dataset.attrs["bin_set"])
```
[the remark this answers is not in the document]
nollety: It does that, yes

src/eradiate/spectral/ckd.py
```diff
@@ -0,0 +1,320 @@
...
+        * the absorption dataset has a ``bin`` coordinate with values in
+          nanometers.
+        * the absorption dataset has a ``bin_set`` attribute with the bin
+          width in nanometers.
+        """
+        if quad is None:
+            quad = Quad.gauss_legendre(8)
+
+        # TODO: PR#311 hack (next 2 lines)
+        wavelengths = np.unique(np.array(dataset.bin.values, dtype=float)) * ureg.nm
```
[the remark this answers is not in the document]
nollety: If I just `np.array(dataset.bin.values)`, I get:
```python
array(['280', '280', '280', ..., '2400', '2400', '2400'], dtype=object)
```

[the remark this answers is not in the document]
leroyvn: Ah right, this is because `bins` is not a dimension. So this line is: "convert to float and deduplicate".

nollety:
@leroyvn I am marking this as ready, not because it is, but rather because I have reached a point from where I am confident that the upcoming CKD integration can proceed relatively smoothly. We have collected many ideas of possible improvements, most of them are listed [here](https://app.clickup.com/t/861mn6rwy), for future reference.
I have updated the changelog (moved to upcoming release section), and rebased on the latest main. I have added a user guide page that illustrate with examples (and figures) how the spectral discretization is set by the experiment, depending on the different relevant factors, i.e., the active mode, the default discretization, the (absorbing) atmosphere spectral discretization, and the spectral respone function. I hope the corresponding behaviour is what you expected. Some adjustments are always an option. All tests passed on my end, but I have observed strange behaviour with our tests suite; perhaps it would be test if you'd run them too.

PR #319, opened 2023-04-12
Fix atmospheric property setup and update

leroyvn:
# Description

This PR mainly fixes incorrect volume texture transforms with spherical shell geometries. This required substantial modifications to the geometric description.

The key point is that `BlendPhaseFunction`, when emitting its a kernel dict template, has to set volume textures. The initial interface simply foresaw the plane parallel case and required only an axis-aligned bounding box for a correct setup; the spherical shell geometry, however, requires to set the min and max radii of the `sphericalcoordsvolume` plugins, which goes beyond what a `BoundingBox` can do. Clearly, more geometric information had to be passed to `BlendPhaseFunction`.

I therefore decided to centralize the geometric information in the `SceneGeometry` objects: I added the `SceneGeometry.ground_altitude` and `SceneGeometry.toa_altitude` members and removed the corresponding information from the `Atmosphere` line. This simplifies the setup of the `zgrid` of heterogeneous atmospheres, which now always match the scene geometry.

A consequence, however, is that atmospheric radiative properties are now always evaluated on a grid which is primarily set after the scene geometry, and not the atmospheric components' radiative properties like it used to be. For molecular atmospheres and the heterogeneous container, users should barely notice anything, except maybe that the `zgrid` no longer has to be adapted when switching between the USSA and AFGL profiles. When dealing with particle layers, however, things change significantly: a layer with an extent [0, 1] km will be evaluated on the `zgrid` which, by default, has 1201 levels between 0 and 120 km. The radiative property profile therefore contains a bunch of zeros outside of the extent of the particle layer. This disrupted greatly the testing suite of the particle layer class—which needs now a complete retrofit.

As I was going through this work and further testing the spherical shell configuration, I realized that providing a default implementation for `Atmosphere._params_phase` was a mistake: failing to provide an implementation, although critically wrong, would just go unnoticed silently. I made this property abstract and provided a missing implementation for `HeterogeneousAtmosphere`.

# Detailed change log

* `Shape`: all shapes can optionally emit their AABB on their own. Only the ones for which the AABB is actually used provide an implementation for now.
* `SceneGeometry`: added atmosphere bounds (`ground_altitude` and `toa_altitude` fields). Geometries now emit their own surface and atmosphere shapes (`surface_shape` and `atmosphere_shape` properties), as well as appropriate volume data transforms (`atmosphere_volume_to_world` property).
* `SceneGeometry`: transferred zgrid data to a `zgrid` field. Instead of syncing both the `geometry` and `zgrid` fields of all heterogeneous atmosphere components, we resort to syncing only the former.
* `Atmosphere`: `zgrid` is transferred to the `SceneGeometry` container. It is now defined as the altitude grid on which volume data is discretized when setting up the experiment, and it must be common to all atmospheric components.
* `Atmosphere`: the `bottom` and `top` properties are renamed `bottom_altitude` and `top_altitude` respectively. They are kept for convenience, but all they do now is forward the underlying geometry's `ground_altitude` and `toa_altitude`.
* `Atmosphere`: `_params_*` properties are now abstract for improved safety.
* `HeterogeneousAtmosphere`: added missing `_params_phase` implementation (this was a critical bug).
* `ParticleLayer`: the `n_layers` field is dropped.
* `BlendPhaseFunction`: the `bbox` field is replaced by a `geometry` field which references the full scene geometry. This grants the phase function access to atmosphere bounds (ground and TOA altitudes), required to set up the volume data transforms in the spherical shell geometry. Support for all scene geometries was added and fixed (this was a critical bug).

# To do

- [x] Fix `BlendPhaseFunction`
- [x] Clean up messed up tests
- [x] ~~Optional: simplify the `HeterogeneousAtmosphere` class and replace its `molecular_atmosphere` and `particle_layers` fields with a single `components` one~~

# Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -102,6 +102,11 @@ Entries marked with a  ︎⚠️ symbol require particular attention during upgr
 * Added absorption and scattering bypass switches to the {class}`.ParticleLayer`
   class ({ghpr}`316`).
 * ⚠️ Moved Mitsuba logs to the `mitsuba` logger ({ghpr}`318`).
+* ⚠️ Centralized geometric information to `SceneGeometry` ({ghpr}`319`).
+* ⚠️ Made {class}`Atmosphere`'s `_params_*` properties abstract for improved
+  safety ({ghpr}`319`).
+* Fixed a major issue in volume definitions and parameter updates of
+  {class}`BlendPhaseFunction` ({ghpr}`319`).
 
 ### Documentation
 
--- a/src/eradiate/experiments/_atmosphere.py
+++ b/src/eradiate/experiments/_atmosphere.py
@@ -55,7 +55,10 @@ class AtmosphereExperiment(EarthObservationExperiment):
                 (PlaneParallelGeometry, SphericalShellGeometry)
             ),
         ),
-        doc="Problem geometry.",
+        doc="Problem geometry. Can be specified as a simple string "
+        '(``"plane_parallel" or "spherical_shell"``), a dictionary interpreted '
+        "by :meth:`.SceneGeometry.convert`, or a :class:`.SceneGeometry` "
+        "instance.",
         type=".PlaneParallelGeometry or .SphericalShellGeometry",
         init_type="str or dict or .SceneGeometry",
         default='"plane_parallel"',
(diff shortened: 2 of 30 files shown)


leroyvn:
All tests passed, merging now!

PR #320, opened 2023-04-18
Add spherical shell geometry support to DEM components

schunkes:
# Description

This PR updates DEM-related components to allow using a spherical shell geometry with the `DEMExperiment` class.

# Checklist

### DIFF
--- a/src/eradiate/experiments/_dem.py
+++ b/src/eradiate/experiments/_dem.py
@@ -5,24 +5,20 @@ import warnings
 
 import attrs
 
-from ._core import EarthObservationExperiment, Experiment
-from ._helpers import (
-    measure_inside_atmosphere,
-    surface_converter,
-)
-from ..attrs import documented, get_doc, parse_docs
-from ..scenes.atmosphere import (
-    Atmosphere,
-    HomogeneousAtmosphere,
-    atmosphere_factory,
-)
+from ._core import EarthObservationExperiment
+from ._helpers import measure_inside_atmosphere, surface_converter
+from ..attrs import documented, parse_docs
+from ..scenes.atmosphere import Atmosphere, HomogeneousAtmosphere, atmosphere_factory
 from ..scenes.bsdfs import LambertianBSDF
 from ..scenes.core import SceneElement
-from ..scenes.geometry import PlaneParallelGeometry, SceneGeometry
+from ..scenes.geometry import (
+    PlaneParallelGeometry,
+    SceneGeometry,
+    SphericalShellGeometry,
+)
 from ..scenes.integrators import Integrator, VolPathIntegrator, integrator_factory
 from ..scenes.measure import DistantMeasure, Measure, TargetPoint
-from ..scenes.shapes import RectangleShape
-from ..scenes.surface import BasicSurface, DEMSurface, surface_factory
+from ..scenes.surface import BasicSurface, DEMSurface
 from ..units import unit_registry as ureg
 
 
@@ -46,19 +42,25 @@ class DEMExperiment(EarthObservationExperiment):
       unsuitable configuration is detected, a :class:`ValueError` will be raised
       during initialization.
 
-    * Currently this experiment is limited to the plane-parallel geometry.
+    * Currently DEM surfaces, which extends below 0 elevation will not yield correct
+      results, when used in combination with atmospheres.
+
+    * When using a DEM which extends below 0 elevation, a .VolPathIntegrator must be
+      used. Otherwise, the low regions of the DEM will be shadowed by the background
+      surface.
     """
 
-    # Currently, only the plane parallel geometry is supported
-    geometry: PlaneParallelGeometry = documented(
+    geometry: SceneGeometry = documented(
         attrs.field(
             default="plane_parallel",
             converter=SceneGeometry.convert,
-            validator=attrs.validators.instance_of(PlaneParallelGeometry),
+            validator=attrs.validators.instance_of(
+                (PlaneParallelGeometry, SphericalShellGeometry)
+            ),
         ),
         doc="Problem geometry.",
-        type=".PlaneParallelGeometry",
-        init_type="str or dict or .PlaneParallelGeometry",
+        type=".SceneGeometry",
+        init_type="str or dict or .PlaneParallelGeometry or .SphericalShellGeometry",
         default='"plane_parallel"',
     )
 
@@ -79,47 +81,34 @@ class DEMExperiment(EarthObservationExperiment):
         default=":class:`HomogeneousAtmosphere() <.HomogeneousAtmosphere>`",
     )
 
-    surface: BasicSurface | None = documented(
+    surface: BasicSurface | DEMSurface | None = documented(
         attrs.field(
             factory=lambda: BasicSurface(bsdf=LambertianBSDF()),
             converter=attrs.converters.optional(surface_converter),
             validator=attrs.validators.optional(
-                attrs.validators.instance_of(BasicSurface)
+                attrs.validators.instance_of((BasicSurface, DEMSurface))
             ),
         ),
         doc="Surface specification. If set to ``None``, no surface will be "
         "added. This parameter can be specified as a dictionary which will be "
         "interpreted by :data:`.surface_factory` and :data:`.bsdf_factory`.",
-        type=".BasicSurface or None",
-        init_type=".BasicSurface or .BSDF or dict, optional",
+        type=".Surface or None",
+        init_type=".BasicSurface or .DEMSurface or .BSDF or dict, optional",
         default=":class:`BasicSurface(bsdf=LambertianBSDF()) <.BasicSurface>`",
-    )
-
-    dem: DEMSurface | None = documented(
-        attrs.field(
-            default=None,
-            converter=attrs.converters.optional(surface_factory.convert),
-            validator=attrs.validators.optional(
-                attrs.validators.instance_of(DEMSurface)
-            ),
-        ),
-        doc="Digital elevation model (DEM) specification. If set to ``None``, no DEM will be "
-        "added. This parameter can be specified as a dictionary which will be "
-        "interpreted by :data:`.surface_factory`",
-        type=".DEMSurface or None",
-        init_type=".DEMSurface or dict, optional",
-        default="None",
     )
 
     _integrator: Integrator = documented(
         attrs.field(
             factory=VolPathIntegrator,
             converter=integrator_factory.convert,
-            validator=attrs.validators.instance_of(Integrator),
+            validator=attrs.validators.instance_of(VolPathIntegrator),
         ),
-        doc=get_doc(Experiment, attrib="_integrator", field="doc"),
-        type=get_doc(Experiment, attrib="_integrator", field="type"),
-        init_type=get_doc(Experiment, attrib="_integrator", field="init_type"),
+        doc="Monte Carlo integration algorithm specification. "
+        "This parameter can be specified as a dictionary which will be "
+        "interpreted by :data:`.integrator_factory`. The DEMExperiment requires"
+        "the use of a .VolPathIntegrator.",
+        type=":class:`.VolPathIntegrator`",
+        init_type=":class:`.VolPathIntegrator` or dict",
         default=":class:`VolPathIntegrator() <.VolPathIntegrator>`",
     )
 
@@ -144,7 +133,7 @@ class DEMExperiment(EarthObservationExperiment):
         for measure in self.measures:
             # Override ray target location if relevant
             if isinstance(measure, DistantMeasure):
-                if self.dem is not None:
+                if isinstance(self.surface, DEMSurface):
                     if measure.target is None:
                         msg = (
                             f"Measure '{measure.id}' has its target unset "
@@ -200,27 +189,18 @@ class DEMExperiment(EarthObservationExperiment):
 
         # Process atmosphere
         if self.atmosphere is not None:
-            objects["atmosphere"] = self.atmosphere
-
-        # Process surface
-        if self.surface is not None:
-            if self.atmosphere is not None:
-                surface_width = self.atmosphere.geometry.width
-                surface_altitude = self.atmosphere.bottom_altitude
-            else:
-                surface_width = self._default_surface_width
-                surface_altitude = 0.0 * ureg.km
-
-            objects["surface"] = attrs.evolve(
-                self.surface,
-                shape=RectangleShape.surface(
-                    altitude=surface_altitude, width=surface_width
-                ),
+            objects["atmosphere"] = attrs.evolve(
+                self.atmosphere, geometry=self.geometry
             )
 
-        # Process DEM
-        if self.dem is not None:
-            objects["surface"] = self.dem
+        # Process surface
+        if self.surface is not None and isinstance(self.surface, BasicSurface):
+            objects["surface"] = attrs.evolve(
+                self.surface,
+                shape=self.geometry.surface_shape,
+            )
+        else:
+            objects["surface"] = self.surface
 
         objects.update(
             {
--- a/src/eradiate/experiments/_dem.py
+++ b/src/eradiate/experiments/_dem.py
@@ -19,15 +19,20 @@ from ..scenes.geometry import (
 from ..scenes.integrators import Integrator, VolPathIntegrator, integrator_factory
 from ..scenes.measure import DistantMeasure, Measure, TargetPoint
 from ..scenes.surface import BasicSurface, DEMSurface
-from ..units import unit_registry as ureg
 
 
 @parse_docs
 @attrs.define
 class DEMExperiment(EarthObservationExperiment):
     """
-    Simulate radiation in a scene with a digital elevation model (DEM)
-    under a 1D atmosphere.
+    Simulate radiation in a scene with a digital elevation model (DEM) under a
+    1D atmosphere.
+
+    Warnings
+    --------
+    * Although technically supported, DEMs extending below 0 elevation may be
+      a tricky case because atmospheric profile behaviour below sea level is
+      undefined. This will be addressed in a future release.
 
     Notes
     -----
@@ -42,12 +47,8 @@ class DEMExperiment(EarthObservationExperiment):
       unsuitable configuration is detected, a :class:`ValueError` will be raised
       during initialization.
 
-    * Currently DEM surfaces, which extends below 0 elevation will not yield correct
-      results, when used in combination with atmospheres.
-
-    * When using a DEM which extends below 0 elevation, a .VolPathIntegrator must be
-      used. Otherwise, the low regions of the DEM will be shadowed by the background
-      surface.
+    * Even without an atmosphere, this experiment requries using a volumetric
+      path tracing integrator.
     """
 
     geometry: SceneGeometry = documented(
@@ -60,7 +61,8 @@ class DEMExperiment(EarthObservationExperiment):
         ),
         doc="Problem geometry.",
         type=".SceneGeometry",
-        init_type="str or dict or .PlaneParallelGeometry or .SphericalShellGeometry",
+        init_type='{"plane_parallel", "spherical_shell"} or dict or '
+        ".PlaneParallelGeometry or .SphericalShellGeometry",
         default='"plane_parallel"',
     )
 
@@ -76,8 +78,8 @@ class DEMExperiment(EarthObservationExperiment):
         "be added. "
         "This parameter can be specified as a dictionary which will be "
         "interpreted by :data:`.atmosphere_factory`.",
-        type=":class:`.Atmosphere` or None",
-        init_type=":class:`.Atmosphere` or dict or None",
+        type=".Atmosphere or None",
+        init_type=".Atmosphere or dict or None",
         default=":class:`HomogeneousAtmosphere() <.HomogeneousAtmosphere>`",
     )
 
@@ -107,8 +109,8 @@ class DEMExperiment(EarthObservationExperiment):
         "This parameter can be specified as a dictionary which will be "
         "interpreted by :data:`.integrator_factory`. The DEMExperiment requires"
         "the use of a .VolPathIntegrator.",
-        type=":class:`.VolPathIntegrator`",
-        init_type=":class:`.VolPathIntegrator` or dict",
+        type=".VolPathIntegrator",
+        init_type=".VolPathIntegrator or dict",
         default=":class:`VolPathIntegrator() <.VolPathIntegrator>`",
     )
 
@@ -156,10 +158,6 @@ class DEMExperiment(EarthObservationExperiment):
 
                 if msg is not None:
                     warnings.warn(UserWarning(msg))
-
-    @property
-    def _default_surface_width(self):
-        return 1.0 * ureg.km
 
     def _dataset_metadata(self, measure: Measure) -> dict[str, str]:
         result = super()._dataset_metadata(measure)
--- a/src/eradiate/scenes/bsdfs/_opacity_mask.py
+++ b/src/eradiate/scenes/bsdfs/_opacity_mask.py
@@ -2,12 +2,28 @@ from __future__ import annotations
 
 import attrs
 import mitsuba as mi
+import numpy as np
 
 from ._core import BSDF, bsdf_factory
 from ._lambertian import LambertianBSDF
 from ..core import traverse
+from ... import converters
 from ...attrs import documented, parse_docs
 from ...kernel import TypeIdLookupStrategy, UpdateParameter
+
+
+def _to_bitmap(value):
+    if isinstance(value, mi.Bitmap):
+        return value
+
+    elif isinstance(value, np.ndarray):
+        return mi.Bitmap(value)
+
+    elif isinstance(value, list):
+        return mi.Bitmap(np.array(value))
+
+    else:
+        return value
 
 
 @parse_docs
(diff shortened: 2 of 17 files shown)


src/eradiate/experiments/_dem.py
```diff
@@ -119,6 +125,40 @@ class DEMExperiment(EarthObservationExperiment):
     def __attrs_post_init__(self):
         self._normalize_atmosphere()
         self._normalize_measures()
+        self._normalize_surface()
+
+    def _normalize_surface(self) -> None:
```
leroyvn, line 130: You'll also have to check if the other normalization steps shouldn't take the DEM into account (*e.g* distant measure targeting and atmosphere geometry).

src/eradiate/scenes/surface/_dem.py
```diff
@@ -1,61 +1,204 @@
...
+    """
+    vertices = []
+    lat_range = list(latitude_points)
+    lon_range = list(longitude_points)
+
+    for i, theta in enumerate(lat_range):
+        for j, phi in enumerate(lon_range):
+            z = elevation[i, j].m_as(ucc.get("length"))
+            vertices.append([theta, phi, z])
+    return vertices
```
leroyvn, line 58: This function has a few issues.

* The loop-based implementation is wasteful (and the in-loop Pint conversion adds even more penalty). The `z` array is just [`elevation.ravel()`](https://numpy.org/doc/stable/reference/generated/numpy.ravel.html) (maybe you'll have to transpose, I can't remember exactly) and the `vertices` array is then `theta.ravel()`, `phi.ravel()` and `elevation.ravel()` stacked. Such a formulation will be vectorized automatically by Numpy.
* The returned value has no units. This is fine for internal implementation details, but at least then please mention the mandatory conversion in the docs. And it's then best to convert to fixed units, not context-dependent ones. Good news though: if you vectorize this, applying Pint units will be cheap.
* Outdated docstring.

src/eradiate/scenes/surface/_dem.py
```diff
@@ -1,61 +1,204 @@
...
+    """
+    vertices_new = []
+    planet_radius_m = planet_radius.m_as(ucc.get("length"))
+    for theta, phi, elevation in vertices:
+        theta_r = np.deg2rad(90 - theta)
+        phi_r = np.deg2rad(phi)
+        x = np.sin(theta_r) * np.cos(phi_r) * (planet_radius_m + elevation)
+        y = np.sin(theta_r) * np.sin(phi_r) * (planet_radius_m + elevation)
+        z = np.cos(theta_r) * (planet_radius_m + elevation)
+        vertices_new.append((x, y, z))
```
leroyvn, line 93: You should vectorize this, it will be easy since all these arrays are linear.

src/eradiate/scenes/surface/_dem.py
```diff
@@ -1,61 +1,204 @@
...
+    lat_scale = 60 / lat_range
+
+    lon_mean = (lon_max + lon_min) / 2.0
+    lon_uv = lon_mean / 360 + 0.5
 
-    id: str | None = documented(
+    lat_mean = (lat_max + lat_min) / 2.0
+    lat_uv = 0.5 - (lat_mean / 180)
+
+    return mi.Transform4f.scale((lon_scale, lat_scale, 1)) @ mi.Transform4f.translate(
```
leroyvn, line 183: This will break in JIT compiled modes. You must use `mi.ScalarTransform4f`.

src/eradiate/scenes/surface/_dem.py
```diff
@@ -1,61 +1,204 @@
...
+    """
+    thetas = []
+    phis = []
+    for theta, phi, _ in vertices:
+        phis.append(phi)
+        thetas.append(theta)
+
+    phis = sorted(phis)
+    thetas = sorted(thetas)
+    return (thetas[0], thetas[-1]), (phis[0], phis[-1])
```
leroyvn, line 128: Why not use `np.min()` and `np.max()`?

src/eradiate/scenes/surface/_dem.py
```diff
@@ -1,61 +1,204 @@
...
+    id: t.Optional[str] = documented(
         attrs.field(
             default="terrain",
             validator=attrs.validators.optional(attrs.validators.instance_of(str)),
         ),
         doc=get_doc(Surface, "id", "doc"),
         type=get_doc(Surface, "id", "type"),
         init_type=get_doc(Surface, "id", "init_type"),
-        default='"surface"',
+        default="surface",
```
leroyvn, line 201: Previous was fine, the quotes denote a string.

src/eradiate/scenes/surface/_dem.py
```diff
@@ -65,12 +208,46 @@ class DEMSurface(Surface):
                 attrs.validators.instance_of((BufferMeshShape, FileMeshShape))
             ),
             kw_only=True,
+            default=None,
         ),
         doc="Shape describing the surface.",
-        type=".BufferMeshShape or .FileMeshShape",
+        type=".BufferMeshShape or .FileMeshShape or None",
         init_type=".BufferMeshShape or .OBJMeshShape or .PLYMeshShape or dict",
+        default="None",
```
leroyvn, line 216: Why do you allow `None`?

src/eradiate/scenes/surface/_dem.py
```diff
@@ -83,224 +260,250 @@ class DEMSurface(Surface):
...
+            )
+            surface_background = SphereShape.surface(
+                id=f"{self.id}_background",
+                altitude=atmo_bottom,
+                planet_radius=geometry.planet_radius,
+                to_world=trafo,
+                bsdf=opacity_bsdf,
+            )
+        elif isinstance(geometry, PlaneParallelGeometry):
+            planet_radius = 6378 * ureg.km
```
leroyvn, line 343: Please use `geometry.planet_radius`

leroyvn (commented):
Hi @schunkes, this is a first very quick review. Key points:

* Please make the scope of this PR clear and as limited as possible, *i.e.* stick to extending `DEMExperiment` with spherical shell geometry support.
* Please review the whole DEM processing code and vectorize it, it will be clearer, more compact and much faster.

src/eradiate/scenes/surface/_dem.py
schunkes: The values inside each tuple have different units. I can't just attach one unit to the entire list, because it would not be correct.

src/eradiate/scenes/surface/_dem.py
schunkes: I can't. This is the PlaneParallelGeometry, which does not hold a planet radius. Maybe I should explain my approach a little: When the dem class is instantiated, a BufferMeshShape is created, which holds the vertex positions given as (latitude, longitude, elevation) tuples, in [degrees, degrees, ucc["length"]] units. from that, I convert to cartesian coordinates later, in the `evolve_dem` method. Therefore, I need a planet radius even in the plane parallel case, in order to convert lat/lon coordinates to cartesian. This approach is as hacky as any other idea I had for the dem...

src/eradiate/scenes/bsdfs/_opacity_mask.py
```diff
@@ -0,0 +1,86 @@
...
+        for key, param in traverse(self.inner_bsdf)[1].items():
+            result[f"nested_bsdf.{key}"] = attrs.evolve(
+                param,
+                lookup_strategy=TypeIdLookupStrategy(
+                    node_type=mi.BSDF,
+                    node_id=self.id,
+                    parameter_relpath=f"nested_bsdf.{key}",
+                )
+                if self.id is not None
+                else None,
```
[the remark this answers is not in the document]
schunkes: That is a good point, but this code snippet is present in all other BSDF classes as well...

src/eradiate/experiments/_dem.py
schunkes: We have to discuss this during the Eradiate meeting.

src/eradiate/scenes/surface/_dem.py
leroyvn: Right, I misread; then I think this should be tunable and default to `eradiate.constants.EARTH_RADIUS`.

src/eradiate/scenes/surface/_dem.py
```diff
@@ -83,224 +295,256 @@ class DEMSurface(Surface):
...
-        for i, x in enumerate(lat_range):
-            if i in [0, len_lat + 1]:
-                # For the rows of vertices which are only the lower points of the wall
-                # set the z coordinate to -1
-                z_list = np.full(len_lon + 2, -1.0)
+        theta_lim, phi_lim = _minmax_coordinates(self.shape.vertices)
+        theta_mean, phi_mean = _mean_coordinates(self.shape.vertices)
+        try:
+            atmo_bottom = geometry.ground_altitude
+        except AttributeError:
```
leroyvn, line 322: When are you expecting this to happen? This attribute is defined by the `SceneGeometry` interface.

src/eradiate/scenes/surface/_dem.py
```diff
@@ -83,224 +295,256 @@ class DEMSurface(Surface):
...
+                vertices = self.shape.vertices
+                vertices_spherical = _transform_vertices_spherical_shell(
+                    vertices, geometry.planet_radius
+                )
+                vertices_new = [trafo.dot(vertex) for vertex in vertices_spherical]
+            elif isinstance(geometry, PlaneParallelGeometry):
+                vertices = self.shape.vertices
+                vertices_new = _transform_vertices_plane_parallel(
+                    vertices,
+                    planet_radius=6378 * ureg.km,
```
leroyvn, line 347: Please sync this with next conditional branching.

src/eradiate/scenes/surface/_dem.py
```diff
@@ -83,224 +295,256 @@ class DEMSurface(Surface):
...
-        # triangles, extending one index in each dimension.
-        face_indices = cls._generate_face_indices(len_lat, len_lon)
-
-        return cls(
-            shape=BufferMeshShape(vertices=vertex_positions, faces=face_indices),
-            bsdf=bsdf,
+                raise ValueError(f"Unsupported atmosphere geometry: {geometry}")
+            self.shape = attrs.evolve(self.shape, vertices=vertices_new)
+            mesh = self.shape.instance
+            mesh.write_ply(str(self.mesh_filename))
```
leroyvn, line 354: Won't this break if the parent directory doesn't exist?

src/eradiate/scenes/surface/_dem.py
```diff
@@ -83,224 +295,256 @@ class DEMSurface(Surface):
...
+
     @classmethod
     def from_analytical(
         cls,
         elevation_function: t.Callable,
-        x_length: pint.Quantity | float,
-        x_steps: int,
-        y_length: pint.Quantity | float,
-        y_steps: int,
+        lat_length: t.Union[pint.Quantity, float],
```
leroyvn, line 466: New-style annotations, please.

leroyvn (changes requested):
Hi @schunkes, thanks for proceeding with vectorization. It looks like a round of cleanup is required prior to proceeding with fine tuning, notably paying attention to:

* unit handling;
* docstring and annotation agreement.

Looking again at this, I think we are in a complicated situation where file buffering fundamentally breaks unit support (we already worked around this but it's not bulletproof). 

Also, passing through a file write is something we've been trying to avoid and could be delegated to the user without losing too much convenience. The pattern would then be something like this (still breaking unit support):
```python
# Tesselate DEM
dem_mesh: mi.Mesh = make_dem_from(some_data)
# Either use it directly, 
dem_surface = DEMSurface(shape=dem_mesh, ...)
# Or through a file
dem_mesh.write_ply("dem.ply")
dem_surface = DEMSurface(shape="dem.ply")
``` 

Let's review your code live after you're done cleaning it up a bit.

src/eradiate/scenes/shapes/_rectangle.py
```diff
@@ -144,6 +145,9 @@ def surface(
         bsdf : BSDF or dict, optional, default: None
             A BSDF specification, forwarded to the main constructor.
 
+        **kwargs
+            Key word arguments forwarded to the main constructor.
```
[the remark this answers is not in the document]
schunkes: Well the mistake was, that I tried to keep the use of SphereShape.surface() and RectangleShape.surface() in DEMExperiment, but I was trying to pass additionall arguments, such as the id and the to_world transform for the sphere. Instead I am now using the default constructor for both shapes.

schunkes:
@leroyvn would you mind, taking another look at the code now? I updated it according to what we discussed. Additionally, I gave the BufferMesh a `write_ply` method. If you want, I can also show you the workflow as it is now. :)

src/eradiate/scenes/surface/_dem.py
```diff
@@ -1,50 +1,320 @@
...
+        lon = ureg.Quantity(da.lon.values, coord_units).m_as(ureg.deg)
+        vertices = _generate_dem_vertices(lat, lon, to_quantity(da))
+        faces = _generate_face_indices(len(lat), len(lon))
+
+    elif "x" in da.coords and "y" in da.coords:
+        if isinstance(geometry, SphericalShellGeometry):
+            planet_radius = geometry.planet_radius
+
+        x = np.rad2deg(ureg.Quantity(da.x.values, coord_units) / planet_radius).m
+        y = np.rad2deg(ureg.Quantity(da.y.values, coord_units) / planet_radius).m
```
leroyvn, line 84: That's odd, why not use `m_as(ureg.deg)`?

src/eradiate/scenes/surface/_dem.py
```diff
@@ -1,50 +1,320 @@
...
 from ...units import unit_registry as ureg
 
 
-@parse_docs
-@attrs.define(eq=False, slots=False)
-class DEMSurface(Surface):
+def mesh_from_dem(
+    da: xr.DataArray,
+    geometry: SceneGeometry,
+    planet_radius: pint.Quantity | float = EARTH_RADIUS,
```
leroyvn, line 31: Let's do something safer (and reject plain floats):
```suggestion
    planet_radius: pint.Quantity | None = None,
```

src/eradiate/scenes/surface/_dem.py
```diff
@@ -1,50 +1,320 @@
...
+    mesh : :class:`.BufferMeshShape`
+        A triangulated mesh representing the DEM
+
+    theta_lim : tuple
+        The limits of the latitude extent of the DEM
+
+    phi_lim : tuple
+        The limits of the longitude extent of the DEM
+    """
+
```
leroyvn, line 71: Safer behaviour (you don't want to accidentally mutate `EARTH_RADIUS`):

```suggestion
    if isinstance(geometry, PlaneParallelGeometry):
        if planet_radius is None:
            planet_radius = EARTH_RADIUS

    else:  # if isinstance(geometry, SphericalShellGeometry)
        if planet_radius is not None:
            # Warn user that the planet_radius arg will be ignored
```

src/eradiate/scenes/surface/_dem.py
```diff
@@ -1,50 +1,320 @@
...
+        vertices_new = [trafo.dot(vertex) for vertex in vertices_spherical]
+
+        mesh = BufferMeshShape(vertices=vertices_new, faces=faces)
+    elif isinstance(geometry, PlaneParallelGeometry):
+        vertices_new = _transform_vertices_plane_parallel(
+            vertices,
+            planet_radius=EARTH_RADIUS,
+            altitude=atmo_bottom,
+        )
+        mesh = BufferMeshShape(vertices=vertices_new, faces=faces)
```
leroyvn, line 112: Please add an `else` for safety.

src/eradiate/scenes/surface/_dem.py
```diff
@@ -1,50 +1,320 @@
...
+
+    return mesh, theta_lim, phi_lim
+
+
+def _generate_dem_vertices(x, y, elevation):
+    """
+    Generate DEM vertex positions as (latitude, longitude, elevation) pairs
+    for a plane parallel atmosphere geometry.
+
+    Note
```
leroyvn, line 122: [Numpydoc ref](https://numpydoc.readthedocs.io/en/latest/format.html#notes):
* It's "Notes"
* It's after "Parameters"

And I'd rather see this as a warning, it's critical.

src/eradiate/scenes/surface/_dem.py
```diff
@@ -1,50 +1,320 @@
...
+    Parameters
+    ----------
+    latitude_points : array-like
+        Points along the latitude axis of the DEM
+
+    longitude_points : array-like
+        Points along the longitude axis of the DEM
+
+    elevation : quantity
+        Elevation data in units of length.
```
leroyvn, line 135: As always: units everywhere or nowhere, for consistency, please. In this case, it's safe IMO to just manipulate plain arrays and let the caller handle units.

src/eradiate/scenes/surface/_dem.py
```diff
@@ -1,50 +1,320 @@
...
+
+    Returns
+    -------
+    vertices : array
+        DEM vertices as (lat, lon, elevation) triplets. Atmospheric geometry
+        is not taken into account at this step
```
leroyvn, line 141: The atmosphere has nothing to do with this part of the code, so let's not confuse people with it:
```suggestion
        DEM vertices as (lat, lon, elevation) triplets.
```

src/eradiate/scenes/surface/_dem.py
```diff
@@ -1,50 +1,320 @@
...
+        )
+        mesh = BufferMeshShape(vertices=vertices_new, faces=faces)
+
+    return mesh, theta_lim, phi_lim
+
+
+def _generate_dem_vertices(x, y, elevation):
+    """
+    Generate DEM vertex positions as (latitude, longitude, elevation) pairs
+    for a plane parallel atmosphere geometry.
```
leroyvn, line 120:
```suggestion
    for a plane parallel geometry.
```

src/eradiate/scenes/surface/_dem.py
```diff
@@ -1,50 +1,320 @@
...
+            altitude=atmo_bottom,
+        )
+        mesh = BufferMeshShape(vertices=vertices_new, faces=faces)
+
+    return mesh, theta_lim, phi_lim
+
+
+def _generate_dem_vertices(x, y, elevation):
+    """
+    Generate DEM vertex positions as (latitude, longitude, elevation) pairs
```
leroyvn, line 119:
```suggestion
    Generate DEM vertex positions as (latitude, longitude, elevation) triplets
```

src/eradiate/scenes/surface/_dem.py
```diff
@@ -1,50 +1,320 @@
...
+    return x * len_y + y
+
+
+def _generate_face_indices(len_x, len_y):
+    """
+    Generate indices for face definition in a mesh with gridded vertex positions
+
+    Parameters
+    ----------
+
```
leroyvn, line 180:
```suggestion
```

src/eradiate/scenes/surface/_dem.py
```diff
@@ -1,50 +1,320 @@
...
+    len_x : float
+        Length of the row dimension of the vertex grid
+
+    len_y : float
+        Length of the column dimension of the vertex grid
+
+    Returns
+    -------
+
```
leroyvn, line 189:
```suggestion
```

src/eradiate/scenes/surface/_dem.py
```diff
@@ -1,50 +1,320 @@
...
+    return face_indices
+
+
+def _transform_vertices_spherical_shell(vertices, planet_radius):
+    """
+    Convert the (lat, lon, elevation) vertices from the initial vertex generation
+    into (x, y, z) values for spherical shell atmosphere geometries.
+    """
+    planet_radius_m = planet_radius.m_as(ucc.get("length"))
```
leroyvn, line 219: You should require units for the `vertices` argument (and add typing annotations for clarity). As always: units everywhere or nowhere. Then, you can operate safely on
```python
vertices_m = vertices.m_as(length_units)
```

src/eradiate/scenes/surface/_dem.py
```diff
@@ -1,50 +1,320 @@
...
+    face_indices_2 = np.array((vertex_se, vertex_sw, vertex_ne)).transpose()
+    face_indices = np.concatenate((face_indices_1, face_indices_2))
+
+    return face_indices
+
+
+def _transform_vertices_spherical_shell(vertices, planet_radius):
+    """
+    Convert the (lat, lon, elevation) vertices from the initial vertex generation
+    into (x, y, z) values for spherical shell atmosphere geometries.
```
leroyvn, line 217:
```suggestion
    into (x, y, z) values for spherical shell geometries.
```

src/eradiate/scenes/surface/_dem.py
```diff
@@ -1,50 +1,320 @@
...
+    in the vertex list.
+    """
+    theta_min = np.min(vertices[:, 0])
+    theta_max = np.max(vertices[:, 0])
+    phi_min = np.min(vertices[:, 1])
+    phi_max = np.max(vertices[:, 1])
+    return (theta_min, theta_max), (phi_min, phi_max)
+
+
+def _transform_vertices_plane_parallel(vertices, planet_radius, altitude):
```
leroyvn, line 254: Maybe group similar functions together?

src/eradiate/scenes/surface/_dem.py
```diff
@@ -1,50 +1,320 @@
...
+    """
+    Convert the (lat, lon, elevation) vertices from the initial vertex generation
+    into (x, y, z) values for plane parallel atmosphere geometries.
+
+    Parameters
+    ----------
+
+    vertices : list
+        List of mesh vertices in (lat, lon, elevation) tuples.
```
leroyvn, line 263: I think this type is inconsistent with the rest of your code. And the usual comment about units: everywhere or nowhere.

src/eradiate/scenes/surface/_dem.py
```diff
@@ -1,50 +1,320 @@
...
+
+    x = phi_local * planet_radius.m_as(ucc.get("length"))
+    y = theta_local * planet_radius.m_as(ucc.get("length"))
+    z = vertices[:, 2] + altitude.m_as(ucc.get("length"))
+
+    vertices_new = np.array((x, y, z)).transpose()
+    return vertices_new
+
+
+def _to_uv(lat_min, lat_max, lon_min, lon_max) -> mi.ScalarTransform4f:
```
leroyvn, line 283: Please add quotes here, otherwise Sphinx won't let it through.
```suggestion
def _to_uv(lat_min, lat_max, lon_min, lon_max) -> "mi.ScalarTransform4f":
```

src/eradiate/scenes/surface/_dem.py
```diff
@@ -55,7 +325,7 @@ class DEMSurface(Surface):
         doc=get_doc(Surface, "id", "doc"),
         type=get_doc(Surface, "id", "type"),
         init_type=get_doc(Surface, "id", "init_type"),
-        default='"surface"',
+        default='"terrain"',
```
leroyvn, line 328: Isn't `surface` okay?

src/eradiate/scenes/surface/_dem.py
```diff
@@ -285,22 +373,131 @@ def _bsdf_id(self):
...
+    @classmethod
+    def from_mesh(
+        cls,
+        id,
+        mesh,
+        lat,
+        lon,
+        geometry=PlaneParallelGeometry(),
+        planet_radius=EARTH_RADIUS,
+        bsdf=LambertianBSDF(),
```
leroyvn, line 412: Please add type annotations.

Please don't use mutables as defaults.

leroyvn (changes requested):
Hi @schunkes, we're getting there on the design side. There are a lot of details to fix. I've been specific, but I probably missed some. Please:

* Check defaults and type annotations. There are many missing or inconsistent entries. You also sometimes use mutables as defaults, which is very dangerous.
* Check unit requirement consistency in all your functions. To keep it simple and safe, either all args should have units, or none (and then you might have to document arg units).
* Please compile and check the documentation.

src/eradiate/scenes/shapes/_buffermesh.py
```diff
@@ -27,13 +26,12 @@ class BufferMeshShape(ShapeInstance):
...
         pinttr.field(
-            validator=pinttr.validators.has_compatible_units,
-            units=ucc.deferred("length"),
+            converter=np.array,
             kw_only=True,
         ),
         doc="List of vertex positions. The passed list must contain a (n, 3) list"
-        "of three dimensional points.\n\nUnit-enabled field (default: ucc['length']).",
-        type="quantity",
+        "of three dimensional points. Vertex positions are assumed to be given in kernel units.",
```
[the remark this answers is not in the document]
schunkes: I don't know, to be honest. Maybe this was changed somewhere in the past and when I rebased my branch, I picked the wrong code.

src/eradiate/scenes/surface/_dem.py
schunkes: Because the result of the division does not have `radians` as a unit. After dividing the two quantities, the result will be `unitless`.

src/eradiate/scenes/surface/_dem.py
schunkes: I am not quite sure, why you would accidentally mutate EARTH_RADIUS in the other case :thinking:

schunkes:
All things we discussed are implemented. I think this can be reviewed again.

leroyvn:
I applied a round of fixes. Main changes are:

* Fixed and updated documentation (docstrings and Sphinx docs).
* Improved conversion protocols (I added a new `to_mi_scalar_transform()`converter for convenience).
* Exposed missing members.
* Fixed mutable defaults.
* Converted `dict()` to literals for consistency.
* Refactored tests a bit to reduce boilerplate.

schunkes:
Looks good :)

PR #328, opened 2023-05-10
Create an Eradiate package [1/3]

wint3ria:
# Description

This "minimal" (not minimalist) PR aims at providing a python wheel for eradiate to end users [through the public Pypi registry](https://pypi.org/project/eradiate/). This high level objective requires [a few changes in the Mitsuba fork](https://github.com/eradiate/mitsuba3/pull/2#issue-1695956485) of the Eradiate project, the Eradiate codebase itself, the documentation of both projects, and the development and installation process of Eradiate. As for Mitsuba, this work is divided into several PRs. This is the first one, focusing on the main features, the second one (#330) will focus on dependencies handling, and the last one will focus on the documentation and testing of the feature.

The package should be installable using the following command:

```
pip install eradiate[production]
```

The usage of the `[production]` suffix is mandatory for now, as discussed in #330 .

## Move the Misuba plugins and their tests to the Mitsuba repo

See https://github.com/eradiate/mitsuba3/pull/2#issue-1695956485

Removed plugins and their tests, renamed tests/02_eradiate to tests/01_eradiate

Removed the deprecated `resources/generate_stub_files.py`script

## Remove the CMake based build system, compile Mitsuba in-tree

See https://github.com/eradiate/mitsuba3/pull/2#issue-1695956485

Since there is no need to compile anything anymore outside of the Mitsuba submodule, All the CMake build system configuration have been removed. 

An additional Makefile `kernel` rule has been written to recompile the kernel in the submodule:
```
make kernel
```
This internally calls CMake on ext/mitsuba/build and then ninja

## Build an Eradiate wheel

Another Makefile rule has been written for the wheel: 

```
make wheel
```
It outputs the wheels in a dist folder. These can directly be installed, or uploaded to Pypi.

**Caution:** This will work for non [PEP440](https://peps.python.org/pep-0440/) Eradiate versions. Due to the usage of setuptools-scm in Eradiate, development (dirty) versions of Eradiate may have non [PEP440](https://peps.python.org/pep-0440/) versions. It means that only properly tagged versions of Eradiate may be uploaded to Pypi, but the package should build in any case. Since there is no good reason to publish a non tagged version, I believe this is not a bug, but a feature.

## CD

A github action has been written to build the package and upload it to Pypi conveniently.

## Dependency management

Two [additional dependency layers](https://eradiate.readthedocs.io/en/stable/rst/maintainer_guide.html#basic-principles) have been added to the project that add the `eradiate_mitsuba` package as a dependency:
- `optional`
- `production`

`production` extends the `recommended` layer and `optional` extends the `dev` layer:

```
    |------ tests <---- dev <---- optional*
    |                    |
    v                    |
 main <---- docs <-------|
    ^                    |
    |                    v
    |---------- recommended <---- production*
```

## Detecting the Eradiate setup

Upon importing Eradiate, the module will detect if the `ERADIATE_SOURCE_DIR` variable is set. Eradiate imports and CLI operations do not fail anymore if `ERADIATE_SOURCE_DIR` is not set, but their behavior change accordingly. Basically Eradiate will assume that the current setup is a production environment if and only if `ERADIATE_SOURCE_DIR` is not set. In this case the config class `source_dir` member will be set to None.

## Data files handling

The main problem here was to serve the files under the small_files data store in production. In development mode, this store has not been touched.

The small_files data store has been changed to use a safe online data store only in production, online mode. Its configuration has been added to `eradiate.config`

Until the small_files find their place on a new place where they can easily be in sync with Github, I used Github raw as a registry. This means that in production, eradiate data fetch, or any small data file cache miss will trigger a download directly from Github. This has some advantages, but is far from being even acceptable:
 - pros: it is simple, and using a revision configuration variable (`ERADIATE_SMALL_FILES_REGISTRY_REVISION`), we can even target a specific commit if required.
 - cons: it's hacky, ugly, and not the best manners towards our Github host to let our users fetch their data this way.
It's a temporary solution until we decide on a better hosting solution for the small files, which should not take long. Changing this is simply a matter of configuration (i.e. updating `ERADIATE_SMALL_FILES_REGISTRY_URL`). Until then, both the development and production workflows can be supported.

If in an offline, production setup, the small files will not be available through `eradiate data fetch`.

## Eradiate CLI changes

- The `data fetch` command will adapt its behavior to a prod/dev environment and source its download list from an appropriate config file in both cases. As a results there are now 2 config files to set the list of downloads. I wonder if it wouldn't be a better idea to simply drop the download.yml files altogether and let the data store infer what it's able to download on its own?
- The `eradiate show` command display meaningful information on the current setup

## Testing

- Some basic functionalities are tested inside the pytest suite

Adding additional tests to the pytest suite and adding CI steps to test in different environment should be done the subsequent PR(s) for this task, since it requires to have the packaging branch of https://github.com/eradiate/mitsuba3 targeted as a submodule.

# Checklist

- [x] The feature branch is rebased on the current state of the ~`main`~ `packaging` branch


wint3ria:
run Eradiate CI

wint3ria:
... It seems that I can not run the GH actions on the PR?

wint3ria:
Here is a first functional version, ready for review

src/eradiate/data/_store.py
```diff
@@ -13,7 +13,7 @@
 data_store: MultiDataStore = None
 
 
-def init_data_store(offline: bool | None = None) -> None:
+def init_data_store(offline: bool | None = None, production: bool | None = None) -> None:
```
leroyvn, line 16: Docstring for this new option is missing. Side comment: the end goal is to generalize the data store setup with a config file-based system and provide defaults for dev and production setups, while still allowing the user to override it.

src/eradiate/_config.py
```diff
@@ -8,6 +8,7 @@
 import enum
 import os.path
 import pathlib
+from importlib.metadata import version, PackageNotFoundError
```
leroyvn, line 11: How about factoring the Mitsuba package import and version detection code out in the `kernel` subpackage? There is already a bit of logic here, we could have a small interface like:
```python
eradiate.kernel.FOUND_DRJIT_MODULE  # True iff drjit module can be imported
eradiate.kernel.FOUND_MITSUBA_MODULE  # True iff mitsuba module can be imported
eradiate.kernel.FOUND_ERADIATE_MITSUBA_PACKAGE  # True iff eradiate-mitsuba package can be imported
eradiate.kernel.ERADIATE_MITSUBA_VERSION  # String with kernel version number (raises if wrong Mitsuba is used)
# All variables set upon `import eradiate.kernel`; failures raise exceptions.
```

leroyvn (changes requested):
Hi @wint3ria, this looks good. I have a few small comments, nothing critical.

docs/rst/getting_started/development.rst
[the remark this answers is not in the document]
wint3ria: Sounds like a good idea, giving fewer and clearer entrypoints to the documentation

src/eradiate/_config.py
wint3ria: Unless I am missing smth, I don't think we can opt out of having a mitsuba or drjit modules available in the current working environment. I don't really see the point of using an extra variable to tell if I found the module, if I raise when I can not find it anyway. Drjit and mitsuba availabilities are already tested in `kernel.__init__.py`. 

However, checking for the existence of the eradiate-mitsuba package, its version, the kernel version and the eradiate patch version can be outsourced in the `eradiate.kernel` module

wint3ria:
Build finished [here](https://github.com/wint3ria/mitsuba3/actions/runs/5254018177/jobs/9494931285)

leroyvn:
Let's go! :rocket:

PR #344, opened 2023-06-20
Custom atmospheric profiles + absorption upgrade

nollety:
# Summary

Brings support for customized atmospheric profiles as well as custom absorption datasets.

:warning: Do not forget to merge `eradiate-data:thermoprops-next` (regression test reference update)
:warning: Do not forget to merge `eradiate-tutorials:thermoprops-next` (tutorials that use molecular atmospheres have been updated)

# Description

## Changes

See the `CHANGELOG.md` file for any change that I did not documented here.

### Atmosphere thermophysical profiles

The package `src/eradiate/thermoprops` has been removed. The handling of atmospheric profiles is now provided by [the `joseki` package](https://github.com/nollety/joseki) that implements all functionalities that existed in `src/eradiate/thermoprops` (essentially altitude interpolation and rescaling of mole fraction data), and more. For more information, refer to the [Joseki documentation](https://nollety.github.io/joseki).

As an example, the U.S. Standard profile from the AFGL (1986) series is produced like so:
```python
import joseki

thermoprops = joseki.make(identifier="afgl_1986-us_standard")
```

### `MolecularAtmosphere`

This class represents a 1D "clear-sky" atmosphere (i.e. no aerosols, no clouds, just air). Its two most important parameters are the atmosphere's thermophysical properties (`thermoprops`) and absorption dataset(s) (`absorption_data`).

### `AtmosphereRadProfile`

This class supersedes both `AFGL1986RadProfile` and `US76ApproxRadProfile`.

:warning: Note that I did not make use of the deprecation system to warn for classes about to be deprecated; these classes are simply not available anymore (`src/eradiate/radprops/_us76_approx.py` and `src/eradiate/radprops/afgl1986.py` are removed.)

### Absorption data

A new data set format has been produced and is documented under _User Guide / Data Guide / Absorption_.

The following absorption data (compliant with this new format) has been uploading to the Eradiate online (stable) store:

| Dataset codename | path | specifications | 
| :--- | :---| :--- |
| `gecko` | `spectra/absorption/mono/gecko/gecko.nc` | [gecko/metadata.json](http://eradiate.eu/data/store/stable/spectra/absorption/mono/gecko/metadata.json) |
| `komodo` | `spectra/absorption/mono/komodo/komodo.nc` | [komodo/metadata.json](http://eradiate.eu/data/store/stable/spectra/absorption/mono/komodo/metadata.json) |
| `monotropa` | `spectra/absorption/ckd/monotropa/*.nc` | [monotropa/metadata.json](http://eradiate.eu/data/store/stable/spectra/absorption/ckd/monotropa/metadata.json) |
| `panellus` | `spectra/absorption/ckd/panellus/*.nc` | [panellus/metadata.json](http://eradiate.eu/data/store/stable/spectra/absorption/ckd/panellus/metadata.json) |

`komodo` `[~220 MB]` tabulates the air volume absorption coefficient against the mole fraction of H2O, CO2 and O3 within the wavelength range [250, 3125] nm (at a relatively low resolution, 1cm^-1, for distribution purposes), and is chosen to serve as default absorption data in monochromatic mode.

`monotropa` datasets (sizes varying depending on the spectral band, see below) are tabulated with respect to 1 to 10 of HITRAN's first 10 molecules' (H2O, CO2, O3, N2O, CO, CH4, O2, NO, SO2, NO2) mole fractions, in 100 cm^-1 wide wavenumber ranges from 3200 to 40000 cm^-1 (namely 250 to 3125 nm), and is chosen to serve as default absorption data in CKD mode.

Note: with `komodo`, it is now possible to evaluate the absorption coefficient of an _AFGL (1986)_ profile in monochromatic mode, and removes the need to rely on the _U.S. Standard Atmosphere 1976_ profile in this mode. As a result, the entire codebase (and in particular the test suite) was refactored so that it uses the _U.S. Standard_ profile in places where the _U.S. Standard Atmosphere 1976_ atmosphere was used. The end results being that the _U.S. Standard_ profile is now the default profile in both spectral modes, monochromatic and CKD.

#### `monotropa` file size histogram

Total number of datasets: 368

monotropa_file_size

#### `panellus` file size histogram

Total number of datasets: 2875

panellus_file_size

## Notes

* [Work in progress on Joseki to accomodate CAMS reanalysis datasets](https://github.com/nollety/joseki/pull/348) ; this will only change the actual data fed into Eradiate, the support for [_Joseki's_ format](https://nollety.github.io/joseki/explanation/#dataset-schema) is already implemented in _Eradiate_ (see `src/eradiate/radprops/absorption.py` in particular).
* ~A bunch of tests are marked as slow because they rely on a rather large monochromatic absorption dataset (takes time to open). A low- spectral resolution low-accuracy monochromatic absorption dataset is being prepared as a replacement.~ EDIT: the slowyness of these tests did not originate in the process of opening them, but rather in the process of creating a list of spectral indices from the corresponding dataset' wavlengthts/CKD bins. Creating a list was not necessary within the scope of these tests and this has been refactored. These tests are not slow anymore.

## TODO

- [x] cleanup pipelines
- [x] `Experiment.context_init()` is hacky; refactor
- [x] `Experiment._contexts()` is unfinished work: finish
- [x] ~~shorten `BinSet` repr~~
- [x] invenstigate failure of system test: `tests/02_eradiate/02_system/test_heterogeneous_atmosphere_expansion.py`
  - [x] Double check how z-grids are managed from `ParticleLayer`/`MolecularAtmosphere` to `HeterogeneousAtmosphere`
- [x] ~~investigate failure of regression test `tests/02_eradiate/03_regression/rami4atm/test_rami4atm_hom00_bla_sd2s_m03_z30a000_brfpp.py`~~
- [x] ~~derive new tests from former `test_us76_approx.py` and `test_afgl_1986.py`~~
- [x] ~~repr of `SceneGeometry` is annoyingly long~~
- [x] ~~`MolecularAtmosphere.thermoprops`: levels must be regularly spaced -> address that in `_convert_thermoprops`~~
- [x] ~~Restore default constructor for `MolecularAtmosphere()`?~~
  - [x] ~~define a default monochromatic dataset~~
  - [x] define a default ckd dataset
  - [x] ~~define a function that takes a spectral interval in and, based on active mode, outputs the default absorption dataset in the corresponding spectral interval~~ 
- [x] Inconsistencies in eval_* method' output: some return coefficient array with shape `(1, n_layers)`, some with shape `(n_layers,)`

### Tests

- [x] Fix/update tests
- [x] ~~Create new regression tests?~~

### Documentation

- [x] Describe the new CKD datasets format
- [x] ~~Describe new CKD databases that Eradiate ships (generate `.rst` file from `metadata.json` file?)~~
- [x] Fix/update documentation

### Tutorials

- [x] Update existing tutorials
- [x] ~~Add a tutorial about how to integrate user absorption data (monochromatic / absorption)~~

# Checklist

- [x] ~~The code generates no new warnings~~ The code warns on missing coordinate and out of bounds interpolation errors during the interpolation of absorption coefficient data. It is possible to configure Eradiate to silent these warnings.
- [x] The feature branch is rebased on the current state of the ~~`main`~~ `next` branch


nollety:
This failing system will be ignored:
`tests/02_eradiate/02_system/test_basic.py::test_radiometric_accuracy[mono_single-1.0-constant-500000.0]`

This failing test should be investigated:
`tests/02_eradiate/02_system/test_heterogeneous_atmosphere_expansion.py`

```
E           AssertionError: 
E           Arrays are not equal
E           
E           Mismatched elements: 12 / 12 (100%)
E           Max absolute difference: 0.01362815
E           Max relative difference: 0.04407644
E            x: array([[[[[0.295566, 0.297883, 0.306182, 0.306511, 0.31239 , 0.306885,
E                      0.309647, 0.314504, 0.312265, 0.303837, 0.313195, 0.297604]]]]])
E            y: array([[[[[0.309194, 0.302339, 0.308939, 0.309404, 0.311354, 0.31372 ,
E                      0.31498 , 0.317593, 0.311461, 0.310758, 0.303773, 0.297825]]]]])
```

This failing regression test should be investigated:
`tests/02_eradiate/03_regression/rami4atm/test_rami4atm_hom00_bla_sd2s_m03_z30a000_brfpp.py`

nollety:
Found the issue with `test_heterogeneous_atmosphere_expansion.py`'s failure: the geometry was not being propagated correctly from the experiment (`AtmosphereExperiment`) down to each atmosphere's components (`ParticleLayer`).

I added a call to `self.atmosphere.update()` in the body of all experiments' `_normalize_atmosphere(self)` method, right after the atmosphere geometry is assigned to the experiment geometry. In the case where the atmosphere is heterogeneous, this will make sure that the `HeterogeneousAtmosphere` updates also its components' geometries.

nollety:
That had the side effect to break the regression test `test_rpv_afgl1986_continental_brfpp `

nollety:
> That had the side effect to break the regression test `test_rpv_afgl1986_continental_brfpp `

Fixed

nollety:
Attaching here the regression output figure:
image
@leroyvn, how trusted are the reference results for this regression test? Was the spectral update traversal properly working when the reference results were computed? From what I can see, the reference results were last modified on June 7th, 2022.

leroyvn:
> how trusted are the reference results for this regression test? Was the spectral update traversal properly working when the reference results were computed? From what I can see, the reference results were last modified on June 7th, 2022.

This was done prior to the traversal upgrade. The reference is compared vs LibRadtran and RTMOM, so I think it's rather solid. There is noise, maybe indeed updating the ref with a higher sample count would be good.

To be honest, I don't expect 0.01% reproducibility. This test is designed to detect big issues, but you won't spot more minor differences. What you are showing looks good IMO.

nollety:
The results you are looking at still fail the chi2 test ; changing the spp does not make it pass.

nollety:
I've started rebasing on main (84 commits to catch up)

src/eradiate/radprops/absorption.py
```diff
@@ -3,163 +3,750 @@
...
+        for c in out_of_bounds:
+            msg += f"\n{c}: range in thermoprops {out_of_bounds[c]['thermoprops']}, "
+            msg += f"range in absorption dataset {out_of_bounds[c]['absorption']}"
+        raise OutOfBoundsCoordinateError(msg)
+
+
+def interp_x(
+    k: xr.DataArray,
+    thermoprops: xr.Dataset,
+    error_handler_config: dict[str, str] = DEFAULT_X_ERROR_HANDLER_CONFIG,
```
leroyvn, line 203: Mutable default

leroyvn (commented):
Hi @nollety, here are a couple of comments after a quick scan of the `RadProfile` and `MolecularAtmosphere` classes.

src/eradiate/radprops/_atmosphere.py
```diff
@@ -0,0 +1,290 @@
...
+        type="Dataset",
+        default=":meth:`joseki.make() <joseki.make>` with "
+        "``identifier='afgl_1986-us_standard'`` and "
+        "``z=np.linspace(0, 120, 61) * ureg.km``"
+        "``additional_moleculs=False``.",
+    )
+
+    @thermoprops.validator
+    def _check_thermoprops(self, attribute, value):
+        if not value.joseki.is_valid:
```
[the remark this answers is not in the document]
nollety: Yes

src/eradiate/radprops/_atmosphere.py
```diff
@@ -0,0 +1,290 @@
...
+            default=True,
+            converter=bool,
+            validator=attrs.validators.instance_of(bool),
+        ),
+        doc="Absorption switch. If ``True``, the absorption coefficient is "
+        "computed. Else, the absorption coefficient is not computed and "
+        "instead set to zero.",
+        type="bool",
+        default="True",
+    )
```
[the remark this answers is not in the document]
nollety: Could you add a click up task so that we remember to discuss that in the future? Thanks!

nollety:
I have updated the tutorials; @leroyvn feel free to take a look and let me know what you think about the new `MolecularAtmosphere` interface.

nollety:
I've added your suggestions in the last commit @leroyvn

nollety:
@leroyvn We still need to find a solution for the `Experiment.context_init()` issue, I think. Beside that, this PR is mostly ready for a review. The sooner we start, the sooner we are done, I suppose. I'd appreciate if you could skip the kind of comments that could easily be addressed in a second pass (like style, minor refactoring issues, and so on) and focus on the important things, so that this work moves forward quickly (as we are late wrt deadlines). Sorry that the diff is so large, but everything ended up being unravelable.
Thanks!

leroyvn:
For the init context: best would be to use a generator for this. The current context list is:

* sorted: this is for convenience, nothing requires bins or quadrature points to be sorted;
* deduplicated: this is to avoid repeating computations for identical contexts already emitted by different sources.

So we want a generator which aggregates multiple generators and avoids duplicates? I think that should do:
```python
class MultiGenerator:
    """
    This generator aggregates several generators and makes sure that items that
    have already been served are not repeated.
    """
    def __init__(self, generators):
        self.generators = generators
        self._i_generator = 0
        self._current_iterator = iter(self.generators[self._i_generator])
        self._visited = set()

    def __iter__(self):
        return self

    def __next__(self):
        try:
            result = next(self._current_iterator)
            if result not in self._visited:
                self._visited.add(result)
                return result
            else:
                return self.__next__()
        except StopIteration:
            if self._i_generator >= len(self.generators) - 1:
                raise
            else:
                self._i_generator += 1
                self._current_iterator = iter(self.generators[self._i_generator])
                return self.__next__()
```
A few basic tests:
```python
assert list(MultiGenerator([range(5), range(10)])) == [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
assert list(MultiGenerator([[0, 1, 2, 3, 0, 2, 1], [0, 1, 2, 3, 0, 2, 1, 4]])) == [0, 1, 2, 3, 4]
assert list(MultiGenerator([[3, 0, 2, 1], [0, 4, 2, 0, 1, 4]])) == [3, 0, 2, 1, 4]
```
Note that this generator has the disadvantage of storing all generated values, but it should not consume a lot of memory in our case.

You can then write a method which emits an aggregate spectral index generator and call `__next__()` once to get the init context:
```python
def _spectral_index_generator(self):
    return MultiGenerator([gen_a, gen_b, ...])

def context_init(self):
    return self._spectral_index_generator().__next__()

def contexts(self):
    return sorted(self._spectral_index_generator())  # I'm omitting sorting details for conciseness
```

Note that the init context is not easily predictable, but it will always be part of the generated context list. This should avoid the annoying situation where the init context could be outside of the range covered by the sensor or the atmospheric absorption dataset.

src/eradiate/spectral/ckd.py
```diff
@@ -143,6 +139,15 @@ def spectral_indices(self) -> t.Generator[CKDSpectralIndex]:
 # ------------------------------------------------------------------------------
 
 
+DEFAULT_QUAD_SPEC = {
+    "type": "fixed",
+    "params": {
+        "type": "gauss_legendre",
+        "n": 1,
+    },
+}
```
leroyvn, line 148: I assume the `G16` variable is no longer used? (See above, I couldn't comment there).

src/eradiate/util/misc.py
```diff
@@ -465,8 +465,17 @@ def summary_repr(value):
     return repr(value)
 
 
+@summary_repr.register
+def _(datasets: list):
+    return [_summary_repr_impl_ds(ds) for ds in datasets]
```
leroyvn, line 470: This returns a `list[str]`, not a repr (which should be a `str`). I'd be more in favour of using a lambda to do this where you need it.

leroyvn (commented):
Here are a few additional comments. The most important is the one about fusing gather and aggregate pipeline steps: if possible, please split these, pipeline debugging is already difficult and bigger, more monolithic steps do not make things easier.

Overall, good job! I'm happy to see this come to production, this is a huge update!

nollety:
> For the init context: best would be to use a generator for this. The current context list is:
   sorted: this is for convenience, nothing requires bins or quadrature points to be sorted;
   deduplicated: this is to avoid repeating computations for identical contexts already emitted by different sources.
   ```python
   assert list(MultiGenerator([[3, 0, 2, 1], [0, 4, 2, 0, 1, 4]])) == [3, 0, 2, 1, 4]
   ```
Is your generator sorting the generated values then?

nollety:
> Note that this generator has the disadvantage of storing all generated values, but it should not consume a lot of memory in our case.

I noticed that computing the generated values can be costly in a monochromatic setup where the absorption data spectral range is large; we probably should keep that in mind if we go with this solution.

nollety:
> Note that the init context is not easily predictable, but it will always be part of the generated context list. This should avoid the annoying situation where the init context could be outside of the range covered by the sensor or the atmospheric absorption dataset.

Yes, this is exactly the problem that we are having now; the current `Experiment.context_init()` ignores the scene spectral features and always return the default context.

src/eradiate/scenes/geometry.py
```diff
@@ -140,6 +142,12 @@ def convert(cls, value: t.Any) -> t.Any:
 
         return value
 
+    def update_with_thermoprops(self, thermoprops: xr.Dataset):
```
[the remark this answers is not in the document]
nollety: In `MolecularAtmosphere.update()` in order to adjust the geometry's `zgrid` to the one specified with `thermoprops`.

src/eradiate/spectral/ckd.py
nollety: You assume correctly ;)

src/eradiate/util/misc.py
nollety: You are right; this function is obsolete actually

leroyvn:
> Is your generator sorting the generated values then?

It does not, and doing so would be a bit complicated. I don't think spectral indices have to be sorted upon generation, then can be upon context list creation. The only difference is that with such a solution, we cannot guarantee that the init context will have the first spectral index.

To be clear about performance impact: this does not make context generation faster than it is now, it simply avoids a full double calculation to generate the init context.

src/eradiate/pipelines/_gather.py
```diff
@@ -134,6 +127,121 @@ def transform(self, x: dict) -> xr.Dataset:
...
+                dataset = dataset.rename({"img": name})
+                dataset[name].attrs.update(metadata)
+
+                _datasets.append(dataset)
+                ix += 1
+
+            # concatenate along 'g'
+            ds = xr.concat(_datasets, dim="g")
+
+            # compute quadrature
```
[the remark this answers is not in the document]
nollety: I think that the reason was that I could not comply with the function signature:
```python
def transform(self, x: dict) -> xr.Dataset:
```
as I would make transform return a list of Dataset.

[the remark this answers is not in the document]
nollety: I'll add a task to address that in our clickup space

nollety:
I've implemented your suggestions and cleanup up a little @leroyvn

src/eradiate/pipelines/_aggregate.py
```diff
@@ -137,46 +128,43 @@ def transform(self, x: t.Any) -> t.Any:
...
+                # ),
+                # "bin_wmax": (
+                #     "w",
+                #     bin_wmaxs,
+                #     {
+                #         "standard_name": "bin_wmax",
+                #         "long_name": "spectral bin upper bound",
+                #         "units": symbol(wunit),
+                #     },
+                # ),
```
leroyvn, line 162: If it's not used, I guess we can remove it:
```suggestion
```

src/eradiate/pipelines/_gather.py
```diff
@@ -134,6 +127,117 @@ def transform(self, x: dict) -> xr.Dataset:
         result = result.rename({"img": var})
         result[var].attrs.update(var_metadata)
 
-        logger.debug("gather: end")
```
leroyvn, line 137: Comment: this convinces that we need something to automate such logging operations, it is very useful for debugging.

src/eradiate/util/misc.py
```diff
@@ -467,6 +467,10 @@ def summary_repr(value):
 
 @summary_repr.register
 def _(ds: xr.Dataset):
+    return _summary_repr_impl_ds(ds)
+
+
+def _summary_repr_impl_ds(ds):
```
leroyvn, line 473: I guess now you can also remove the intermediate?
```suggestion
```

src/eradiate/radprops/absorption.py
```diff
@@ -3,163 +3,802 @@
...
+        self.action = action
+    def __call__(self, e: InterpolationError):
+        if self.action == "raise":
+            raise e
+        elif self.action == "warn":
+            warnings.warn(str(e), UserWarning)
+        elif self.action == "ignore":
+            pass
+        else:
+            raise NotImplementedError(f"Action {self.action} not implemented.")
```
leroyvn, line 75: I think you don't need this anymore.

leroyvn (approved):
Only very minor comments are left, we're almost good to go! Thanks!

PR #354, opened 2023-07-17
Change the data folder path to src/data in the CLI

wint3ria:
# Description

It seems a path was incorrect in the CLI for data fetch. It may come from a mistake I made when rewriting the CLI.
This PR fixes this problem.

Step to reproduce: in a clean installation of Eradiate, perform the command "eradiate data fetch"

# Checklist


leroyvn:
I was wondering where these files had gone; thanks for fixing that.

PR #355, opened 2023-07-17
Add a RTLS class to Eradiate

wint3ria:
# Description

Add an adapter class for the RTLS BSDF exposed by the eradiate_mitsuba kernel

# Checklist

- [x] The feature branch is rebased on the current state of the ~`main`~ `next` branch


wint3ria:
I'm not a 100% sure about the update mechanism implementation, especially for scalar parameters h, r and b.

wint3ria:
It seems that I was misunderstanding how the update mechanism works. If I'm correct, the `params` method only treats parameters that may be updated during the lifecycle of the mitsuba scene. It is not the case for the `h`, `r` and `b` hyperparameters of the RTLS model. These should stay the same for any spectral band for a given set of `f_iso`, `f_vol` and `f_geo` parameters. As a result the current implementation of `params` should be correct

CHANGELOG.md
```diff
@@ -48,6 +48,7 @@ Entries marked with a ︎⚠️ symbol require particular attention during upgra
 * Added support for custom atmospheric profiles ({ghpr}`344`).
 * Added module `radprops/absorption` to compute the air volume absorption
   coefficient from absorption datasets (monochromatic or CKD) ({ghpr}`344`).
+* Added the RTLS BSDF class to interface with the corresponding kernel plugin.
```
leroyvn, line 51: Let's add a reference to this PR here:
```suggestion
* Added the RTLS BSDF class to interface with the corresponding kernel plugin
  ({ghpr}`355`).
```

src/eradiate/scenes/bsdfs/_rtls.py
```diff
@@ -0,0 +1,184 @@
...
+    --------
+    :ref:`plugin-bsdf-rtls`
+
+    Implements the Ross-Thick, Li-Sparse (RTLS) BRDF
+    :cite:`BU-MODISBRDFAlbedoProductATBD1999`.
+
+    The RTLS class exposes the 3 parameters :math:`f_{iso}`, :math:`f_{vol}` and
+    :math:`f_{geo}` configuring the linear combination the wreights of the
+    isometric, volumetric and geometric kernel values in the BSDF definition
+    respectively.
```
leroyvn, line 34: Suggestion to better comply with the Numpydoc style and remove redundant information:
```suggestion
    RTLS BSDF [``rtls``].

    This class implements the RossThick-LiSparse (RTLS) BRDF as described
    by the MODIS BRDF/Albedo Product ATBD :cite:`BU-MODISBRDFAlbedoProductATBD1999`.

    See Also
    --------
    :ref:`plugin-bsdf-rtls`
```
Also, the other parameters (h, r, b) are also configurable; this

src/eradiate/scenes/bsdfs/_rtls.py
```diff
@@ -0,0 +1,184 @@
...
+        attrs.field(
+            default=0.209741,
+            converter=spectrum_factory.converter("dimensionless"),
+            validator=[
+                attrs.validators.instance_of(Spectrum),
+                validators.has_quantity("dimensionless"),
+            ],
+        ),
+        doc="Isotropic scattering kernel parameter. "
+        r"Defaults to the Mitsuba plugin default (equal to :math:`f_{iso}`).",
```
leroyvn, line 47: Am I understand this correctly? From what I read here, the default is simply 0.209741, not the Mitsuba plugin default.

src/eradiate/scenes/bsdfs/_rtls.py
```diff
@@ -0,0 +1,184 @@
...
+            default=0.004140,
+            converter=spectrum_factory.converter("dimensionless"),
+            validator=[
+                attrs.validators.instance_of(Spectrum),
+                validators.has_quantity("dimensionless"),
+            ],
+        ),
+        doc="Volumetric scattering from horizontally homogeneous leaf canopies "
+        r"kernel parameter. Defaults to the Mitsuba plugin default "
+        r"(equal to :math:`f_{vol}`).",
```
leroyvn, line 64: Same as above.

src/eradiate/scenes/bsdfs/_rtls.py
```diff
@@ -0,0 +1,184 @@
...
+            ),
+            validator=attrs.validators.optional(
+                [
+                    attrs.validators.instance_of(Spectrum),
+                    validators.has_quantity("dimensionless"),
+                ]
+            ),
+        ),
+        doc="Geometric-optical surface scattering kernel parameter. "
+        r"Defaults to the Mitsuba plugin default (equal to :math:`f_{geo}`).",
```
leroyvn, line 84: Same as above

leroyvn (commented):
Hi @wint3ria, I have minor comments (on documentation only). Apart from that, looks good to me!

src/eradiate/scenes/bsdfs/_rtls.py
wint3ria: 0.209741 [is the plugin default](https://github.com/eradiate/mitsuba3/blob/373e367b2c5720c9d4f3a8e83477b4cbc838cc2d/src/eradiate_plugins/bsdfs/rtls.cpp#L23). I thought implementing conditionals here would be more tedious than simply copying a value that end user will override anyway. I can look into something nicer

leroyvn: I'm fine using a hard-coded default as well. So you have two possible solutions: either change the behaviour (default to `None` and specify the plugin init value and parameters only if a non-default value is passed) or change the docs to mention explicitly the defualt value.

wint3ria: Let me know what you think

leroyvn: Looks good to me now :+1:

PR #381, opened 2024-01-10
To world shape

schunkes:
# Description

This PR implements the `to_world` member for the `Shape` class, to make in available in all child classes.
Since the `Buffermesh` does not support transformations, it will raise a warning, if the transform is set. A separate PR will add a tutorial that explains how to use the helper function, which is also added here, to manually transform the vertices for that buffermesh.

# Checklist


tutorials
leroyvn, line 1: The tutorial submodule is updated, but you didn't add it to the list of how-to guides: is that intentional? If so, maybe revert this until the content is ready?

leroyvn (changes requested):
Hi @schunkes, this is a welcome interface update. I have a few comments.

src/eradiate/scenes/shapes/_core.py
```diff
@@ -62,6 +64,31 @@ class Shape:
...
+    to_world: "mitsuba.ScalarTransform4f" = documented(
+        attrs.field(
+            converter=converters.to_mi_scalar_transform,
+            default=None,
+        ),
+        doc="Transform to scale, shift and rotate the shape. "
+        "In the `Rectangle` and `Cuboid` classes, this transform"
+        "will replace the parametrization through the other parameters. "
+        "In the `Sphere` class it will prepend the radius and therefore"
+        "multiply the radius.",
```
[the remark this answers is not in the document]
schunkes: I'll update the documentation for the subclasses accordingly. About the difference with the mitsuba plugins: Should I document that, when there is a difference? I am not quite sure, what to do with that.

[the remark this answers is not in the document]
leroyvn: No need to refer to the plugins, documenting the wrapper's behaviour is enough

tutorials
schunkes: It is not a tutorial, it is a how-to, and it should be added to the index of the how-to guides, no?

tests/01_eradiate/01_unit/scenes/shapes/test_sphere.py
```diff
@@ -21,6 +21,10 @@ def test_sphere_construct_kernel_dict(modes_all, kwargs, expected_reflectance):
         assert mi_wrapper.parameters["bsdf.reflectance.value"] == expected_reflectance
 
 
+def test_sphere_construct_trafo(modes_all):
```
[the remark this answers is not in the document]
schunkes: I am trying to rewrite the tests, but they fail and I don't understand why. Maybe we need to look at that together, on Monday.

tutorials
leroyvn: Yes, but it's not automatic: you have to add a reference to the notebook in the documentation. The how-to gallery for this branch can be viewed [here](https://eradiate.readthedocs.io/en/to_world_shape/tutorials/howto/index.html#tutorials-howto).

schunkes:
I realized that I will have to make more changes to the code. In the `Sphere` and `Cuboid` plugins, the `bbox` and `contains` methods cannot handle the to_world transforms.

leroyvn:
> I realized that I will have to make more changes to the code. In the `Sphere` and `Cuboid` plugins, the `bbox` and `contains` methods cannot handle the to_world transforms.

Indeed, these won't update automatically. The solution could be to implement a transform protocol for bounding boxes. You could leverage Mitsuba's transforms for that purpose. Another way would be to simply remove our bounding box abstraction and only rely on Mitsuba's (not sure about all the implications, to be checked). We can discuss this when you want.

schunkes:
I am confused about this change in the release notes. I am not sure how I am supposed to resolve this...

tests/01_eradiate/01_unit/scenes/shapes/test_sphere.py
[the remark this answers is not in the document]
schunkes: I updated the tests and they are more thorough now.

schunkes:
Other than the odd merge conflict, I think this is ready for review.

src/eradiate/warp.py
```diff
@@ -174,3 +174,29 @@ def uniform_disk_to_square_concentric(p: ArrayLike) -> np.ndarray:
...
+    t : mi.ScalarTransform4f
+        A transformation, to be applied to the list of points.
+
+    x : array-like
+        A list of points in (n, 3)-shape, to be transformed.
+
+    Returns
+    -------
+    ndarray
+        List of transformed points.
```
leroyvn, line 194: Can you please specify the shape of the returned value?

src/eradiate/scenes/shapes/_sphere.py
```diff
@@ -50,37 +53,20 @@ class SphereShape(ShapeNode):
...
-        return BoundingBox(self.center - offset, self.center + offset)
+        offset = np.full_like(self.center, self.radius.m_as(length_units))
+        p1 = self.center.m_as(length_units) - offset
+        p2 = self.center.m_as(length_units) + offset
+
+        if self.to_world is not None:
+            p1 = self.to_world @ p1
+            p1 = pint.Quantity(np.array(p1), length_units)
+            p2 = self.to_world @ p2
+            p2 = pint.Quantity(np.array(p2), length_units)
```
leroyvn, line 67: Not quite sure how this will work out if you apply anything else than a translation. Notably, if there is a rotation in the mix, you will rotate AABB points, while it should not happen (the Mitsuba plugin itself forbids nonuniform scalings and shears, so those will crash in a not-so-nice way but at least no weird stuff will happen).

IMO, the simplest is to recalculate the new centre and radius, and use the same formula as for the no-transform case.

Side comment: You are supposed to instantiate `Quantity` from the registry-specify type, not from the global Pint type. Please prefer the safer syntax:
```suggestion
            p1 = (np.array(self.to_world @ p1)) * length_units
            p2 = (np.array(self.to_world @ p2)) * length_units
```

leroyvn (changes requested):
Hi @schunkes, sorry for the late reply. I have another round of comments. I think the `SphereShape.contains()` method needs an update, too.

src/eradiate/scenes/shapes/_sphere.py
schunkes: I am not sure, how to compute the new radius. Mitsuba transforms don't handle scalars. The center point is clear.

schunkes:
I think I adressed all the issues, do you want to take another look?

src/eradiate/scenes/shapes/_sphere.py
```diff
@@ -29,6 +28,11 @@ class SphereShape(ShapeNode):
     Sphere shape [``sphere``].
 
     This shape represents a sphere parametrized by its centre and radius.
+
+    Notes
+    -----
+    * If the `to_world` parameter is set, it will be appended to the position
+      and scaling due to the `center` and `radius` parameters.
```
leroyvn, line 35:
```suggestion
      and scaling defined by the `center` and `radius` parameters.
```

src/eradiate/warp.py
```diff
@@ -174,3 +174,29 @@ def uniform_disk_to_square_concentric(p: ArrayLike) -> np.ndarray:
     b = np.where(quadrant_0_or_2, t, r)
 
     return np.vstack(((a + 1.0) * 0.5, (b + 1.0) * 0.5)).T
+
+
+def transform_affine(t, x):
```
leroyvn, line 179: Please move this to `eradiate.kernel.transform`, it's not a sample warping function.

leroyvn (commented):
Hi @schunkes, we're almost good to go. After those final changes, please rebase and merge to `next`. Thanks!

schunkes:
All comments are addressed, all tests pass. Merging.

PR #398, opened 2024-04-07
Add retry option to online data stores

leroyvn:
# Description

This PR adds a retry option to online data stores. It also fixes an incorrect path to the `komodo` dataset.

# Checklist

### DIFF
--- a/src/eradiate/data/_blind_online.py
+++ b/src/eradiate/data/_blind_online.py
@@ -1,7 +1,9 @@
 from __future__ import annotations
 
+import logging
 import os
 import shutil
+import time
 import typing as t
 from pathlib import Path
 
@@ -14,6 +16,8 @@ from ..attrs import documented, parse_docs
 from ..exceptions import DataError
 from ..typing import PathLike
 from ..util.misc import LoggingContext
+
+logger = logging.getLogger(__name__)
 
 
 @parse_docs
@@ -34,6 +38,13 @@ class BlindOnlineDataStore(DataStore):
         type="Path",
         init_type="path-like",
         doc="Path to the local cache location.",
+    )
+
+    attempts: int = documented(
+        attrs.field(default=3, converter=int),
+        type="int",
+        doc="Number of download attempts to make before giving up because of "
+        "connection errors.",
     )
 
     @property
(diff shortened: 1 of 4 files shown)


src/eradiate/data/_blind_online.py
```diff
@@ -36,6 +40,13 @@ class BlindOnlineDataStore(DataStore):
         doc="Path to the local cache location.",
     )
 
+    retry_if_failed: int = documented(
```
wint3ria, line 43: just call it "retry" or "attempts"

src/eradiate/data/_blind_online.py
```diff
@@ -36,6 +40,13 @@ class BlindOnlineDataStore(DataStore):
...
+        attrs.field(default=1, converter=int),
+        type="int",
+        doc="Retry a file download the specified number of times if it fails "
```
wint3ria, line 46: *if a network related exception occurs

src/eradiate/data/_blind_online.py
```diff
@@ -88,41 +99,63 @@ def fetch(self, filename: PathLike, downloader: t.Callable | None = None) -> Pat
...
+                # If we get here, it means download failed
+                retries_left = download_attempts - (i + 1)
+                logger.info(
+                    "Failed to download '%s'. "
+                    "Will attempt the download again %d more time%s.",
+                    fname,
+                    retries_left,
+                    "s" if retries_left > 1 else "",
+                )
+                time.sleep(min(i + 1, max_wait))
```
wint3ria, line 151: I recommend using a queue of downloads. We would just push the failed download at the end of the queue. A simple dict of attempt numbers can be kept locally and updated when necessary. It would only be the size of the number of failed downloads which in practice is (hopefully) not too big. This could also prepare for an async implementation of the download loop, potentially further improving performances. You could use the size of the failed download counts as a parameter for the time.sleep feature instead, avoiding to reset the delay for each new file.

src/eradiate/data/_safe_online.py
```diff
@@ -33,6 +33,10 @@ class SafeOnlineDataStore(DataStore):
     registry_fname : path-like, optional
         Path to the registry file, relative to `path`.
 
+    retry_if_failed : int, default: 1
```
wint3ria, line 36: We probably want the default value to be 3. Users can't care less about our API at that point, they would not even notice the retry feature if we keep using a retry count of 1. Advanced users who would want to optimize their IOs over multiple distributed hosts will probably automate this task of downloading the data and would like to make it run only once, trying to avoid script failures as much as possible too.

wint3ria (changes requested):
@leroyvn , I made a few comments. We can elaborate on the sleep mechanism maybe.

For the rest it's only a matter of choosing the default parameters and documentation.

src/eradiate/data/_blind_online.py
leroyvn: At this point, I don't see a practical way to implement this nicely in store components. As I understand it, the queue should be common to all files; however, store queries are currently on individual files and download tasks are greedy, so I don't think we can fit the queue in the current data store design. I think at this point, the queue system would be implemented by the data fetch program, but I'm not sure if it's really worth the effort since we plan to redesign the data store in a not-so-distant future.

leroyvn:
Hi @wint3ria, I made a few edits to account for your comments. The queue feature is a deeper change, so I think regardless of the feasibility of it in the framework of the current architecture, it's better to make it a separate PR.

leroyvn:
I'll merge now and add the queue feature to our todo list.

src/eradiate/data/_blind_online.py
wint3ria: Ah yes, I just saw the problem. Hum. The retry mechanism of the data fetch program could indeed be different from the one of the data store. I'm not entirely sure we need both. But let's keep it simple for now. The main objective is achieved

PR #404, opened 2024-04-29
Experimental dataset validation infrastructure

leroyvn:
# Description

This PR introduces an experimental data validation framework. This is a best-effort solution since specialized validation frameworks are not mature enough for production use.

## What is this for?

This work started during a subsequent documentation update. I realized that we have so far documented our data formats, but nothing exists in our code base to help users verify if their data are correct and, if not, how to fix them.

With this framework, the user can validate a dataset against a registered schema and, in case of a failure, get a full diagnostic of errors.

## How does it work?

This PR provides a simple interface that wraps around the [Cerberus library](https://docs.python-cerberus.org/) (new dependency), which performs validation on a dict-based document extracted from an xarray dataset.

Schemas are saved and versioned alongside the code. They are automatically discovered by the validator.

## How can I use it?

Instantiate a validator:

```python
from eradiate.data._validation import DatasetValidator

v = DatasetValidator()
```

then validate a dataset:

```python
v.validate(ds, schema="particle_dataset_v1")
```

In case of a failure, the `v.errors` member should explain well what is wrong.

## Any issues or caveats?

* This is experimental, we can change the infrastructure at any moment.
* I wish I could use Pandera for this ([xarray integration is planned](https://github.com/unionai-oss/pandera/issues/705), but doesn't seem to progress).
* This system can only validate the dataset schema and metadata, not the data themselves
* Cerberus's rule system is a bit complicated and thorough testing is required.
* The only schema that is registered at the moment is likely incomplete and might give a pass to files that will not work.

# Checklist

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,4 +1,5 @@
 include src/eradiate/config/*.toml
 include src/eradiate/data/downloads_all.yml
 include src/eradiate/data/downloads_minimal.yml
+include src/eradiate/data/schemas/*.yml
 include src/eradiate/units.txt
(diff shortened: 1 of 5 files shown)


leroyvn:
We'll merge this now, it's in alpha stage but we clearly state it.

PR #407, opened 2024-05-03
Fix a bug where geometry vs atmosphere extent check would not work as intended

leroyvn:
# Description

This commit fixes a bug where, for some experiment classes, the check of scene geometry parameters against molecular atmosphere extent would unexpectedly pass or fail.

The code performing this check is refactored to reduce the amount of duplicates.

# Checklist

### DIFF
--- a/src/eradiate/experiments/_dem.py
+++ b/src/eradiate/experiments/_dem.py
@@ -6,7 +6,11 @@ import warnings
 import attrs
 
 from ._core import EarthObservationExperiment
-from ._helpers import measure_inside_atmosphere, surface_converter
+from ._helpers import (
+    check_geometry_atmosphere,
+    measure_inside_atmosphere,
+    surface_converter,
+)
 from ..attrs import documented, parse_docs
 from ..scenes.atmosphere import (
     Atmosphere,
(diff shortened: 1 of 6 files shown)


schunkes:
I have a hard time understanding the difference between the old code and the new, other than the fact that it's now a standalone function instead of a method. Can you tell me why the old code failed and how it is prevented now?

src/eradiate/experiments/_dem.py
```diff
@@ -235,9 +200,9 @@ def _context_kwargs(self) -> dict[str, t.Any]:
 
         for measure in self.measures:
             if measure_inside_atmosphere(self.atmosphere, measure):
-                kwargs[
-                    f"{measure.sensor_id}.atmosphere_medium_id"
-                ] = self.atmosphere.medium_id
+                kwargs[f"{measure.sensor_id}.atmosphere_medium_id"] = (
+                    self.atmosphere.medium_id
```
schunkes, line 204: Is this good coding style? I suppose this happens because of the line length limit, no?

leroyvn: This is because, I think, the pr-ecommit hook for formatting now uses Ruff instead of Black. The two styles are very similar, but not identical.

leroyvn:
The code that got de-duplicated was repeated multiple times, sometimes with an indexing bug.

leroyvn:
More precisely: [this](https://github.com/eradiate/eradiate/pull/407/files#diff-5806b8777b37bc375b7d829e1ef11c93ae6e3e016888355b58b51f6da675b379L210) is what was wrong. The correct code is [here](https://github.com/eradiate/eradiate/pull/407/files#diff-7cef7461c1b4ba2b8aa10369d3d60e83e4bc0eacd0e2ea299dd5db30338b4aafR107).

PR #409, opened 2024-05-21
illumination: Add common interface for directional-like illuminants

leroyvn:
# Description

This PR adds a common interface for the `DirectionalIllumination` and `AstroObjectIllumination` classes. This makes sense, since it is strange to say that "an `AstroObjectIllumination` is a `DirectionalIllumination`".

In addition, we fix the azimuth field init sequence to allow all values and warn if the user inputs something out of the expected range. The idea is to still force values within a known range, but still to avoid modifying user input without notice.

# Checklist

### DIFF
--- a/docs/rst/reference_api/scenes.rst
+++ b/docs/rst/reference_api/scenes.rst
@@ -279,6 +279,7 @@
    :toctree: generated/autosummary/
 
    Illumination
+   AbstractDirectionalIllumination
 
 **Factories**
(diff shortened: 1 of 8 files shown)


wint3ria:
Yes the Astro object whole point is to not be a directional illumination x) This should not have an impact on the API that could interfere with my work.

wint3ria (approved):

Issue #377, opened 2023-11-15
Spectral response function to have one leading zero and one trailing zero


nollety:
# Feature request

## Summary

Provide a way to have spectral response data start with a zero and end with a zero. 
This is a feature request from @Ygovs and @lucio-f (please correct if the details are wrong)

## Reason and benefit

Pre-requisite of another software that takes advantage of the SRF datasets shipped with Eradiate, whose SRF integration procedure is more accurate when the spectral response data have a leading zero and a trailing zero.

## Details

Update the `src/eradiate/srf_tool.py` module to add this feature. The SRF datasets with leading and trailing zero can then be locally generated for use within the said software. Eradiate does not require leading/trailing zeros in the spectral response data.

PR #420, opened 2024-07-01
Improvements to `srf_tools` utilities

leroyvn:
# Description

This PR adds two features to the `srf_tools` module. Changes are as follows:

* A new SRF filtering algorithm is added. It ensures the conservation of the mean wavelength and produces a filtered SRF support symmetrical with respect to it.

  image

  *Example for MODIS band 14. The previous strategy is labelled "walk", and the new one is labelled "symmetry".*

* A new filter is added and allows to pad SRF data with leading and trailing zeros. This contributes to solving #377.

# Checklist

### DIFF
--- a/src/eradiate/srf_tools.py
+++ b/src/eradiate/srf_tools.py
@@ -3,14 +3,16 @@
 from __future__ import annotations
 
 import datetime
-import typing as t
 import warnings
 from pathlib import Path
+from typing import Literal
 
 import matplotlib.pyplot as plt
 import numpy as np
+import numpy.typing as npt
 import pint
 import rich
+import scipy.integrate as spint
 import xarray as xr
 from rich.prompt import Confirm
 from rich.table import Table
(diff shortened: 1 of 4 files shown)


src/eradiate/srf_tools.py
```diff
@@ -539,119 +536,113 @@ def threshold_filter(
...
-    halffraction = (1 - (percentage / 100)) / 2
 
-    # wavelength points that contribute to less than half the percentage of the
-    # integrated response function, starting from the smallest wavelength values
-    w_left = cumsum.where(cumsum < halffraction * cumsum_max, drop=True).w.values
+def _integral_filter_bounds_symmetry(
+    x: npt.ArrayLike, y: npt.ArrayLike, fraction: float
+) -> tuple[tuple[int, int], float]:
+    # Insert mean in x array
+    xmean = np.trapz(y * x, x) / np.trapz(y, x)
```
schunkes, line 558: I suppose this is the mean wavelength, weighted by the SRF?

schunkes (approved):

src/eradiate/srf_tools.py
leroyvn: This is $\int_a^b x f(x) \ \mathrm{d}x / \int_a^b f(x) \ \mathrm{d}x$. It's not the mean of $x f(x)$ (which would be $\frac{1}{b-a} \int_a^b f(x) \ \mathrm{d}x$). I'm actually not sure how to call it.

PR #421, opened 2024-07-03
[Feature] Integration of the new piecewise medium and integrator into Eradiate.

frettini:
# Description

These changes integrate the new piecewise medium and piecewise volumetric integrator added to the eradiate/mitsuba3 repo into Eradiate (see eradiate/mitsuba3#9). Those changes aim to improve the performance of plane-parallel geometry simulations by doing analytical distance sampling and transmittance evaluation.

Plane-parallel geometry simulations will now use the `PiecewiseVolumetricPathIntegrator` by default. Spherical-shell geometry will revert back to the standard `VolumetricPathIntegrator`. The latter can also be enforced by specifying the integrator in the experiment, or by setting the flag `force_majorant` to true in the `Atmosphere` object:
 
```python
majorant_atmosphere = ertsc.atmosphere.MolecularAtmosphere(
    force_majorant=True, scale = 5.0
)
# or
majorant_exp = eradiate.experiments.AtmosphereExperiment(
    ...
    integrator=ertsc.integrators.VolPathIntegrator(),
)
```

The new medium and integrator were tested in various scenarios, using CKD and mono modes, with and without canopy, and over the full spectrum. The full test suite was also run.

# Checklist

- [x] The feature branch is rebased on the current state of the `main` (`next`?) branch

### DIFF
--- a/src/eradiate/experiments/_core.py
+++ b/src/eradiate/experiments/_core.py
@@ -13,8 +13,8 @@ from hamilton.driver import Driver
 
 import eradiate
 
+from .. import converters, validators
 from .. import pipelines as pl
-from .. import validators
 from ..attrs import AUTO, documented, parse_docs
 from ..contexts import KernelContext, MultiGenerator
 from ..kernel import MitsubaObjectWrapper, mi_render, mi_traverse
@@ -27,7 +27,7 @@ from ..scenes.illumination import (
     DirectionalIllumination,
     illumination_factory,
 )
-from ..scenes.integrators import Integrator, PathIntegrator, integrator_factory
+from ..scenes.integrators import Integrator, integrator_factory
 from ..scenes.measure import (
     Measure,
     MultiDistantMeasure,
@@ -95,26 +95,23 @@ class Experiment(ABC):
         default=":class:`MultiDistantMeasure() <.MultiDistantMeasure>`",
     )
 
-    _integrator: Integrator = documented(
+    integrator: Integrator = documented(
         attrs.field(
-            factory=PathIntegrator,
-            converter=integrator_factory.convert,
-            validator=attrs.validators.instance_of(Integrator),
+            default=AUTO,
+            converter=converters.auto_or(integrator_factory.convert),
+            validator=validators.auto_or(
+                attrs.validators.instance_of(Integrator),
+            ),
         ),
         doc="Monte Carlo integration algorithm specification. "
         "This parameter can be specified as a dictionary which will be "
-        "interpreted by :data:`.integrator_factory`.",
-        type=":class:`.Integrator`",
-        init_type=":class:`.Integrator` or dict",
-        default=":class:`PathIntegrator() <.PathIntegrator>`",
+        "interpreted by :data:`.integrator_factory`."
+        "The integrator defaults to :data:`AUTO`, which will choose the appropriate "
+        "integrator depending on the experiment's configuration. ",
+        type=":class:`.Integrator` or AUTO",
+        init_type=":class:`.Integrator` or dict or AUTO",
+        default="AUTO",
     )
-
-    @property
-    def integrator(self) -> Integrator:
-        """
-        :class:`.Integrator`: Integrator used to solve the radiative transfer equation.
-        """
-        return self._integrator
 
     _results: dict[str, xr.Dataset] = attrs.field(factory=dict, repr=False)
 
--- a/src/eradiate/experiments/_helpers.py
+++ b/src/eradiate/experiments/_helpers.py
@@ -1,8 +1,12 @@
 from __future__ import annotations
 
-from ..scenes.atmosphere import Atmosphere, MolecularAtmosphere
+from ..scenes.atmosphere import (
+    AbstractHeterogeneousAtmosphere,
+    Atmosphere,
+    MolecularAtmosphere,
+)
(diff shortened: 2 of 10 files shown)


frettini:
Hi @leroyvn,
I have pushed the new integrator API. However, there might be a flaw in how it handles normalizing bad configurations. If the integrator passed to the experiment has a property set to a value, say `rr_depth=10`, but the integrator is not compatible with the configuration so that it is changed to another integrator, that new instance will have `rr_depth=None`. Is there a graceful way to handle that? Thanks.

src/eradiate/scenes/atmosphere/_core.py
```diff
@@ -372,6 +372,21 @@ class AbstractHeterogeneousAtmosphere(Atmosphere, ABC):
...
+        attrs.field(
+            default=False,
+            kw_only=True,
+            converter=attrs.converters.optional(bool),
+            validator=attrs.validators.optional(attrs.validators.instance_of(bool)),
+        ),
+        doc="""If set to true, uses heterogeneous medium, which is compatible with all
+        integrators except PiecewiseVolpathIntegrator. Otherwise, uses a piecewise medium which 
+        is compatible with PiecewiseVolPathIntegrator and other integrators. This setting 
+        only affects PlaneParallelGeometry, other geometries use heterogeneous mediums.""",
```
leroyvn, line 385: Unfortunately, our docstring formatting is rather fragile and will break with triple-quoted strings. Instead use single-quoted strings and chain them carefully:
```suggestion
        doc="If set to ``True``, use a generic heterogeneous medium implementation, compatible with all "
        "integrators except :class:`.PiecewiseVolpathIntegrator`. "
        "Otherwise, use a piecewise heterogeneous medium which "
        "is compatible with :class:`.PiecewiseVolPathIntegrator` and other integrators. "
        "This setting only applies when using a :class:`.PlaneParallelGeometry`; "
        "other geometries use the generic heterogeneous medium implementation.",
```

leroyvn, line 385: Unfortunately, our docstring formatting is rather fragile and will break with triple-quoted strings. Instead use single-quoted strings and chain them carefully:
```suggestion
        doc="If set to ``True``, use a generic heterogeneous medium implementation, compatible with all "
        "integrators except :class:`.PiecewiseVolpathIntegrator`. "
        "Otherwise, use a piecewise heterogeneous medium which "
        "is compatible with :class:`.PiecewiseVolPathIntegrator` and other integrators. "
        "This setting only applies when using a :class:`.PlaneParallelGeometry`; "
        "other geometries use the generic heterogeneous medium implementation.",
```

leroyvn (changes requested):
Hi @frettini, here are a few minor comments. Fundamentals are what we discussed offline, it's more about details at this point. A few things to do:

* rebase onto the latest next (I aligned it with `main`);
* a tiny bit of cleanup in the code;
* apply a Ruff linting and formatting pass on modified files (if not done already, I recommend to activate [pre-commit hooks](https://eradiate.readthedocs.io/en/stable/rst/contributing.html#style)).

Then we'll merge this!

src/eradiate/experiments/_core.py
```diff
@@ -95,27 +95,24 @@ class Experiment(ABC):
...
+            ),
         ),
         doc="Monte Carlo integration algorithm specification. "
         "This parameter can be specified as a dictionary which will be "
-        "interpreted by :data:`.integrator_factory`.",
-        type=":class:`.Integrator`",
-        init_type=":class:`.Integrator` or dict",
-        default=":class:`PathIntegrator() <.PathIntegrator>`",
+        "interpreted by :data:`.integrator_factory`."
+        "The integrator defaults to :data:`AUTO`, which will choose the appropriate"
```
leroyvn, line 109:
```suggestion
        "interpreted by :data:`.integrator_factory`. "
        "The integrator defaults to :data:`AUTO`, which will choose the appropriate "
```

src/eradiate/experiments/_helpers.py
leroyvn, line 1: Formatter pass is missing

src/eradiate/experiments/_helpers.py
```diff
@@ -119,3 +123,51 @@ def check_geometry_atmosphere(
...
+    atmosphere and geometry.
+
+    Parameters
+    ----------
+    geometry : SceneGeometry
+        An experiment geometry.
+
+    atmosphere : MolecularAtmosphere
+        A molecular atmosphere.
+    """
```
leroyvn, line 142: Please document the returned value.

src/eradiate/experiments/_helpers.py
```diff
@@ -119,3 +123,51 @@ def check_geometry_atmosphere(
...
+def check_volpath_compatible( atmosphere: Atmosphere) -> tuple[bool, str]:
+    """
+    Check that the volpath integrator is compatible with the experiment's
+    atmosphere.
+
+    Parameters
+    ----------
+    atmosphere : MolecularAtmosphere
+        A molecular atmosphere.
+    """
```
leroyvn, line 167: Ditto

leroyvn (approved):
Hi @frettini, looks almost good to me, I only have very minor comments. You can squash the commits that have to be, then we can merge. Thanks!

leroyvn:
Thanks you, let's merge this now! :rocket:

PR #426, opened 2024-07-30
[Feature] Add radiance variance as an output of the simulation.

frettini:
# Description

These changes add the radiance's variance as an output of an Eradiate simulation. This is done by leveraging Mitsuba's `moment` integrator, which outputs the raw second moment in a separate bitmap. The latter is transformed by Eradiate's post processing pipeline into the radiance's variance. 

To enable variance calculations, set `moment=True` in the integrator of your experiment, for example:

```python
integrator = ertsc.integrators.PiecewiseVolPathIntegrator(moment=True)
```

# Checklist

### DIFF
--- a/src/eradiate/experiments/_core.py
+++ b/src/eradiate/experiments/_core.py
@@ -7,6 +7,7 @@ from datetime import datetime
 
 import attrs
 import mitsuba as mi
+import numpy as np
 import pinttr
 import xarray as xr
 from hamilton.driver import Driver
@@ -478,11 +479,25 @@ class EarthObservationExperiment(Experiment, ABC):
 
         for ctx_index, spectral_group_dict in mi_results.items():
             for sensor_id, mi_bitmap in spectral_group_dict.items():
-                measure = sensor_to_measure[sensor_id]
-                measure.mi_results[ctx_index] = {
-                    "bitmap": mi_bitmap,
-                    "spp": spp if spp > 0 else measure.spp,
-                }
+                if self.integrator.moment:
+                    split = mi_bitmap.split()
+                    m2 = split[1][1]
+                    m2_np = np.array(m2, copy=False)[:, :, [0]]
+                    m2 = mi.Bitmap(m2_np, mi.Bitmap.PixelFormat.Y)
+
+                    measure = sensor_to_measure[sensor_id]
+                    measure.mi_results[ctx_index] = {
+                        "bitmap": split[0][1],
+                        "m2": m2,
+                        "spp": spp if spp > 0 else measure.spp,
+                    }
+
+                else:
+                    measure = sensor_to_measure[sensor_id]
+                    measure.mi_results[ctx_index] = {
+                        "bitmap": mi_bitmap,
+                        "spp": spp if spp > 0 else measure.spp,
+                    }
 
     def postprocess(self) -> None:
         # Inherit docstring
@@ -501,7 +516,7 @@ class EarthObservationExperiment(Experiment, ABC):
         # Inherit docstring
         if isinstance(measure, int):
             measure = self.measures[measure]
-        config = pl.config(measure)
+        config = pl.config(measure, integrator=self.integrator)
         return eradiate.pipelines.driver(config)
 
     def _pipeline_inputs(self, i_measure: int):
--- a/src/eradiate/pipelines/core.py
+++ b/src/eradiate/pipelines/core.py
@@ -14,6 +14,7 @@ from rich.table import Table
 import eradiate
 
 from .._mode import Mode
+from ..scenes.integrators import Integrator
 from ..scenes.measure import Measure
 from ..scenes.spectra import InterpolatedSpectrum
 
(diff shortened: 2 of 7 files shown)


src/eradiate/experiments/_core.py
```diff
@@ -478,11 +479,30 @@ def process(
...
-                }
+                
+                if self.integrator.moment :
+                    split = mi_bitmap.split()
+                    
+                    m2 = split[1][1]
+
+                    # convert to single channel
+                    if m2.channel_count() > 1:
+                        m2_np = np.array(m2, copy=False)[:,:,0:1] # note the slice is to keep the channel dimension
```
leroyvn, line 490: I'd rather make this explicit, either with
```suggestion
                        m2_np = np.array(m2, copy=False)[:,:,[0]] # note the slice is to keep the channel dimension
```
or
```suggestion
                        m2_np = np.atleast_3d(np.array(m2, copy=False)[..., 0])
```

src/eradiate/experiments/_core.py
```diff
@@ -478,11 +479,30 @@ def process(
...
-                    "bitmap": mi_bitmap,
-                    "spp": spp if spp > 0 else measure.spp,
-                }
+                
+                if self.integrator.moment :
+                    split = mi_bitmap.split()
+                    
+                    m2 = split[1][1]
+
+                    # convert to single channel
```
leroyvn, line 488: Why do we need to do this conditionally? For sure, with a single component, it wouldn't do much, but that would simplify a bit the control flow.

src/eradiate/experiments/_core.py
```diff
@@ -478,11 +479,30 @@ def process(
...
             for sensor_id, mi_bitmap in spectral_group_dict.items():
-                measure = sensor_to_measure[sensor_id]
-                measure.mi_results[ctx_index] = {
-                    "bitmap": mi_bitmap,
-                    "spp": spp if spp > 0 else measure.spp,
-                }
+                
+                if self.integrator.moment :
+                    split = mi_bitmap.split()
+
```
leroyvn, line 485:
```suggestion
```

src/eradiate/pipelines/core.py
```diff
@@ -14,14 +14,19 @@
...
 # Disable Hamilton telemetry (we don't want to bother our users with it)
 telemetry.disable_telemetry()
 
 
-def config(measure: Measure, mode: Mode | str | None = None) -> dict:
+def config(
+    measure: Measure,
+    mode: Mode | str | None = None,
+    integrator: Integrator | None = None,
+) -> dict:
```
leroyvn, line 29: You'll have to update the docstring here.

src/eradiate/pipelines/core.py
```diff
@@ -62,6 +67,9 @@ def config(measure: Measure, mode: Mode | str | None = None) -> dict:
     # Shall we apply spectral response function weighting (a.k.a convolution)?
     result["apply_spectral_response"] = isinstance(measure.srf, InterpolatedSpectrum)
 
+    # Should we calculate the variance in the result.
```
leroyvn, line 70:
```suggestion
    # Should we calculate the variance in the result?
```

src/eradiate/experiments/_core.py
leroyvn, line 1: I think this file is missing a Ruff formatting pass.

src/eradiate/scenes/integrators/_core.py
```diff
@@ -53,3 +53,14 @@ class Integrator(NodeSceneElement, ABC):
...
+    moment: bool = documented(
+        attrs.field(
+            default=False,
+            converter=attrs.converters.optional(bool),
+        ),
+        doc="If true, calculates the variance of the output film",
+        type="bool",
+        init_type="bool, optional",
+        default="False",
+    )
```
leroyvn, line 66: The famous trailing newline:
```suggestion
    )

```

src/eradiate/scenes/integrators/_path_tracers.py
```diff
@@ -61,6 +61,10 @@ def template(self) -> dict:
         if self.hide_emitters is not None:
             result["hide_emitters"] = self.hide_emitters
 
+        if self.moment:
+            result = { "type":"moment", "nested":result }
+        
+
```
leroyvn, line 67:
```suggestion
```

leroyvn (commented):
Hi @frettini, thanks for this. I have a few minor comments. Once they are addressed, we can merge. I would also appreciate if you could add an entry to the change log to make sure we don't forget anything on the next release.

src/eradiate/experiments/_core.py
frettini: Do you mean remove the check for more than one channel altogether? I have removed it.

frettini:
I have pushed the suggested changes, but would need some pointers on where I can modify the changelogs. Thanks!

leroyvn (approved):
Looks good to me, let's merge!

PR #428, opened 2024-08-01
Refactor DEM handling components

leroyvn:
# Description

This PR refactors digital elevation model support components. Notably:

* The `mesh_from_dem()` function is broken down into more fundamental pieces. Its logic is also simplified and clarified based on four cases, two of which are unsafe. The actual meshing work is done by an independent function that simply creates a triangulation from a 2D array of data points.
* Longitude/latitude from/to x/y conversions are done using a Mercator projection. Better than before, but still not idea. We recommend that users perform this operation themselves. We might introduce support for automatic projection in the future.
* Edge vertex snap-to-zero is tested and works, but it is not automated: this is left to the user.
* Test code is reorganized and cleaned up after the changes made to the various utility functions in the DEM module. I renamed one of the test files to disambiguate during pytest case collection.

To be a bit more concrete, here are simulated radiance records obtained with [GEBCO](https://www.gebco.net/data_and_products/gridded_bathymetry_data/) bathymetry data (I extracted Brittany from this dataset and scaled elevation to make things a bit more visible):

| Plane-parallel | Spherical-shell |
|---|---|
| image | image |

The plane-parallel version uses the simple Mercator projection and is wrong: as a user, it's most likely more appropriate to use a more sophisticated and correct projection.

In both cases, mesh edge vertices are snapped to the background manually (takes 2 lines with xarray). Users might want to move the background or offset the elevation to adjust the seam.

# Checklist

### DIFF
--- a/docs/rst/reference_api/scenes.rst
+++ b/docs/rst/reference_api/scenes.rst
@@ -291,6 +291,7 @@ Quick access
    :toctree: generated/autosummary/
 
    mesh_from_dem
+   triangulate_grid
 
 .. _module-eradiate.scenes.bsdfs:
 
--- a/docs/src/release_notes/v0.29.x.md
+++ b/docs/src/release_notes/v0.29.x.md
@@ -7,9 +7,11 @@
 % ### Removed
 %
 % ### Added
-%
-% ### Changed
-%
+
+### Changed
+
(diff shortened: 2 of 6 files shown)


schunkes (approved):
Looks good. Much clearer than what I wrote before.

PR #460, opened 2024-10-31
[Feature] Depolarization factor for Rayleigh phase function.

frettini:
# Description

This PR introduces methods to calculate the depolarization factor from the radiative properties of the atmosphere. It also modifies the interface for the depolarization factor in `RayleighPhaseFunction` and `MolecularAtmosphere`. 

`MolecularAtmosphere`: Has a new `rayleigh_depolarization` field which accepts an `ndarray` or `str`. The latter case indicates the use of a method based on radiative properties to calculate the depolarization factor. This can be either `bates` or `bodhaine` and those are implemented in `radprops/_atmosphere.py`. `bates` is dependent on wavelength only and `bodhaine` on wavelength and CO2 concentration.

`RayleighPhaseFunction` : Accepts a depolarization factor as an `array`, `float`, or `callable`. Will resolve it into a 1D array when rendering the kernel dictionary.

Notes:
- The depolarization factor is only relevant when using a polarization mode.
- This change breaks the interface of `MolecularAtmosphere` as you cannot pass a phase anymore, it is assumed to be a Rayleigh phase function.

# Checklist


src/eradiate/scenes/phase/_rayleigh.py
```diff
@@ -19,26 +27,142 @@ class RayleighPhaseFunction(PhaseFunction):
...
     """
 
-    depolarization: float | None = documented(
+    depolarization: np.ndarray | t.Callable[[KernelContext], np.ndarray] = documented(
         attrs.field(
-            default=None,
-            converter=attrs.converters.optional(float),
+            converter=lambda x: x if callable(x) else np.array(x, dtype=np.float64),
+            kw_only=True,
+            default=np.array([0.0]),
```
leroyvn, line 34: Avoid mutable defaults: here, you need a factory.
```suggestion
            factory=lambda: np.array([0.0]),
```

tests/01_unit/scenes/phase/test_rayleigh.py
```diff
@@ -13,3 +17,39 @@ def test_rayleigh(modes_all_double):
...
+    check_scene_element(phase, mi.PhaseFunction)
+
+
+def test_rayleigh_spherical_shell(modes_all_double):
+    phase = RayleighPhaseFunction(geometry=SphericalShellGeometry())
+    check_scene_element(phase, mi.PhaseFunction)
+
+
+def test_depolarization(mode_ckd_polarized):
+    si = si = CKDSpectralIndex(w=550 * ureg.nm)
```
leroyvn, line 34:
```suggestion
    si = CKDSpectralIndex(w=550 * ureg.nm)
```

tests/01_unit/scenes/phase/test_rayleigh.py
```diff
@@ -13,3 +17,39 @@ def test_rayleigh(modes_all_double):
...
+
+    phase = RayleighPhaseFunction()
+    depol = phase.eval_depolarization_factor(si)
+    kdict, _ = traverse(phase)
+    check_scene_element(phase, mi.PhaseFunction)
+    assert isinstance(depol, np.ndarray)
+    assert depol.shape[0] == 1
```
leroyvn, line 41: Maybe make these a full shape check, not just the first dimension?

tests/01_unit/scenes/atmosphere/test_molecular_atmosphere.py
```diff
@@ -137,3 +138,37 @@ def test_molecular_atmosphere_switches(
...
+
+
+def test_molecular_atmosphere_depolarization(mode_ckd):
+    atmosphere = MolecularAtmosphere(rayleigh_depolarization=0.5)
+    si = default_spectral_index(atmosphere)
+    depol = atmosphere.eval_depolarization_factor(si)
+    template, _ = traverse(atmosphere)
+    assert template.render(KernelContext(si=si))
+    assert isinstance(depol, pint.Quantity)
+    assert len(depol) == 1
```
leroyvn, line 150: How about turning this into a shape check? (you can use `np.shape()` if `depol` is not an array).

leroyvn (changes requested):
Hi @frettini, thanks for this PR. I have a few comments (nothing fundamental, we already sorted out the majors issues).

src/eradiate/radprops/rayleigh.py
```diff
@@ -22,13 +22,46 @@
...
+            self._load_dataset()
+        return self._king_factor
+
+    @property
+    def wavelength(self):
+        if self._wavelength is None:
+            self._load_dataset()
+        return self._wavelength
+
+    def interp(self, w: pint.Quantity = ureg.Quantity(0.550, "micron")):
```
leroyvn, line 40: Since all fields are unitless, I think we can move the conversion to the upper level.
```suggestion
    def interp(self, w: float = 0.550):
```

src/eradiate/radprops/rayleigh.py
```diff
@@ -22,13 +22,46 @@
...
+        """
+        Interpolate `king_factor` over the `wavelength` dimension.
+
+        Parameters
+        ----------
+        w : quantity
```
leroyvn, line 46:
```suggestion
        w : float
```

src/eradiate/radprops/rayleigh.py
```diff
@@ -22,13 +22,46 @@
...
+
+    @property
+    def wavelength(self):
+        if self._wavelength is None:
+            self._load_dataset()
+        return self._wavelength
+
+    def interp(self, w: pint.Quantity = ureg.Quantity(0.550, "micron")):
+        """
+        Interpolate `king_factor` over the `wavelength` dimension.
```
leroyvn, line 42:
```suggestion
        Interpolate ``king_factor`` over the ``wavelength`` dimension.
```

src/eradiate/radprops/rayleigh.py
```diff
@@ -22,13 +22,46 @@
...
+        Returns
+        -------
+        ndarray
+            Bates' King factor
+        """
+        left = self.king_factor[0]
+        right = self.king_factor[-1]
+
+        return np.interp(
+            w.m_as("micron"), self.wavelength, self.king_factor, left=left, right=right
```
leroyvn, line 58:
```suggestion
            w, self.wavelength, self.king_factor, left=left, right=right
```

src/eradiate/radprops/rayleigh.py
```diff
@@ -82,12 +115,8 @@ def compute_sigma_s_air(
     # as well, meaning that this conversion is anyway necessary.
     w = wavelength.to("micron")
```
leroyvn, line 116:
```suggestion
    w = wavelength.m_as("micron")
```

src/eradiate/scenes/phase/_rayleigh.py
```diff
@@ -19,26 +27,142 @@ class RayleighPhaseFunction(PhaseFunction):
...
         attrs.field(
-            default=None,
-            converter=attrs.converters.optional(float),
+            converter=lambda x: x if callable(x) else np.array(x, dtype=np.float64),
+            kw_only=True,
+            factory=lambda: np.array([0.0]),
         ),
         doc="The ratio of intensities parallel and perpendicular to the"
         "plane of scattering for light scattered at 90 deg. Only relevant"
         "when using a polarization mode.",
```
leroyvn, line 38:
```suggestion
        doc="The ratio of intensities parallel and perpendicular to the "
        "plane of scattering for light scattered at 90 deg. Only relevant "
        "when using a polarization mode.",
```

leroyvn (changes requested):
Just a few more small comments, and I think we're good to go!

src/eradiate/radprops/rayleigh.py
frettini: Are there any cases where `w` can be an array?

leroyvn: That is a good point, I overlooked that. Indeed, `w` should accept an array-like.

PR #465, opened 2024-12-05
[Fix] Set all volume's filter types to `nearest`

frettini:
# Description

Hello, I found that the `gridvolumes` used in Eradiate don't set their filter types, which means they default to a trilinear interpolation filter. This in turn yields inexact values during the integration. This is usually not a problem as Eradiate uses many layers to represent the atmosphere. Furthermore, the null collision integrator is robust to this filter. This is not the case for the piecewise integrator, where inexact values of the extinction coefficient can lead to positive throughput, and thus very large errors.
This fix makes sure that all instance of `gridvolumes` see their `filter_type` set to `nearest`. This also includes a change to the blend phase unit test that need to check for dictionary parity.

# Checklist

- [X] The code follows the relevant coding guidelines
- [X] The code generates no new warnings
- [X] The code is appropriately documented
- [X] The code is tested to prove its function
- [X] The feature branch is rebased on the current state of the `next` branch
- [X] I give permission that the Eradiate project may redistribute my contributions under the terms of its license

### DIFF
--- a/docs/src/release_notes/v0.30.x.md
+++ b/docs/src/release_notes/v0.30.x.md
@@ -10,8 +10,8 @@
 
 ### Removed
 
-* The {class}`MolecularAtmosphere` constructor no longer accepts a `phase_function`
-  parameter ({ghpr}`460`).
+* The {class}`MolecularAtmosphere` constructor no longer accepts a
+  `phase_function` parameter ({ghpr}`460`).
 
 ### Added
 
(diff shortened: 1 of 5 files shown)


leroyvn (approved):
Hi Nicolae, thanks a lot for debugging and fixing this. Looks good to me, I'll merge this after updating the change log.

leroyvn:
We're good to go, I'll merge this now.

PR #468, opened 2024-12-17
[Feature] New `ArrayRadProfile` object

frettini:
# Description

Hello, this PR introduces ArrayRadProp, a new RadProfile that enables the user to pass radiative properties manually. Namely, it allows to pass `sigma_s` and `sigma_a` as `DataArray`. Those values are then interpolated given a `ZGrid`. The need for such a class has come to light when working on the IPRT benchmark, in which absoprtion and scattering data are directly provided for given scenarios. This feature can thus be useful for benchmarking as well as debugging.

Of note are the functions `eval_sigma_s_mono` and and `eval_sigma_a_mono`. Please also note that I have left the `ZGrid` and `levels` properties as `NotImplementedError` as I was not able to understand how those properties were used. I would also be interested in any ways of making the check in `_helpers.py` more elegant. 

The pytest  suite was run, the new feature tested in my work for IPRT and additional tests were written. Let me know if you have any comment or feedback. Many Thanks.

# Checklist

### DIFF
--- a/src/eradiate/radprops/_array.py
+++ b/src/eradiate/radprops/_array.py
@@ -35,10 +35,10 @@ class ArrayRadProfile(RadProfile):
         attrs.field(
             default=None,
         ),
-        doc="``DataArray`` of absorption coefficients "
+        doc="``DataArray`` of absorption coefficients. "
         "The ``DataArray`` is composed of two dimensions ``{w, z}`` representing the "
         "wavelength and altitude respectively. Note that ``w`` must be of length "
-        "2 minimum to be correctly interpolated. ",
+        "2 minimum to be correctly interpolated.",
         type="DataArray or None",
         init_type="DataArray or None",
         default="None",
@@ -48,10 +48,10 @@ class ArrayRadProfile(RadProfile):
         attrs.field(
             default=None,
         ),
-        doc="``DataArray`` of scattering coefficients "
+        doc="``DataArray`` of scattering coefficients. "
         "The ``DataArray`` is composed of two dimensions ``{w, z}`` representing the "
         "wavelength and altitude respectively. Note that ``w`` must be of length "
-        "2 minimum to be correctly interpolated. ",
+        "2 minimum to be correctly interpolated.",
         type="DataArray or None",
         init_type="DataArray or None",
         default="None",
@@ -94,16 +94,19 @@ class ArrayRadProfile(RadProfile):
         "A ``ndarray`` will be interpreted as a description of the depolarization "
         "factor at different levels of the atmosphere. Must be shaped (N,) with "
         "N the number of layers.",
+        init_type="array-like, optional",
+        default="[0]",
     )
 
-    interpolation_method: str = documented(
+    interpolation_method: t.Literal["nearest", "linear"] = documented(
         attrs.field(
             default="nearest",
             converter=str,
-            validator=attrs.validators.instance_of(str),
+            validator=attrs.validators.in_(["nearest", "linear"]),
         ),
-        doc="Method of interpolation of the absorption and scattering coefficients. ",
+        doc="Method of interpolation of the absorption and scattering coefficients.",
         type="str",
+        init_type="{'nearest', 'linear'}",
         default="nearest",
     )
 
@@ -113,8 +116,9 @@ class ArrayRadProfile(RadProfile):
             converter=dict,
             validator=attrs.validators.instance_of(dict),
         ),
-        doc="Interpolation arguments passed to xarray during interpolation. ",
+        doc="Keyword arguments passed to :meth:`xarray.DataArray.interp` when called.",
         type="dict",
+        init_type="dict, optional",
     )
 
     def __attrs_post_init__(self):
@@ -122,6 +126,14 @@ class ArrayRadProfile(RadProfile):
 
     def update(self) -> None:
         pass
+
+    @property
+    def zbounds(self) -> tuple[pint.Quantity, pint.Quantity]:
+        z_sigma_s = to_quantity(self.sigma_s.z)
+        z_sigma_a = to_quantity(self.sigma_a.z)
+        z_max = min(z_sigma_s.max(), z_sigma_a.max())
+        z_min = max(z_sigma_s.min(), z_sigma_a.min())
+        return z_min, z_max
 
     @property
     def levels(self) -> pint.Quantity:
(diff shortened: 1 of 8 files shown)


src/eradiate/radprops/_array.py
```diff
@@ -0,0 +1,254 @@
...
+    """
+
+    sigma_a: xr.DataArray | None = documented(
+        attrs.field(
+            default=None,
+        ),
+        doc="``DataArray`` of absorption coefficients "
+        "The ``DataArray`` is composed of two dimensions ``{w, z}`` representing the "
+        "wavelength and altitude respectively. Note that ``w`` must be of length "
+        "2 minimum to be correctly interpolated. ",
```
leroyvn, line 41:
```suggestion
        doc="``DataArray`` of absorption coefficients. "
        "The ``DataArray`` is composed of two dimensions ``{w, z}`` representing the "
        "wavelength and altitude respectively. Note that ``w`` must be of length "
        "2 minimum to be correctly interpolated.",
```

src/eradiate/radprops/_array.py
```diff
@@ -0,0 +1,254 @@
...
+    )
+
+    sigma_s: xr.DataArray | None = documented(
+        attrs.field(
+            default=None,
+        ),
+        doc="``DataArray`` of scattering coefficients "
+        "The ``DataArray`` is composed of two dimensions ``{w, z}`` representing the "
+        "wavelength and altitude respectively. Note that ``w`` must be of length "
+        "2 minimum to be correctly interpolated. ",
```
leroyvn, line 54:
```suggestion
        doc="``DataArray`` of scattering coefficients. "
        "The ``DataArray`` is composed of two dimensions ``{w, z}`` representing the "
        "wavelength and altitude respectively. Note that ``w`` must be of length "
        "2 minimum to be correctly interpolated.",
```

src/eradiate/radprops/_array.py
```diff
@@ -0,0 +1,254 @@
...
+        default="True",
+    )
+
+    rayleigh_depolarization: np.ndarray = documented(
+        attrs.field(
+            converter=lambda x: np.array(x, dtype=np.float64),
+            kw_only=True,
+            factory=lambda: np.array(0.0),
+        ),
+        type="ndarray",
```
leroyvn, line 92: I missed this previously: specifying the default and the allowed init types would be good (see also on the `AtmosphereRadProfile` side).

src/eradiate/radprops/_array.py
```diff
@@ -0,0 +1,254 @@
...
+        "N the number of layers.",
+    )
+
+    interpolation_method: str = documented(
+        attrs.field(
+            default="nearest",
+            converter=str,
+            validator=attrs.validators.instance_of(str),
+        ),
+        doc="Method of interpolation of the absorption and scattering coefficients. ",
```
leroyvn, line 105:
```suggestion
        doc="Method of interpolation of the absorption and scattering coefficients.",
```

src/eradiate/radprops/_array.py
```diff
@@ -0,0 +1,254 @@
...
+        type="str",
+        default="nearest",
```
leroyvn, line 107: Which methods do we support? I think making the allowed options explicit would prevent unpleasant mistakes. That would imply changing the type to something like `Literal["nearest", "linear"]` and the `init_type` doc field to something like `{"nearest", "linear"}`. The validator can then be `attrs.validators.in_`.

src/eradiate/radprops/_array.py
```diff
@@ -0,0 +1,254 @@
...
+
+    interpolation_kwargs: dict[str, t.Any] = documented(
+        attrs.field(
+            factory=dict,
+            converter=dict,
+            validator=attrs.validators.instance_of(dict),
+        ),
+        doc="Interpolation arguments passed to xarray during interpolation. ",
+        type="dict",
+    )
```
leroyvn, line 118:
```suggestion
        init_type="dict, optional",
    )
```

src/eradiate/radprops/_array.py
```diff
@@ -0,0 +1,254 @@
...
+        default="nearest",
+    )
+
+    interpolation_kwargs: dict[str, t.Any] = documented(
+        attrs.field(
+            factory=dict,
+            converter=dict,
+            validator=attrs.validators.instance_of(dict),
+        ),
+        doc="Interpolation arguments passed to xarray during interpolation. ",
```
leroyvn, line 116:
```suggestion
        doc="Keyword arguments passed to :meth:`xarray.DataArray.interp` when called.",
```
Also,

src/eradiate/scenes/atmosphere/_molecular.py
```diff
@@ -155,13 +157,22 @@ def _switch_validator(self, attribute, value):
...
-        init=False,
-        repr=False,
+        doc="Radiative properties object. "
+        "The default is None. If ``thermoprops`` is not None, "
+        "``radprops_profile`` will automatically create an .AtmosphereRadProfile. "
+        "Note that at least ``thermoprops`` or ``radprops_profile`` should be set "
+        "to a valid value. ",
+        type=".RadProfile or None",
+        init_type=".RadProfile or None",
+        default="None",
```
leroyvn, line 175:
```suggestion
        doc="Radiative property profile. "
        "If ``thermoprops`` is not ``None``, this field is automatically "
        "overridden with an :class:`.AtmosphereRadProfile` during intialization. "
        "Note that at least one of ``thermoprops`` or ``radprops_profile`` must "
        "not be ``None``.",
        type=".RadProfile or None",
        init_type=".RadProfile or None",
        default="None",
```

leroyvn (changes requested):
Hi @frettini, a couple of comments and suggestions in addition to an offline discussion. Thanks a lot!

src/eradiate/radprops/_array.py
frettini: I have added the changes, I am also unsure when to add "optional" to the "init_type". Not all attributes with defaults have it, is there a specific rule for it?

leroyvn: The rule is defined in the [Numpydoc style guide](https://numpydoc.readthedocs.io/en/latest/format.html#parameters). Arguably, we haven't always applied it very well.

leroyvn:
Thanks you @frettini, we're good to go!

leroyvn (approved):
Thank you @frettini, we're good to go!

PR #471, opened 2025-02-24
Add option to select specific measures to process

leroyvn:
# Description

This PR updates the `run()` function and the `Experiment.process()` and `Experiment.postprocess()` methods to allow selecting the measures that will be computed. The default behaviour remains unchanged.

# Checklist


frettini (approved):
Looks good, it could be interesting being able to refer to the sensors using the sensor_id or the measure object directly, but doing so could add too much complexity.

leroyvn:
Yes; ideally, I'd like to have a bidirectional measure id-index resolver. That's easy to implement, and this correspondence is needed in several places.

PR #478, opened 2025-03-13
Test report refactoring

wint3ria:
# Description

This PR integrates the Robot Framework alongside pytest in order to benefit from its flexible report generation tool. In particular, Robot allows to generate HTML and to include arbitrary charts directly from the individual test code.

Relevant dependencies and pixi commands have been added to the pyproject.toml configurate file. A base CI has been implemented to generate new reports and save the HTML and PDF versions as Github artefacts.

Existing tests have been adapted to take advantage of the Robot logging features. In particular, existing statistical regression tests have been adapted to generate HTML charts and include them in the Robot report.

# Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,7 +20,7 @@ jobs:
           pip install eradiate-mitsuba
       - name: Run unit tests
         run: |
-          pytest -m "not slow" tests/01_unit
+          pytest -m "not slow" tests/01_unit/scenes/spectra # subset for dev purposes
       - name: Setup chrome
         if: always()
         uses: browser-actions/setup-chrome@v1
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,7 +20,7 @@ jobs:
           pip install eradiate-mitsuba
       - name: Run unit tests
         run: |
-          pytest tests/01_unit
+          pytest -m "not slow" tests/01_unit
       - name: Setup chrome
         uses: browser-actions/setup-chrome@v1
       - name: Robot report
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,20 +33,25 @@ jobs:
       - name: Save HTML log
         uses: actions/upload-artifact@v4
         with:
+          name: log_html
           path: ./log.html
       - name: Save HTML report
         uses: actions/upload-artifact@v4
         with:
+          name: report_html
           path: ./report.html
       - name: Save PDF log
         uses: actions/upload-artifact@v4
         with:
+          name: log_pdf
           path: ./log.pdf
       - name: Save PDF report
         uses: actions/upload-artifact@v4
         with:
+          name: report_pdf
           path: ./report.pdf
       - name: Save test outputs
         uses: actions/upload-artifact@v4
         with:
+          name: output_xml
           path: ./output.xml
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,8 +22,10 @@ jobs:
         run: |
           pytest -m "not slow" tests/01_unit
       - name: Setup chrome
+        if: always()
         uses: browser-actions/setup-chrome@v1
       - name: Robot report
+        if: always()
         run: |
           rebot --expandkeywords name:* ./output.xml
           chrome --no-sandbox --virtual-time-budget=10000 --headless --print-to-pdf=./log.pdf    ./log.html    --no-pdf-header-footer
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,3 +30,23 @@ jobs:
           rebot --expandkeywords name:* ./output.xml
           chrome --no-sandbox --virtual-time-budget=10000 --headless --print-to-pdf=./log.pdf    ./log.html    --no-pdf-header-footer
           chrome --no-sandbox --virtual-time-budget=10000 --headless --print-to-pdf=./report.pdf ./report.html --no-pdf-header-footer
+      - name: Save HTML log
+        uses: actions/upload-artifact@v4
+        with:
+          path: ./log.html
+      - name: Save HTML report
+        uses: actions/upload-artifact@v4
+        with:
+          path: ./report.html
+      - name: Save PDF log
+        uses: actions/upload-artifact@v4
+        with:
+          path: ./log.pdf
+      - name: Save PDF report
+        uses: actions/upload-artifact@v4
+        with:
+          path: ./report.pdf
+      - name: Save test outputs
+        uses: actions/upload-artifact@v4
+        with:
+          path: ./output.xml
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,30 @@
+name: Eradiate test
+
+on:
+  workflow_dispatch:
+  pull_request:
+
+jobs:
+  robot:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Install Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install .[doc,recommended,test,dev]
+          pip install eradiate-mitsuba
+      - name: Run unit tests
+        run: |
+          pytest tests/01_unit
+      - name: Setup chrome
+        uses: browser-actions/setup-chrome@v1
+      - name: Robot report
+        run: |
+          rebot --expandkeywords name:* ./output.xml
+          chrome --no-sandbox --virtual-time-budget=10000 --headless --print-to-pdf=./log.pdf    ./log.html    --no-pdf-header-footer
+          chrome --no-sandbox --virtual-time-budget=10000 --headless --print-to-pdf=./report.pdf ./report.html --no-pdf-header-footer
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,9 +18,9 @@ jobs:
           python -m pip install --upgrade pip
           pip install .[doc,recommended,test,dev]
           pip install eradiate-mitsuba
-      - name: Run unit tests
+      - name: Run fast tests
         run: |
-          pytest -m "not slow" tests/01_unit/scenes/spectra # subset for dev purposes
+          pytest tests --robot-tagstatinclude unit --robot-tagstatinclude system --robot-tagstatinclude regression --robot-tagstatinclude slow -m "not slow"
       - name: Setup chrome
         if: always()
         uses: browser-actions/setup-chrome@v1
@@ -31,26 +31,31 @@ jobs:
           chrome --no-sandbox --virtual-time-budget=10000 --headless --print-to-pdf=./log.pdf    ./log.html    --no-pdf-header-footer
           chrome --no-sandbox --virtual-time-budget=10000 --headless --print-to-pdf=./report.pdf ./report.html --no-pdf-header-footer
       - name: Save HTML log
+        if: always()
         uses: actions/upload-artifact@v4
         with:
           name: log_html
           path: ./log.html
       - name: Save HTML report
+        if: always()
         uses: actions/upload-artifact@v4
         with:
           name: report_html
           path: ./report.html
       - name: Save PDF log
+        if: always()
         uses: actions/upload-artifact@v4
         with:
           name: log_pdf
           path: ./log.pdf
       - name: Save PDF report
+        if: always()
         uses: actions/upload-artifact@v4
         with:
           name: report_pdf
           path: ./report.pdf
       - name: Save test outputs
+        if: always()
         uses: actions/upload-artifact@v4
         with:
           name: output_xml
--- a/src/eradiate/scenes/spectra/_uniform.py
+++ b/src/eradiate/scenes/spectra/_uniform.py
@@ -23,7 +23,7 @@ class UniformSpectrum(Spectrum):
     value: pint.Quantity = documented(
         attrs.field(
             converter=lambda x: float(x) if isinstance(x, int) else x,
-            repr=lambda x: f"{x:~P}",
+            repr=lambda x: str(x),
             kw_only=True,
         ),
         doc="Uniform spectrum value. If a quantity is passed, units are "
--- a/src/eradiate/test_tools/regression.py
+++ b/src/eradiate/test_tools/regression.py
@@ -91,11 +91,13 @@ def regression_test_plots(
     plt.tight_layout()
 
 
-def render_svg_report():
+def render_svg_chart():
     str_i = StringIO()
     plt.savefig(str_i, format="svg", transparent=True)
     svg = str_i.getvalue()
 
+    # Include some CSS in the SVG to render nicely in Robot report's dark and
+    # light modes
     return "\n".join(
         [
             "<svg",
@@ -107,9 +109,6 @@ def render_svg_report():
             "    path {",
             "        fill: var(--text-color);",
             "        stroke: var(--text-color);",
-            "    }",
-            "    #legend_1 > :first-child {",
-            "        opacity: 0.1;",
             "    }",
             "</style>",
             svg,
@@ -391,7 +390,7 @@ class RegressionTest(ABC):
 
             regression_test_plots(ref, val, vza, (self.METRIC_NAME, metric_value))
 
-        html_svg = render_svg_report()
+        html_svg = render_svg_chart()
         logger.info(html_svg, html=True, also_console=False)
         logger.info(f"Saving PNG report chart to {fname_plot}", also_console=True)
 
--- a/src/eradiate/test_tools/regression.py
+++ b/src/eradiate/test_tools/regression.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import os
 import typing as t
 from abc import ABC, abstractmethod
+from io import StringIO
 from pathlib import Path
 
 import attrs
@@ -10,6 +11,7 @@ import matplotlib.pyplot as plt
 import mitsuba as mi
 import numpy as np
 import xarray as xr
+from robot.api import logger
 
 from .. import data
 from ..attrs import define, documented
@@ -22,33 +24,44 @@ def regression_test_plots(
     ref: np.typing.ArrayLike,
     result: np.typing.ArrayLike,
     vza: np.typing.ArrayLike,
-    filename: PathLike,
     metric: tuple[str, float],
+    ref_var: t.Optional[np.typing.ArrayLike] = None,
+    result_var: t.Optional[np.typing.ArrayLike] = None,
 ) -> None:
     """
-    Create regression test report plots.
+    Create regression test report plots. Plot errorbars if both ref_var and
+    result_var are set.
 
     Parameters
     ----------
     ref : array-like
-        BRF values for the reference data
+        Variable values for the reference data
 
     result : array-like
-        BRF values for the simulation result
+        Variable values for the simulation result
 
     vza : array-like
         VZA values for plotting
 
-    filename : path-like
-        Path to the output file for the plot
-
     metric : tuple
         A tuple of the form (metric name, value) to be added to the plots.
+
+    ref_var: array-like (optional)
+        Variable variance for the reference data
+
+    result_var : array-like (optional)
+        Variable variance for the simulation result
+
     """
     fig, axes = plt.subplots(2, 2, figsize=(8, 6))
 
-    axes[0][0].plot(vza, ref, label="reference")
-    axes[0][0].plot(vza, result, label="result")
+    if ref_var is None or result_var is None:
+        axes[0][0].plot(vza, ref, label="reference")
+        axes[0][0].plot(vza, result, label="result")
+    else:
+        axes[0][0].errorbar(vza, ref, yerr=np.sqrt(ref_var), label="reference")
+        axes[0][0].errorbar(vza, result, yerr=np.sqrt(result_var), label="result")
+
     axes[0][0].set_title("Reference and test result")
     handles, labels = axes[0][0].get_legend_handles_labels()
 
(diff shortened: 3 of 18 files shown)


wint3ria:
@leroyvn I face an issue with some UTC deprecation warnings, did you already investigate these? I also face an issue with the datasets that are not available in the CI. I could have implemented their download using the eradiate CLI but I wanted to agree with you on a strategy for this.

src/eradiate/scenes/spectra/_uniform.py
```diff
@@ -23,7 +23,7 @@ class UniformSpectrum(Spectrum):
     value: pint.Quantity = documented(
         attrs.field(
             converter=lambda x: float(x) if isinstance(x, int) else x,
-            repr=lambda x: f"{x:~P}",
+            repr=lambda x: str(x),
```
leroyvn, line 26: Any reason to change this? This format spec was a conscious choice at the time.

src/eradiate/test_tools/regression.py
```diff
@@ -22,33 +24,44 @@ def regression_test_plots(
...
     vza : array-like
         VZA values for plotting
 
-    filename : path-like
-        Path to the output file for the plot
-
     metric : tuple
         A tuple of the form (metric name, value) to be added to the plots.
+
+    ref_var: array-like (optional)
```
leroyvn, line 49: As per the [Numpydoc spec](https://numpydoc.readthedocs.io/en/latest/format.html#parameters):
```suggestion
    ref_var: array-like, optional
```

src/eradiate/test_tools/regression.py
```diff
@@ -22,33 +24,44 @@ def regression_test_plots(
...
+        Variable variance for the reference data
+
+    result_var : array-like (optional)
```
leroyvn, line 52:
```suggestion
    result_var : array-like, optional
```

src/eradiate/test_tools/regression.py
```diff
@@ -62,17 +75,50 @@ def regression_test_plots(
...
+            horizontalalignment="center",
+        )
+    else:
+        axes[0][1].text(
+            0.5, 0.5, f"{metric[0]}: {metric[1]:.4}", horizontalalignment="center"
+        )
 
     plt.tight_layout()
-    plt.savefig(filename)
-    plt.close()
```
leroyvn, line 71: Since you are no longer closing the figure nor returning it, I suppose you are relying on pyplot's internal state to use in another scope? This is generally unsafe, you should instead return the created figure instance.

src/eradiate/test_tools/regression.py
```diff
@@ -62,17 +75,50 @@ def regression_test_plots(
...
+
+
+def render_svg_chart():
+    str_i = StringIO()
+    plt.savefig(str_i, format="svg", transparent=True)
```
leroyvn, line 96: Just to avoid having text cut off:
```suggestion
    plt.savefig(str_i, format="svg", transparent=True, bbox_inches="tight")
```

src/eradiate/test_tools/regression.py
```diff
@@ -62,17 +75,50 @@ def regression_test_plots(
...
+            0.5,
+            f'Metric "{metric[0]}" is not available',
+            horizontalalignment="center",
+        )
+    else:
+        axes[0][1].text(
+            0.5, 0.5, f"{metric[0]}: {metric[1]:.4}", horizontalalignment="center"
+        )
 
     plt.tight_layout()
```
leroyvn, line 91: I am not a big fan of the tight layout, I'd rather keep the basic one:

```suggestion
```

src/eradiate/test_tools/regression.py
```diff
@@ -62,17 +75,50 @@ def regression_test_plots(
...
+            "</style>",
+            svg,
+            "</svg>",
+        ]
+    )
 
 
 def reference_converter(
-    value: os.PathLike | xr.Dataset | None,
+    value: os.PathLike | xr.Dataset | str | bytes | None,
```
leroyvn, line 121: We have a `PathLike`  type alias in the `eradiate.typing` module to make this a little more concise:

```suggestion
    value: PathLike | xr.Dataset | None,
```

tests/03_regression/atmospheres/test_rpv_afgl1986_continental.py
```diff
@@ -35,6 +36,8 @@ def test_rpv_afgl1986_continental_brfpp(
     )
     result = eradiate.run(exp, spp=10000)
 
+    logger.info(result._repr_html_(), html=True)
```
leroyvn, line 39: That's nice!

tests/conftest.py
```diff
@@ -70,6 +82,20 @@ def pytest_configure(config):
         "(deselect with -m 'not regression')",
     )
 
+    # Disable Eradiate and Mitsuba's progressbars
+    eradiate.config.settings.progress = 0
+
+    # Mitsuba's logger is configured to be silent, unless Pytest's configuration
+    # explicitely requires an increased verbosity (e.g, passing -v option to its
```
leroyvn, line 89:
```suggestion
    # explicitly requires increased verbosity (e.g., passing the -v option to its
```

tests/conftest.py
```diff
@@ -70,6 +82,20 @@ def pytest_configure(config):
...
+    # CLI)
+    eradiate.kernel.install_logging()
+    pytest_verbosity = config.get_verbosity()
+    if pytest_verbosity <= 1:
+        logging.getLogger("mitsuba").setLevel(logging.WARNING)
```
leroyvn, line 94: Could we maybe adopt the reverse approach? I.e. assign to the handler a `WARNING`, and increase the log level of Mitsuba and Eradiate if pytest's verbosity is 2 or more?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,62 @@
...
+      - uses: actions/checkout@v4
+      - name: Install Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install .[doc,recommended,test,dev]
+          pip install eradiate-mitsuba
```
leroyvn, line 20: How about using the `kernel` extra to resolve all dependencies in a single pass?
```suggestion
          pip install .[kernel,doc,recommended,test,dev]
```

However, I think we are at risk of this failing if the tested branch uses an unreleased kernel version. Maybe we should hold this until the full CI pipeline is ready for roll out?

src/eradiate/test_tools/regression.py
```diff
@@ -62,17 +75,50 @@ def regression_test_plots(
...
+        axes[0][1].text(
+            0.5, 0.5, f"{metric[0]}: {metric[1]:.4}", horizontalalignment="center"
+        )
 
     plt.tight_layout()
-    plt.savefig(filename)
-    plt.close()
+
+
+def render_svg_chart():
```
leroyvn, line 94: I think this function should take a `Figure` instance as an input.

leroyvn (changes requested):
Hi @wint3ria, thanks a lot for this PR. I have a few comments, but in general, this transition to robot-based reporting looks very good.

leroyvn:
> UTC deprecation warnings

I remember this coming from one of our dependencies. If this is the case, let's just ignore the warning in pytest.

> issue with the datasets

Our routine tests should not require large data, unless we can persist it. We should sit down together and figure out the best solution.

src/eradiate/scenes/spectra/_uniform.py
wint3ria: The former version was causing an error in the test suite only when using Robot. I must admit I forgot about this issue and should not have committed the change. Do you know a [piece of documentation](https://docs.python.org/3/reference/lexical_analysis.html#formatted-string-literals) about this format specifier?

Here is a traceback for `tests/01_unit/scenes/spectra/test_uniform.py`. Note that the test suite fails at initialization, it is not an individual test failure. Using the option `-p no:robotframework` results in the test suite executing successfully, indeed wo robot...

```
================================================================= test session starts ==================================================================
platform linux -- Python 3.12.9, pytest-8.3.4, pluggy-1.5.0
rootdir: /home/msp/git/eradiate
configfile: pyproject.toml
plugins: pytest_robotframework-4.3.0, xdist-3.6.1, metadata-3.1.1, json-report-1.5.0, anyio-4.8.0
collected 76 items                                                                                                                                     
INTERNALERROR> Traceback (most recent call last):
INTERNALERROR>   File "/home/msp/.conda/envs/erd-robot/lib/python3.12/site-packages/robot/running/builder/parsers.py", line 160, in _parse
INTERNALERROR>     suite = method(source, defaults) if accepts_defaults else method(source)
INTERNALERROR>             ^^^^^^^^^^^^^^^^^^^^^^^^
INTERNALERROR>   File "/home/msp/.conda/envs/erd-robot/lib/python3.12/site-packages/pytest_robotframework/_internal/robot/listeners_and_suite_visitors.py", line 158, in parse
INTERNALERROR>     ":".join([
INTERNALERROR>              ^
INTERNALERROR>   File "/home/msp/.conda/envs/erd-robot/lib/python3.12/site-packages/pytest_robotframework/_internal/robot/listeners_and_suite_visitors.py", line 160, in <genexpr>
INTERNALERROR>     *(str(arg) for arg in cast(tuple[object, ...], marker.args)),
INTERNALERROR>       ^^^^^^^^
INTERNALERROR>   File "<attrs generated repr eradiate.scenes.spectra._uniform.UniformSpectrum>", line 13, in __repr__
INTERNALERROR>     return f'{self.__class__.__qualname__.rsplit(">.", 1)[-1]}(id={self.id!r}, quantity={quantity_repr(self.quantity)}, value={value_repr(self.value)})'
INTERNALERROR>                                                                                                                                ^^^^^^^^^^^^^^^^^^^^^^
INTERNALERROR>   File "/home/msp/git/eradiate/src/eradiate/scenes/spectra/_uniform.py", line 26, in <lambda>
INTERNALERROR>     repr=lambda x: f"{x:~P}",
INTERNALERROR>                      ^^^^^^
INTERNALERROR> ValueError: Invalid format specifier '~P' for object of type 'float'
INTERNALERROR> 
INTERNALERROR> During handling of the above exception, another exception occurred:
INTERNALERROR> 
INTERNALERROR> Traceback (most recent call last):
INTERNALERROR>   File "/home/msp/.conda/envs/erd-robot/lib/python3.12/site-packages/robot/running/builder/builders.py", line 256, in _build_suite_file
INTERNALERROR>     suite = parser.parse_suite_file(source, defaults)
INTERNALERROR>             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
INTERNALERROR>   File "/home/msp/.conda/envs/erd-robot/lib/python3.12/site-packages/robot/running/builder/parsers.py", line 146, in parse_suite_file
INTERNALERROR>     return self._parse(self.parser.parse, source, defaults)
INTERNALERROR>            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
INTERNALERROR>   File "/home/msp/.conda/envs/erd-robot/lib/python3.12/site-packages/robot/running/builder/parsers.py", line 166, in _parse
INTERNALERROR>     raise DataError(f"Calling '{self.name}.{method_name}()' failed: "
INTERNALERROR> robot.errors.DataError: Calling 'PythonParser.parse()' failed: ValueError: Invalid format specifier '~P' for object of type 'float'
INTERNALERROR> 
INTERNALERROR> During handling of the above exception, another exception occurred:
INTERNALERROR> 
INTERNALERROR> Traceback (most recent call last):
INTERNALERROR>   File "/home/msp/.conda/envs/erd-robot/lib/python3.12/site-packages/_pytest/main.py", line 283, in wrap_session
INTERNALERROR>     session.exitstatus = doit(config, session) or 0
INTERNALERROR>                          ^^^^^^^^^^^^^^^^^^^^^
INTERNALERROR>   File "/home/msp/.conda/envs/erd-robot/lib/python3.12/site-packages/_pytest/main.py", line 337, in _main
INTERNALERROR>     config.hook.pytest_runtestloop(session=session)
INTERNALERROR>   File "/home/msp/.conda/envs/erd-robot/lib/python3.12/site-packages/pluggy/_hooks.py", line 513, in __call__
INTERNALERROR>     return self._hookexec(self.name, self._hookimpls.copy(), kwargs, firstresult)
INTERNALERROR>            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
INTERNALERROR>   File "/home/msp/.conda/envs/erd-robot/lib/python3.12/site-packages/pluggy/_manager.py", line 120, in _hookexec
INTERNALERROR>     return self._inner_hookexec(hook_name, methods, kwargs, firstresult)
INTERNALERROR>            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
INTERNALERROR>   File "/home/msp/.conda/envs/erd-robot/lib/python3.12/site-packages/pluggy/_callers.py", line 139, in _multicall
INTERNALERROR>     raise exception.with_traceback(exception.__traceback__)
INTERNALERROR>   File "/home/msp/.conda/envs/erd-robot/lib/python3.12/site-packages/pluggy/_callers.py", line 122, in _multicall
INTERNALERROR>     teardown.throw(exception)  # type: ignore[union-attr]
INTERNALERROR>     ^^^^^^^^^^^^^^^^^^^^^^^^^
INTERNALERROR>   File "/home/msp/.conda/envs/erd-robot/lib/python3.12/site-packages/_pytest/logging.py", line 803, in pytest_runtestloop
INTERNALERROR>     return (yield)  # Run all the tests.
INTERNALERROR>             ^^^^^
INTERNALERROR>   File "/home/msp/.conda/envs/erd-robot/lib/python3.12/site-packages/pluggy/_callers.py", line 122, in _multicall
INTERNALERROR>     teardown.throw(exception)  # type: ignore[union-attr]
INTERNALERROR>     ^^^^^^^^^^^^^^^^^^^^^^^^^
INTERNALERROR>   File "/home/msp/.conda/envs/erd-robot/lib/python3.12/site-packages/_pytest/terminal.py", line 673, in pytest_runtestloop
INTERNALERROR>     result = yield
INTERNALERROR>              ^^^^^
INTERNALERROR>   File "/home/msp/.conda/envs/erd-robot/lib/python3.12/site-packages/pluggy/_callers.py", line 103, in _multicall
INTERNALERROR>     res = hook_impl.function(*args)
INTERNALERROR>           ^^^^^^^^^^^^^^^^^^^^^^^^^
INTERNALERROR>   File "/home/msp/.conda/envs/erd-robot/lib/python3.12/site-packages/pytest_robotframework/_internal/pytest/plugin.py", line 652, in pytest_runtestloop
INTERNALERROR>     _robot_run_tests(session)
INTERNALERROR>   File "/home/msp/.conda/envs/erd-robot/lib/python3.12/site-packages/pytest_robotframework/_internal/pytest/plugin.py", line 374, in _robot_run_tests
INTERNALERROR>     _ = _run_robot(session, robot_options)
INTERNALERROR>         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
INTERNALERROR>   File "/home/msp/.conda/envs/erd-robot/lib/python3.12/site-packages/pytest_robotframework/_internal/pytest/plugin.py", line 292, in _run_robot
INTERNALERROR>     robot.main(  # pyright:ignore[reportUnknownMemberType]
INTERNALERROR>   File "/home/msp/.conda/envs/erd-robot/lib/python3.12/site-packages/robot/run.py", line 462, in main
INTERNALERROR>     suite = builder.build(*datasources)
INTERNALERROR>             ^^^^^^^^^^^^^^^^^^^^^^^^^^^
INTERNALERROR>   File "/home/msp/.conda/envs/erd-robot/lib/python3.12/site-packages/robot/running/builder/builders.py", line 158, in build
INTERNALERROR>     self.rpa).parse(structure)
INTERNALERROR>               ^^^^^^^^^^^^^^^^
INTERNALERROR>   File "/home/msp/.conda/envs/erd-robot/lib/python3.12/site-packages/robot/running/builder/builders.py", line 218, in parse
INTERNALERROR>     structure.visit(self)
INTERNALERROR>   File "/home/msp/.conda/envs/erd-robot/lib/python3.12/site-packages/robot/parsing/suitestructure.py", line 65, in visit
INTERNALERROR>     visitor.visit_file(self)
INTERNALERROR>   File "/home/msp/.conda/envs/erd-robot/lib/python3.12/site-packages/robot/running/builder/builders.py", line 223, in visit_file
INTERNALERROR>     suite = self._build_suite_file(structure)
INTERNALERROR>             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
INTERNALERROR>   File "/home/msp/.conda/envs/erd-robot/lib/python3.12/site-packages/robot/running/builder/builders.py", line 260, in _build_suite_file
INTERNALERROR>     raise DataError(f"Parsing '{source}' failed: {err.message}")
INTERNALERROR> robot.errors.DataError: Parsing '/home/msp/git/eradiate/tests/01_unit/scenes/spectra/test_uniform.py' failed: Calling 'PythonParser.parse()' failed: ValueError: Invalid format specifier '~P' for object of type 'float'
```

src/eradiate/test_tools/regression.py
wint3ria: Good point, I'll rework this

tests/03_regression/atmospheres/test_rpv_afgl1986_continental.py
wint3ria: Yes! Suggestions are welcome in order to add additional charts and visual constructs to the report. I believe this should be done per test, case by case, and not systematically. We could also encourage the design of meaningful debug charts when developers submit new tests

tests/conftest.py
wint3ria: For confirmation: assign warning loglevel to the default handler or the entire logger?

.github/workflows/ci.yml
wint3ria: Should we actually merge a version that is built on a unreleased kernel? Shouldn't we let such resulting CI actions fail as a marker of their unmerged development status? Your formulation is more elegant

wint3ria:
> > issue with the datasets
> 
> Our routine tests should not require large data, unless we can persist it. We should sit down together and figure out the best solution.

Sure

src/eradiate/scenes/spectra/_uniform.py
leroyvn: Yes, [Pint documents this](https://pint.readthedocs.io/en/stable/user/formatting.html). The traceback seems to imply that the problem is that some of the tests, the fact that `x` is a `float` prevents the use of the formatted. I fixed it on `main` (and `next`).

tests/conftest.py
leroyvn: Correct me if I'm wrong—we have only one handler, right? In that case, it should make a difference, so we can assign the warning level to the logger itself.

.github/workflows/ci.yml
leroyvn: Only releases require a wheel; even `main` might have stable, but not yet released, kernel code. So I think it's legit to build the kernel as part of CI.

Anyway, I think it's more appropriate to split this PR between separate CI and testing tooling parts (basically, move the CI update commits to a separate branch) (see my general comment).

leroyvn:
Hi @wint3ria, I finally had time to get back to this. In addition to the minor changes that are still pending, I suggest splitting this PR into separate CI and testing tooling parts. The CI will come with proper CI deployment, while this one will only be about the framework.

src/eradiate/scenes/spectra/_uniform.py
wint3ria: Great, thank you

wint3ria:
> Hi @wint3ria, I finally had time to get back to this. In addition to the minor changes that are still pending, I suggest splitting this PR into separate CI and testing tooling parts. The CI will come with proper CI deployment, while this one will only be about the framework.

I'll split it. Thanks for the review :)

tests/conftest.py
wint3ria: Moved to the CI PR #482

tests/conftest.py
wint3ria: Moved to the CI PR #482

.github/workflows/ci.yml
wint3ria: Moved to CI PR #482

tests/conftest.py
wint3ria: > Correct me if I'm wrong—we have only one handler, right? In that case, it should make a difference, so we can assign the warning level to the logger itself.

I set the level using the loggers because I really don't want any handler to pollute my robot logs at that point. We could consider that we should allow users to configure their own handlers and mess change the setup even in pytest/robot. I asked for the precision because you used the term handler 

> Could we maybe adopt the reverse approach? I.e. assign to the handler a `WARNING`, and increase the log level of Mitsuba and Eradiate if pytest's verbosity is 2 or more?

I proposed some changes on the CI branch, let me know

wint3ria:
@leroyvn This is really close from ready. As requested, a part of the PR has been moved to #482

docs/src/release_notes/v0.30.x.md
```diff
@@ -31,7 +41,7 @@ available as `mycena_v1`.
 
 * Added several ocean BRDF models: 6SV-like / "legacy" ({ghpr}`453`),
   specular polarized / "Mishchenko" ({ghpr}`466`), GRASP ({ghpr}`480`).
-* 🖥️ Improved importance sampling for ocean BRDFs ({ghpr`480`}).
+* 🖥️ Improved importance sampling for ocean BRDFs ({ghpr `480`}).
```
leroyvn, line 44:
```suggestion
* 🖥️ Improved importance sampling for ocean BRDFs ({ghpr}`480`).
```

src/eradiate/test_tools/regression.py
```diff
@@ -22,33 +24,44 @@ def regression_test_plots(
     ref: np.typing.ArrayLike,
     result: np.typing.ArrayLike,
     vza: np.typing.ArrayLike,
-    filename: PathLike,
     metric: tuple[str, float],
+    ref_var: t.Optional[np.typing.ArrayLike] = None,
+    result_var: t.Optional[np.typing.ArrayLike] = None,
 ) -> None:
```
leroyvn, line 30: You should change the return type (and associated docstring entry).

src/eradiate/test_tools/regression.py
```diff
@@ -62,17 +75,68 @@ def regression_test_plots(
...
+    Parameters
+    ----------
+    fig : plt.Figure
+        Matplotlib figure to render in HTML.
+
+    Returns
+    -------
+    str
+        Rendered HTML <svg> tag with styling.
+
```
leroyvn, line 109:
```suggestion
```

src/eradiate/test_tools/regression.py
```diff
@@ -62,17 +75,68 @@ def regression_test_plots(
...
+        )
+    else:
+        axes[0][1].text(
+            0.5, 0.5, f"{metric[0]}: {metric[1]:.4}", horizontalalignment="center"
+        )
+
+    return fig, axes
+
+
+def figure_to_html(fig: plt.Figure):
```
leroyvn, line 94:
```suggestion
def figure_to_html(fig: plt.Figure) -> str:
```

src/eradiate/test_tools/regression.py
```diff
@@ -213,21 +300,40 @@ def run(self) -> bool:
         # if no valid reference is found, store the results as new ref and fail
         # the test
         if not self.reference:
-            print("No reference data was found! Storing test results as reference.")
+            logger.info(
```
leroyvn, line 303: Would it be appropriate to raise the log level to a warning? (I actually do not know if that would also result in a warning being emitted)

leroyvn (commented):
Hi @wint3ria, thanks. Almost all good. I made a few minor comments and suggestions; then we can merge.

src/eradiate/test_tools/regression.py
wint3ria: There is here a limitation of the Robot framework logger: the `info` function of the logger is the only one allowing to also write on the console. I could write the following, but I find it a bit cumbersome:

```
warning_message = f"No reference data was found! Storing test results as reference to {fname_reference}"
logger.warn(warning_message)
logger.console(warning_message)
```

The above code would result in a warning in the Robot report, but a console message in the plain Pytest suite. 

We could also provide a separate logger for our tests, allowing more flexibility on the provided logging features, I considered it a bit far fetched for the scope of this PR at first.

We could also issue a warning, and decide to fail the test suite on this specific warning or to ignore it in the pytest configuration.

Let me know your preferred solution here.

leroyvn:
Hi @wint3ria, I finally had time to check out the code and play with it a bit. Some fixes are needed (e.g. missing variable names in regression tests); but also, and more importantly, I don't have the feeling that both robot and no-robot modes are 1st-class citizens at the moment. 

What I mean is that if you are debugging with robot deactivated, robot is still used for output, which makes it silent. I'd like to dynamically route logs to either the pretty report or just stdout depending on whether robot is detected. This is no easy task, but I've make a few experiments which, hopefully, will put us on the right track.

leroyvn:
I made the following additional changes:

* A few fixes that most likely slipped through during rebase
* Made robot opt-in so that the default behaviour remains as close to the current one as possible
* Added a couple of fixtures to allow detecting robot during a pytest session. This is useful, e.g. to change the output when not in robot mode (notably thinking about plots). It's not exploited yet, but it's there.
* Streamlined `pyproject.toml` contents a bit.

With that, I think we're good to go. Thanks!

PR #482, opened 2025-04-04
CI and Automatic report

wint3ria:
# Description

Add a CI step to launch routine tests and produce a minimal report when pushing. Based on #483 .
When a PR is marked as ready for review, an additional, longer test suite is executed.

This CI builds the Eradiate kernel and acesses the binaries as in a development environment.

This is the first run using ARS, which updates the current working sha of the PR with a "check-run", the Github status displayed on this page. This status badge gives information about the procedure to provision a VM suitable to run a given job/workflow.

The kernel build is cached, using the github ref of the submodule as the main criterion for flushing the binaries.

## Performances

Using Github's hosted workers, the Mitsuba build takes roughly 35 min.
Using the ARS spot instance `ubuntu_x86_64_eradiatecompile_spot`, compilation is reduced to ~10min. 
All the tests taking more than 1 min to complete have been removed from the routine CI, reducing the completion time from 1h to 12min. The importance of some disabled tests can be discussed.
Overall, when recompiling the kernel on ARS, the entire routine CI takes roughly 25min to complete. 
The overhead from launching a VM in ARS is barely visible.

## Possible improvements

The test suite does not benefit well from extra CPUs available, because pytest execution is sequential. ~Without extra benefits from running the routine tests of the CI in larger machines, these tests have been moved back to a Github-hosted runner.~ I can't continue testing the caching feature since the tests are failing, preventing me from validating any new cache version. 

Open to suggestions for the VM types naming convention.

# Checklist

- [x] The code follows the relevant coding guidelines
- [x] The code generates no new warnings
- [x] The code is appropriately documented
- [x] The code is tested to prove its function
- [x] The feature branch is rebased on the current state of the `main` branch
- [x] I updated the change log if relevant
- [x] I give permission that the Eradiate project may redistribute my contributions under the terms of its license

PR #483, opened 2025-04-04
Stat tests

wint3ria:
# Description

Based on #478 

Create a set of additional regression tests using statistical tests on bias-variance ratios:
 - IndependantStudentTTest
 - PairedStudentTTest
 - ZTest
 - SidakTTest

Implement additional diagnostic charts in regression tests. These charts are meant for debugging the tests themselves, or to further investigate the results of a given test.

An abstract classes is also provided: `AbstractStudentTTest` implements the `_diagnostic_chart` method for the `IndependentStudentTTest` and `PairedStudentTTest`.

Compatible with the Robot framework and #478 in particular.

# Checklist


wint3ria:
@leroyvn I took some decisions with this PR, I am interested in your thoughts on the overall setup

leroyvn (commented):
Hi @wint3ria, thanks for this update. I made a few adjustments to the code (mostly style-related):

- Matplotlib code should use the object API
- I removed the CI workflow since it is part of #482
- I fixed some broken tests and aligned ruff in the CI linter and the pre-commit hook

The major point now is the regression testing framework benchmarking code: Since you wrote this as a thorough benchmarking suite not meant to be run routinely as part of CI, I suggest moving this to the resource directory. It will still be available to check the behaviour of the test suite, but it won't be collected by pytest.

leroyvn:
I think we are there now. I migrated our regression tests to the new framework (including reference updates when relevant). This is good to go, I'll merge now.

PR #485, opened 2025-04-28
CI and Automatic report (rebased)

leroyvn:
# Description

*This PR supersedes #482. The following description is the one originally written by @wint3ria.*

Add a CI step to launch routine tests and produce a minimal report when pushing. Based on #483 .
When a PR is marked as ready for review, an additional, longer test suite is executed.

This CI builds the Eradiate kernel and acesses the binaries as in a development environment.

This is the first run using ARS, which updates the current working sha of the PR with a "check-run", the Github status displayed on this page. This status badge gives information about the procedure to provision a VM suitable to run a given job/workflow.

The kernel build is cached, using the github ref of the submodule as the main criterion for flushing the binaries.

## Performances

Using Github's hosted workers, the Mitsuba build takes roughly 35 min.
Using the ARS spot instance `ubuntu_x86_64_eradiatecompile_spot`, compilation is reduced to ~10min. 
All the tests taking more than 1 min to complete have been removed from the routine CI, reducing the completion time from 1h to 12min. The importance of some disabled tests can be discussed.
Overall, when recompiling the kernel on ARS, the entire routine CI takes roughly 25min to complete. 
The overhead from launching a VM in ARS is barely visible.

## Possible improvements

The test suite does not benefit well from extra CPUs available, because pytest execution is sequential. ~Without extra benefits from running the routine tests of the CI in larger machines, these tests have been moved back to a Github-hosted runner.~ I can't continue testing the caching feature since the tests are failing, preventing me from validating any new cache version. 

Open to suggestions for the VM types naming convention.

# Checklist


leroyvn (commented):
Hi @wint3ria, those changes are overall and I think we are close to the final workflow.

My main comment at this point is that since the fast and full CI pipelines are almost identical, it would make sense IMO to merge them into a single pipeline split into different jobs that will execute conditionally and depend on each other (using the [`needs`](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idneeds) keyword). I could see the following jobs:

* `build` (always runs, caches its output; indexing the cache by the submodule build is indeed reliable)
* `test_fast` (needs `build`, runs on draft PRs)
* `test_full` (needs `build`, runs on non-draft PRs and all commits to `next` and `main`; why not use the cached build as well?)
* `report` (needs `test_fast` or `test_full`; nice to have because it's almost half of the pipeline code; that however implies uploading and downloading artefacts)

A question: I see `runs-on: [self-hosted, ars, ubuntu_x86_64_eradiatecompile_spot]`: what does it mean in practice? Is a runner type chosen by the pipeline automatically, or is it something else?

leroyvn:
FWIW, I lazily asked Claude AI to generate some pipeline code to implement that behaviour and I got this (I didn't check if it runs successfully, it might be entirely wrong):

```yaml
name: Multi-Stage Pipeline

on:
  push:
    branches: [ main ]
  pull_request:
    branches: [ main ]

jobs:
  job1:
    name: Base Job
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v3
      - name: Run job1 tasks
        run: echo "Running job1"

  job2:
    name: Draft PR Job
    needs: job1
    if: github.event_name == 'pull_request' && github.event.pull_request.draft == true
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v3
      - name: Run job2 tasks
        run: echo "Running job2 for draft PR"

  job3:
    name: Main/Non-Draft PR Job
    needs: job1
    if: (github.event_name == 'pull_request' && github.event.pull_request.draft == false) || github.ref == 'refs/heads/main'
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v3
      - name: Run job3 tasks
        run: echo "Running job3 for non-draft PR or main branch"

  job4:
    name: Final Job
    needs: [job2, job3]
    if: success() && (needs.job2.result == 'success' || needs.job3.result == 'success')
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v3
      - name: Run job4 tasks
        run: echo "Running job4 after successful job2 or job3"
```

wint3ria:
I do not really get why the readthedocs check fails. 

Did you address the failing tests, or can I do it when I come back?

We need to think about an ARS configuration that fits your requirements for VM size and budget. This includes giving you the required rights to use the current deployment. As already mentioned in the copypasta, the naming convention is open to discussion.

Concerning the extra workflow, I wanted to have a separate badge for ready-to-merge PRs, but it's a detail. Given sufficient testing of Claude's version, we could consider integrating it. Could we have Claude factorizing this in a separate standalone action instead, to have the best of both approaches ? If practical, I can vibe-code that when I coming back as well

wint3ria:
I updated ARS doc, have a look and let me know if something isn't clear there

wint3ria:
> > Concerning the extra workflow, I wanted to have a separate badge for ready-to-merge PRs, but it's a detail. Given sufficient testing of Claude's version, we could consider integrating it. Could we have Claude factorizing this in a separate standalone action instead, to have the best of both approaches ? If practical, I can vibe-code that when I come back as well
> 
> I suppose we can factor things as [local actions]. My main motivation for having dependent jobs is that they make visualizing the pipeline more convenient (note: the publishing workflow could use some improvements, like skipping unused jobs)

Implemented, alongside minor changes.

Let me know your thoughts on the last few open points.

.github/workflows/cd.yml
```diff
@@ -1,7 +1,7 @@
 # See PyPA packaging guide:
 # https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/
 
-name: Publish to PyPI
+name: Eradiate delivery
```
leroyvn, line 4: For conciseness (see dashboard and graph) I'd prefer to keep the previous one:
```suggestion
name: Publish to PyPI
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,89 @@
...
+
+env:
+  CC: clang-17
+  CXX: clang++-17
+  ERADIATE_SOURCE_DIR: ${{ github.workspace }}
+  PYTHONPATH: ${{ github.workspace }}/ext/mitsuba/build/python
+
+jobs:
+  compile:
+    name: Compile Eradiate
```
leroyvn, line 19:
```suggestion
    name: Build kernel
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,89 @@
...
+      - id: prepare
+        uses: ./.github/actions/prepare-ci
+      - name: Build kernel
+        if: steps.prepare.outputs.mitsuba-cache-hit != 'true'
+        run: |
+          cmake -S ext/mitsuba -B ext/mitsuba/build -DCMAKE_BUILD_TYPE=Release -GNinja --preset eradiate
+          cmake --build ext/mitsuba/build
+
+  test:
+    name: Execute routine tests
```
leroyvn, line 35:
```suggestion
    name: Run tests (routine only)
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,89 @@
...
+      - name: Robot report
+        if: always()
+        run: |
+          rebot --tagstatinclude unit --tagstatinclude system --tagstatinclude regression --tagstatinclude slow --expandkeywords name:* ./output.xml
+          chrome --no-sandbox --virtual-time-budget=10000 --headless --print-to-pdf=./log.pdf    ./log.html    --no-pdf-header-footer
+          chrome --no-sandbox --virtual-time-budget=10000 --headless --print-to-pdf=./report.pdf ./report.html --no-pdf-header-footer
+      - uses: ./.github/actions/save-ci-logs
+
+  report:
+    name: Execute all tests
```
leroyvn, line 64:
```suggestion
    name: Run tests (all)
```

.github/actions/prepare-ci/action.yml
```diff
@@ -0,0 +1,36 @@
...
+    uses: actions/setup-python@v5
+    with:
+      python-version: "3.12"
+  - name: Install Chrome
+    uses: browser-actions/setup-chrome@v1
+  - name: Install dependencies
+    shell: bash
+    run: |
+      python -m pip install --upgrade pip
+      pip install .[doc,recommended,test,dev]
```
leroyvn, line 21: Something we might want to discuss (not here, let's focus on the pipeline for this PR): this does not use a reproducible testing env. Whether the testing env should be frozen has pros and cons, and I don't have a settled opinion on the matter.

leroyvn (commented):
Hi @wint3ria, we're almost there. I have a couple of additional comments, small details.

.github/actions/prepare-ci/action.yml
wint3ria: Sure

leroyvn:
@wint3ria maybe we can check how things look like once rebased onto #488?

tests/01_unit/experiments/test_atmosphere.py
```diff
@@ -166,6 +166,7 @@ def test_atmosphere_experiment_real_life(
     assert mi_sensors["radiancemeter"].medium().id() == "medium_atmosphere"
 
 
+@pytest.mark.slow
```
wint3ria, line 169: We need to review the performance of these tests and remove the mark if applicable.

tests/01_unit/experiments/test_atmosphere.py
```diff
@@ -190,6 +191,7 @@ def test_atmosphere_experiment_run_basic(
     assert isinstance(eradiate.run(exp, spp=100), xr.Dataset)
 
 
+@pytest.mark.slow
```
wint3ria, line 194: Same here

tests/01_unit/experiments/test_atmosphere.py
```diff
@@ -252,6 +254,7 @@ def test_atmosphere_experiment_run_detailed(
     assert np.all(results["radiance"].data > 0.0)
 
 
+@pytest.mark.slow
```
wint3ria, line 257: Same here

tests/02_system/test_heterogeneous_atmosphere_flags.py
```diff
@@ -5,6 +5,7 @@
 from eradiate import unit_registry as ureg
 
 
+@pytest.mark.slow
```
wint3ria, line 8: Same here

wint3ria (changes requested):
I wrongfully marked some tests as slow due to the regression related to performance issues caused by the usage of the latest xarray version. We need to review these tests and decide which ones should be marked

wint3ria:
>> @wint3ria maybe we can check how things look like once rebased onto #488?
>
> I wrongfully marked some tests as slow due to the regression related to performance issues caused by the usage of the latest xarray version. We need to review these tests and decide which ones should be marked

I will do that in a moment

wint3ria:
The vCPU count in ARS VM configs has been doubled.

leroyvn:
Thanks! Let's merge now 🚀

PR #493, opened 2025-06-09
Data delivery

leroyvn:
# Description

This PR refactors the data delivery and handling system. This is a major modification that will potentially impact users. We provide guidelines for migrating to the new interface. These changes overall reduce the technical debt induced by data handling. Most importantly, on-disk file lookups are now decoupled from data delivery.

Changes are as follows:

* The `DataStore` abstraction, which used to combine local and online file lookups, is removed and replaced by two separate components: the `FileResolver` handles local lookups, and the `AssetManager` handles online data.
* The new `FileResolver` maintains a list of absolute paths on the hard drive and uses them to resolve local paths. It should be used to register both default and user data locations, which can then be looked up in a unified way. Path ordering implicitly defines priorities among the tracked locations.
* The new `AssetManager` is a lightweight package manager that can download and unpack archives (*resources*) from an online location. Disk usage can be controlled precisely: cached, unpacked, and installed data can be selectively deleted per package. The installation process uses symlinks and avoids data duplication.
* All data registry management components have been removed.
* The data CLI has been entirely rewritten.
* Eradiate has two unique instances of the `FileResolver` and `AssetManager`. Unless instructed to, the unique `AssetManager` does nothing and stays offline. The unique `FileResolver` also works entirely offline.

While this refactoring action was progressing, the following changes were also made:

* Outdated settings have been removed, and new settings have been introduced to support the configuration of the new data delivery system.
* The `config` subpackage has been refactored, with defaults now built-in rather than loaded from a settings file. All settings now have validation and conversion protocols. The `ENV` variable was removed because it has never been used. The `SOURCE_DIR` variable now just serves the corresponding settings parameter. These changes are motivated by the need to improve the consistency and user-friendliness of the `config.settings` member. Additional refactoring of the settings interface will happen in the future, in particular to have some parameter updates cascade down when they are used to initialize global objects (*e.g.* the `DATA_PATH` or `DATA_URL` variables).
* The `scenes.biosphere.canopies` submodule, which provides functionality entirely covered by the canopy scene loader, was removed.
* To minimize the amount of downloaded data during tests, the default monochromatic dataset is now `komodo`.
* Converters were refactored to provide a more modular data conversion toolkit.
* The canopy loader was updated and still supports automatic downloads through the Pooch library. This situation is temporary until all data is migrated to the new data registry. 
* Development requirements are now specified using [PEP 735](https://peps.python.org/pep-0735/) dependency groups instead of extras. This makes our `pyproject.toml` more portable, and avoids cluttering our extra list with entries irrelevant to end-users. This bumps the minimal required Pixi version to v0.38.
* CLI terminal display components have been factored into a small module to avoid code duplication.

To do:

* [x] Propagate the new data management paradigm to the rest of the code base
  * [x] Remove all references to the data store
  * [x] Use the file resolver
  * [x] Fix the canopy loader
* [x] Deploy officially the new data registry
  * [x] Freeze the new root URL
  * [x] Freeze the manifest format
  * [x] Freeze the data layout
* [x] Documentation
  * [x] Update docstrings
  * [x] Update API docs
  * [x] Rewrite relevant data guide entries
  * [x] Update configuration guide
* [x] ~~Fix CI~~ *CI now runs properly, but one z-test fails; this is not reproducible on my dev setup: we will investigate this as a separate issue*
* [x] Clean up data repository
* [x] Add tests
  * [x] New converters
  * [x] ~~Asset manager~~
  * [x] File resolver

# Checklist

### DIFF
--- a/docs/rst/data/intro.rst
+++ b/docs/rst/data/intro.rst
@@ -99,7 +99,11 @@ referencing the target resources by their IDs, *e.g.*:
 
 Resource archives that are not already available locally will be downloaded from
 the remote data registry. They will be unpacked and linked to the version-
-dependent installation directory.
+dependent installation directory. The cache and unpacking locations are, by
+default, common to all Eradiate versions. That means that if you use default
+settings, data that was downloaded with a given version of Eradiate will not be
+downloaded or unpacked again after an upgrade — only the symbolic links created
+in the version-specific installation directory will be created again.
 
 For convenience, some resources are aliased. The list of aliases can be
 displayed as follows:


leroyvn:
@wint3ria this is almost ready for merge, I just have to fix the CI. I think you'll be the most impacted by this since it completely changes how data is shipped.

leroyvn:
We are now ready to merge. The only issue that remains is the failing CI pipeline: one z-test fails, but does not on my dev workstation. I think we can merge this and fix the CI issue in a separate PR or commit.

PR #517, opened 2025-09-25
Test refactoring

wint3ria:
# Description

This PR introduces several new regression tests.

This PR aims at identifying and fixing tests that are either:
 - deprecated
 - poorly scoped
 - insufficient
 - improperly organized
 - unreliable

# Coverage

Coverage has been evaluated before changes introduced in this PR.
Some tooling modules and peripheral utilities report low test coverage:
 - CLI (0%)
 - Benchmark tools (0%)
 - sys-info utility (0%)
 - Data asset manager (24%)
 - Plot utilities (26%)

Average coverage is 70.47%.
If overall coverage can always be improved, the situation is not alarming.

See attached [coverage.json](https://github.com/user-attachments/files/22537064/coverage.json)

# Plot functions

 - Diverse pyplot charts made during tests have been disabled when the suite is not launched for reporting. 
 - They can be explicitly enabled using the "--plot 1" option for pytest.
 - This is done for all tests in the codebase, not just for regression tests.
 - Considering [this snippet](https://github.com/eradiate/eradiate/blob/faf1acdf49916a506093f9b2788e2ecac4512c99/src/eradiate/test_tools/fixtures/_plt.py), at least a few things need to be adjusted: 
    - The current strategy is to "trust" the developer to use a "reasonable" amount of time to perform their plot logic, rather than forcing a behavior. For instance, in the current implementation, the test suite may still perform some plots in case of test failure. We can change this behavior, or refine the method for selecting the plt module (mock one or actual matplotlib.pyplot). I also agree that the current approach has already led to excessive testing time spent in pyplot.
    - By default without additional options passed to pytest, the test suite should perform the plots. Otherwise developers are left with a dep injection by default that is a bit counter intuitive.
    - On the cosmetic side, I would prefer another naming convention than reusing "plt". On the same line, an explicitly specific import logic could be preferred, in the form of a method returning the reference to the module for example.
    - It must be integrated with the report tooling, obviously. This part is the easiest, to the limit that some charts may only be useful to developers, not in the report.
    - The Mock object could directly inspect the signature of the mocked module, such that it could at least detect invalid function calls. Some level of error handling could avoid confusing behaviors.

# System test directory

Tests in the `tests/02_System` directory have been evaluated for relevance. Overall, each test fits within the definition of an integration test. The following comments can be made:
 - test_kernel_render_benchmark seems to be unfinished partial work. Given the unstable nature of the kernel update system for now, this is not surprising. Further modification of this part of the codebase would lead to its update.
 - test_maximum_scene_size seems slightly deprecated?
 - Plots have been disabled by default in this PR.
 - Some test docstrings are a bit verbose, but nothing critical.

Overall, these integration tests are made on the fly and no proper systematic strategy to implement new ones has been thought through to the best of my knowledge. I'm not sure we need to act on this, regression tests such as RAMI4ATM ones cover many possible 1D combinations. 

3D combinations could be considered, including canopies, spherical shell and/or DEMs.

Polarization could be further tested as well.

# Spherical shell regression test

The configuration for the spherical shell regression test is the following:
 - recycled from RAMI4ATM
 - molecular atmosphere AFGL1986 US STD
 - VZA in the principal plane from 85° to 65°
 - MSI band 4

# RAMI4ATM regression tests

A selection of 22 extra cases from RAMI4ATM has been included in the regression tests with associated reference values. A baseline test is done for each case using the TOA BRF in the principal plane only.
For two of the selected cases, one above an RPV surface, another above a white surface, both using an optically thin atmosphere, extra tests are performed on the BOA BHR and HDRF.
All the new RAMI4ATM regression tests are done on S2A MSI band 4.

# Unreliable tests

### ZTest and SidakTTest tools

These to classes have been fixed to always allow at least one of the paired tests to fail.

### tests/03_regression/ocean/test_ocean_grasp.py::test_ocean_grasp_open_atm

This test has been modified to improve its robustness. It benefits from the new possibility to disable regression test plots that are only applicable to single wavelength tests. The proposed solution does not output any chart in the report anymore.

@frettini all failed cases were observed for the wavelength 2200nm, but I have not explored this in details. You may want to update the threshold value as well in order to ensure sufficient power of the test.

# Additional comments

 - Pytest, Robot-framework and pytest-robotframework are used in the Eradiate CI. Inconsistent versions of these packages led to the impossibility to execute pytest independently from the Robot-framework. An update of the packages has been performed.

 - We could set the seed value in a more systematic way for regression tests. It's desirable to forbid test developers from finding the "right" seed for their tests, or to design their tests only for one specific seed. On the other hand, tests should at least be reproducible from the developer's machine to the CI workers.

 - Some comments can be made for the unit tests:

    - 01_unit/kernel/test_grid_volume: only initialization
    - 01_unit/scenes/biosphere/test_discrete: only initialization, not fond of inline PLYs in strings
    - 01_unit/scenes/biosphere/test_leaf_cloud: generated leaf angle distributions could be tested, not found of inline PLYs in strings
    - 01_unit/scenes/bsdfs/* Many BSDF could benefit from a test against a reference
    - 01_unit/scenes/illumination/* Illuminations could benefit from testing against references
    - 01_unit/scenes/integrators/* only initialization
    - 01_unit/scenes/phase/test_gh,test_isotropic,test_rayleigh only initialization
    - 01_unit/scenes/test_loader It requires refactoring, but the loader itself as well.
    - 01_unit/spectral/test_ckd_quad only initialization

Overall, scene elements instantiating a plugin rely on Mitsuba tests for ensuring actual correctness of the element's output.

# Limits

Included homogeneous uniform discrete RAMI4ATM canopy case takes significantly longer to terminate. Without any modification from the setup used in RAMI4ATM, the test case lasts 19min on average. The canopy size has thus been divided by 5, and there is no proper assessment of the impact on recorded reflectances. The resulting compute time is about 45 seconds on a standard development machine at Rayference.

Bottom of atmosphere cases are slower to compute.

# Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -34,7 +34,7 @@ dependencies = [
   "typer>=0.9.0",
   "tqdm",
   "shellingham!=1.5.1", # Typer dependency; exclude this version which raises while previous don't
-  "xarray>=2023,!=2025.1.*,!=2025.3.*,!=2025.4.*,!=2025.5.*,!=2025.6.*",
+  "xarray>=2023,<2025.0.0",
 ]
 
 [project.optional-dependencies]


leroyvn (commented):
Hi @wint3ria, this is a first reaction limited to a code review — broader discussion about how we move forward will happen offline. I made some relatively minor changes which, hopefully, improve the maintainability of the codebase and the UX. The largest is a refactoring of the RAMI4ATM case definitions which had become a bit too redundant IMO. Apart from that, it's good for me :) I also took the liberty to rebase the branch onto the latest `next`.

leroyvn:
Also: We still have a couple of tests that fail: 

* A BOA HDRF which should return 1 and has a bit of noise ⇒ we can use a simpler test criterion
* One test that matches very well the reference but for which the statistical test fails ⇒ worth a check

PR #560, opened 2026-03-16
BandSRF: Add central wavelength computation

leroyvn:
# Description

This PR adds an interface point to retrieve the central wavelength of a `BandSRF`:

```python
srf = BandSRF(...)
srf.central_wavelength()
```

For convenience, the post-processing pipeline also attaches a new `w_srf` **scalar** coordinate to SRF-weighted variables:
image

The choice of promoting it to a dimension coordinate is left to the user, in particular because, as a side effect, this coordinate will appear also for spectral variables, for which it is irrelevant. This is a limitation of xarray, which can be addressed only by changing more deeply the output data structure.

Now, we can concatenate two SRF-weighted variables as follows:
image

I also took the opportunity to refactor the SRF test suite.

Thanks to @pdevis for suggesting this feature.

# Checklist

### DIFF
--- a/docs/release_notes/v1.1.x.md
+++ b/docs/release_notes/v1.1.x.md
@@ -59,6 +59,10 @@
   ({ghpr}`543`).
 * 🖥️ The phase function of the {class}`.MolecularAtmosphere` class can now be
   forced to a user-defined value for debugging purposes ({ghpr}`550`).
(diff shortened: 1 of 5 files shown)


leroyvn:
@pdevis would you mind checking if the added scalar coordinate does what you want?

pdevis:
This looks great! Thanks for adding this feature!

PR #565, opened 2026-04-01
Add new `EOVolPathIntegrator` interface

frettini:
# Description

This PR includes the interface to the new `EOVolPathIntegrator`. It is a copy of the `VolPathIntegrator` onto which we can apply our changes, easing the process of rebasing onto latest versions of Mitsuba 3. As part of this PR, the integrator implements the DDIS variance reduction technique. As such, the interface exposes parametrization of the ddis threshold. The Russian Roulette was also found to terminate photons too quickly in presence of large optical thickness. By default, `EOVolPathIntegrator` has `ddis_threshold=0.1` and `rr_depth=1000`. 

Note that  `EOVolPathIntegrator` will be subject to changes in the coming PRs, as such it is not used as default integrator in experiments.

Regression test were added to compare `VolPathIntegrator` and `EOVolPathIntegrator`.

# Checklist

### DIFF
--- a/src/eradiate/scenes/integrators/_path_tracers.py
+++ b/src/eradiate/scenes/integrators/_path_tracers.py
@@ -28,13 +28,11 @@ class MonteCarloIntegrator(Integrator):
         init_type="int, optional",
     )
 
-    rr_depth: int | None = documented(
-        attrs.field(default=None, converter=attrs.converters.optional(int)),
+    rr_depth: int = documented(
+        attrs.field(default=5, converter=int),
         doc="Minimum path depth after which the implementation starts applying "
-        "the Russian roulette path termination criterion. If unset, the kernel "
-        "default value (5) is used.",
-        type="int or None",
-        init_type="int, optional",
+        "the Russian roulette path termination criterion.",
+        type="int",
     )
 
     hide_emitters: bool | None = documented(
@@ -56,14 +54,15 @@ class MonteCarloIntegrator(Integrator):
         Override this method in subclasses to add integrator-specific parameters.
         The base implementation handles common Monte Carlo integrator parameters.
         """
-        result = {"type": self.kernel_type}
+        result = {
+            "type": self.kernel_type,
+            "rr_depth": self.rr_depth,
+        }
 
         if self.timeout is not None:
             result["timeout"] = self.timeout
         if self.max_depth is not None:
             result["max_depth"] = self.max_depth
-        if self.rr_depth is not None:
-            result["rr_depth"] = self.rr_depth
         if self.hide_emitters is not None:
             result["hide_emitters"] = self.hide_emitters
 
@@ -130,31 +129,41 @@ class EOVolPathIntegrator(MonteCarloIntegrator):
 
     This integrator samples paths using random walks starting from the sensor.
     It supports multiple scattering, accounts for volume interactions, and
-    implements the DDIS variance reduction method.
+    implements the DDIS variance reduction method as described in
+    :cite:t:`Buras2011EfficientUnbiasedVariance`.
     """
 
-    rr_depth: int | None = documented(
-        attrs.field(default=1000, converter=attrs.converters.optional(int)),
+    rr_depth: int = documented(
+        attrs.field(
+            default=1000,
+            converter=int,
+            validator=attrs.validators.instance_of(int),
+        ),
         doc="Minimum path depth after which the implementation starts applying "
         "the Russian roulette path termination criterion.",
-        type="int or None",
-        init_type="int, optional",
+        type="int",
     )
 
     rr_factor = documented(
-        attrs.field(default=0.97, converter=attrs.converters.optional(float)),
+        attrs.field(
+            default=0.97,
+            converter=float,
+            validator=attrs.validators.instance_of(float),
+        ),
         doc="Specifies the maximum probability to keep a path when Russian "
         "Roulette is evaluated.",
         type="float",
-        init_type="float",
     )
 
     ddis_threshold = documented(
-        attrs.field(default=0.1, converter=attrs.converters.optional(float)),
+        attrs.field(
+            default=0.1,
+            converter=attrs.converters.optional(float),
+            validator=attrs.validators.instance_of(float),
+        ),
         doc="Specifies the probability to importance sample the phase using the "
         "emitter as incident direction. Set to <0. to deactivate.",
         type="float",
-        init_type="float",
     )
 
     @property
--- a/src/eradiate/scenes/integrators/_path_tracers.py
+++ b/src/eradiate/scenes/integrators/_path_tracers.py
@@ -49,8 +49,13 @@ class MonteCarloIntegrator(Integrator):
     def kernel_type(self) -> str:
         raise NotImplementedError
 
-    @property
-    def template(self) -> dict:
+    def _build_kernel_dict(self) -> dict:
+        """
+        Build the kernel-specific dictionary.
+
+        Override this method in subclasses to add integrator-specific parameters.
+        The base implementation handles common Monte Carlo integrator parameters.
+        """
         result = {"type": self.kernel_type}
 
         if self.timeout is not None:
@@ -62,9 +67,18 @@ class MonteCarloIntegrator(Integrator):
         if self.hide_emitters is not None:
             result["hide_emitters"] = self.hide_emitters
 
+        return result
+
+    @property
+    def template(self) -> dict:
+        # Validation
         if self.stokes and not eradiate.mode().is_polarized:
             raise RuntimeError("stokes should only be set to True in polarized mode.")
 
+        # Build the kernel dict (children can override _build_kernel_dict)
+        result = self._build_kernel_dict()
+
+        # Apply wrapping layers
         if self.moment:
             result = {"type": "moment", "nested": result}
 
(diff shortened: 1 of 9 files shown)


src/eradiate/scenes/integrators/_path_tracers.py
```diff
@@ -109,6 +123,51 @@ def kernel_type(self) -> str:
...
 
 
+@define(eq=False, slots=False)
+class EOVolPathIntegrator(MonteCarloIntegrator):
+    """
+    A thin interface to the EO volumetric path tracer kernel plugin [``eovolpath``].
+
+    This integrator samples paths using random walks starting from the sensor.
+    It supports multiple scattering, accounts for volume interactions, and
+    implements the DDIS variance reduction method.
```
leroyvn, line 133: Add a reference to the VROOM paper.

src/eradiate/scenes/integrators/_path_tracers.py
```diff
@@ -109,6 +123,51 @@ def kernel_type(self) -> str:
...
+    """
+
+    rr_depth: int | None = documented(
+        attrs.field(default=1000, converter=attrs.converters.optional(int)),
+        doc="Minimum path depth after which the implementation starts applying "
+        "the Russian roulette path termination criterion.",
```
leroyvn, line 139: What happens if it is unset?

src/eradiate/scenes/integrators/_path_tracers.py
```diff
@@ -109,6 +123,51 @@ def kernel_type(self) -> str:
...
+        type="int or None",
+        init_type="int, optional",
+    )
+
+    rr_factor = documented(
+        attrs.field(default=0.97, converter=attrs.converters.optional(float)),
+        doc="Specifies the maximum probability to keep a path when Russian "
+        "Roulette is evaluated.",
+        type="float",
+        init_type="float",
```
leroyvn, line 149: Here the converter allows `None` and provide a default, but these are not documented. If `None` is indeed applicable, what is the associated behaviour?

src/eradiate/scenes/integrators/_path_tracers.py
```diff
@@ -109,6 +123,51 @@ def kernel_type(self) -> str:
...
+    )
+
+    ddis_threshold = documented(
+        attrs.field(default=0.1, converter=attrs.converters.optional(float)),
+        doc="Specifies the probability to importance sample the phase using the "
+        "emitter as incident direction. Set to <0. to deactivate.",
+        type="float",
+        init_type="float",
```
leroyvn, line 157: Same as for `rr_factor`.

src/eradiate/test_tools/test_cases/integrators.py
```diff
@@ -0,0 +1,207 @@
...
+    - Sensor: Multi-distant measure covering the principal plane, from -75° to 75° with 2° increments, delta SRF positioned at λ = 550 nm
+    - Integrator: Integrator passed in `integrator_dict`.
+    """
+
+    return AtmosphereExperiment(
+        surface={"type": "rpv", "k": 0.95, "g": -0.1, "rho_0": 0.027685},
+        illumination={
+            "type": "directional",
+            "zenith": 20 * ureg.deg,
+            "irradiance": 20.0,
```
leroyvn, line 28: What is the reason to set it to this constant high value instead of using a solar irradiance spectrum?

src/eradiate/test_tools/test_cases/integrators.py
leroyvn, line 1: IMO it would be clearer for maintainers if all functions that are not intended to be part of the public API of this module would be underscore-prefixed.

tests/03_regression/integrators/test_eovolpath_canopy.py
```diff
@@ -0,0 +1,51 @@
...
+
+    reference = reference_converter(
+        "tests/regression_test_references/eovolpath_canopy_ref.nc"
+    )
+
+    if not reference:
+        # Ensure that volpath is used when creating a reference
+        exp = create_volpath_canopy(absorption_database_error_handler_config)
+    else:
+        exp = create_volpath_canopy(absorption_database_error_handler_config)
```
leroyvn: Shouldn't it be `create_eovolpath_canopy()` here?

tests/03_regression/integrators/test_eovolpath_canopy.py
```diff
@@ -0,0 +1,51 @@
...
+
+    result = eradiate.run(exp, spp=10000)
```
leroyvn, line 38: I suppose you'd like a higher sample count to create the reference?

src/eradiate/test_tools/test_cases/integrators.py
```diff
@@ -0,0 +1,207 @@
...
+@append_doc(create_integrator_canopy, prepend=True)
+def create_volpath_canopy(absorption_database_error_handler_config):
+    r"""
+    *Volpath Integrator*
+
+    """
+
+    integrator = {
+        "type": "volpath",
+        "moment": True,
```
leroyvn, line 202:
```suggestion
        "moment": True
```

src/eradiate/test_tools/test_cases/integrators.py
```diff
@@ -0,0 +1,207 @@
...
+def create_eovolpath_canopy(absorption_database_error_handler_config):
+    r"""
+    *EOVolpath Integrator*
+
+    """
+
+    integrator = {
+        "type": "eovolpath",
+        "ddis_threshold": 0.1,
+        "moment": True,
```
leroyvn:
```suggestion
        "moment": True
```

src/eradiate/test_tools/test_cases/integrators.py
```diff
@@ -0,0 +1,207 @@
...
+def create_eovolpath_surface(absorption_database_error_handler_config):
+    r"""
+    *EOVolpath Integrator*
+
+    """
+
+    integrator = {
+        "type": "eovolpath",
+        "ddis_threshold": 0.1,
+        "moment": True,
```
leroyvn, line 150:
```suggestion
        "moment": True
```

src/eradiate/test_tools/test_cases/integrators.py
```diff
@@ -0,0 +1,207 @@
...
+@append_doc(create_integrator_surface, prepend=True)
+def create_volpath_surface(absorption_database_error_handler_config):
+    r"""
+    *Volpath Integrator*
+
+    """
+
+    integrator = {
+        "type": "volpath",
+        "moment": True,
```
leroyvn, line 167:
```suggestion
        "moment": True
```

leroyvn (changes requested):
Hi @frettini, thanks a lot for this PR! Just a few small changes to make.

src/eradiate/test_tools/test_cases/integrators.py
frettini: I thought it was good practice to have trailing commas on multiline containers (https://peps.python.org/pep-0008/#when-to-use-trailing-commas).

src/eradiate/test_tools/test_cases/integrators.py
frettini: I just copied that over from the case in `atmospheres.py`, I guess it makes it more lightweight but I can also just go for the solar irradiance spectrum.

src/eradiate/scenes/integrators/_path_tracers.py
frettini: I copied this from its parent class and modified the default, but now I'm wondering, why can it be None? Shouldn't it just be an int with a default value?

tests/03_regression/integrators/test_eovolpath_canopy.py
frettini: Good catch, I moved things around 100 times and lost track of it.

tests/03_regression/integrators/test_eovolpath_canopy.py
frettini: I am unsure. I followed the test in `test_rpv_afgl1986_continental.py` and it provides a result to ZTest. When I looked into `regression.py`, at line 346, it seems to archive the result as is in case the reference is not valid. I don't know whether the reference of a ztest needs to be more accurate than the data it is tested against.

frettini:
@leroyvn , Thank you for the review. I have resolved some comments and responded to others.

src/eradiate/scenes/integrators/_path_tracers.py
leroyvn: Now that I think about it, it's likely that I allowed `None` initially to leave the value unspecified in the kernel dict and let the plugin apply its default. For `eovolpath`, it's better to be more explicit. And actually maybe we should just set an explicit default everywhere for clarity.

src/eradiate/test_tools/test_cases/integrators.py
leroyvn: Since this is a integrator test, freezing this value makes sense. Then let's just change it to something more realistic like 2.

tests/03_regression/integrators/test_eovolpath_canopy.py
leroyvn: The ref should be computed with a high sample count, otherwise the statistical test should account for the uncertainty of it, which the z-test does not. But let's not deviate from the other tests: revisiting this should be part of a dedicated PR.

src/eradiate/test_tools/test_cases/integrators.py
leroyvn: If you remove that comma, ruff will reformat the container on a single line ;)

TBH I don't have a justification for this except I find it pointless to split short lines as it does not significantly improve code readability. That's me splitting hairs 😬

tests/03_regression/integrators/test_eovolpath_canopy.py
leroyvn: Adding to this: just make sure that the ref has a noise level similar to the one used in the other tests. I think the sample count is adjusted manually.

src/eradiate/scenes/integrators/_path_tracers.py
frettini: You mean the defaults of the kernel plugin? Then yes, since we need slightly more control over the RR depth in Eradiate, it might be better to set it as an explicit default on the Eradiate side.

tests/03_regression/integrators/test_eovolpath_canopy.py
frettini: Looking at the reports of those regression tests it does indeed look like references have less noise than the test data. I am not sure how those reference are generated but I modified it so that references use a higher sample count. We can discuss it further offline.

frettini:
@leroyvn I have addressed the rest of the comments and added one comment of my own. I have also updated the changelog. Remaining are to bump push the new references to the data repo and bump the submodules.

leroyvn:
CI is broken, but all tests pass locally: let's merge!

PR #568, opened 2026-04-21
[Feature] Experimental extremum structure interface

frettini:
# Description

This PR includes an experimental interface for extremum structures. Extremum Structures are superstructures that are constructed from volumes (atmosphere and particle layer) to provide local upper and lower bounds of extinction coefficients. Using those local bounds drastically improves the performance of delta tracking in case of heterogeneous volumes with peaked density. 

While the kernel provides flexibility on the choice of structures in the form of plugins, the interface in Eradiate is more restrictive: the structure is tied to the geometry, namely `extremum_grid` for `plane_parallel` and `extremum_spherical` for `spherical_shell`. Since the two structures are based on grids, we expose `extremum_resolution` for both through the abstract heterogeneous interface. A few things to note:
- A resolution of (1,1,1), defaults to not describing an extremum. (future implementation might default to global implementations).
- The resolution will have different meaning for plane parallel (x, y, z) and spherical shell (r, theta, phi). The current implementation of `extremum_spherical` only accept `r` to be bigger than one and will throw otherwise.
- A resolution of (0,0,0) in plane parallel will trigger adaptive resolution search, which will find the optimal resolution. This will trigger at each spectral update.

Within this PR is also included a check to the kernel color mode that sets `has_spectral_extinction`. This is mostly for performance reasons as we can save on a few operations in non-spectral mode (order of magnitude of a 2 to 3%). Note also that the current implementation of extremum structures is not compatible with spectral extinction coefficients, which will be fixed in the next kernel update. This change unveiled that there might be some error in non-spectral integration of `piecewise` media. A fix should be done in a separate PR.

This is an experimental feature: 
- The interface is likely to change as new structures are introduced.
- There are incoming structural changes that will restrict the use of extremum structures to `eovolpath`.

A couple of unit tests were added to ensure the correct medium and extremum type are used.

# Checklist

### DIFF
--- a/src/eradiate/scenes/atmosphere/_core.py
+++ b/src/eradiate/scenes/atmosphere/_core.py
@@ -379,6 +379,25 @@ class AbstractHeterogeneousAtmosphere(Atmosphere, ABC):
         init_type="bool, optional",
     )
 
+    extremum_resolution: tuple[int, int, int] = documented(
+        attrs.field(
+            default=(1, 1, 1),
+            validator=validators.is_vector3,
+        ),
+        doc="[EXPERIMENTAL] Resolution of the extremum structure. Applies to "
+        "both plane parallel and spherical shell geometries but the latter only "
+        "accepts a resolution greater than one along its radial dimension "
+        "(first dimension). Also note than in plane parallel the dimensions are "
+        "[x, y, z] and in spherical shell [r, theta, phi]. "
+        "The default value (1, 1, 1) falls back to a global majorant. In plane "
+        "parallel, set to (0,0,0) to trigger an adaptive search of the optimal"
+        "resolution. Note that this is an expensive search that triggers at "
+        "spectral update.",
+        type="tuple[int, int, int]",
+        init_type="tuple[int, int, int]",
+        default="(1,1,1)",
+    )
+
     def update(self) -> None:
         """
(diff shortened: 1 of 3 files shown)


src/eradiate/scenes/atmosphere/_core.py
```diff
@@ -676,9 +701,19 @@ def _template_medium(self) -> dict:
                     ),
                     "to_world": to_world,
                     "filter_type": "nearest",
+                    "accel": False,
```
leroyvn: I never cared to set this; why do you need it to be `False`?

leroyvn (approved):
Hi @frettini, thank you! This look good to me, it would be nice if you could add a change log entry. We can then merge.

src/eradiate/scenes/atmosphere/_core.py
frettini: This was a remnant of developing the extremum structures. I have removed it now.

frettini:
@leroyvn , thank you for the feedback, I have adressed it, and moved extremum_resolution as I mistakenly put it in `Atmosphere` instead of `AbstractHeterogenousAtmosphere`.

leroyvn:
Thanks! Merging now :rocket:

PR #569, opened 2026-04-28
Residual Ratio Tracking control parameter and user warning guard

frettini:
# Description

Hello,
This PR adds a user parameter to control which algorithm is used to estimate the transmittance. This parameter is part of the Atmosphere interface as it applies to the medium in Mistuba. Note however that only the `EOVolpathIntegrator` currently consumes this parameter.

The PR also adds a user warning when specifiying an `extremum_resolution` with an incompatible integrator. Note that the simulation will pursue and will just ignore the parameter.

# Checklist

### DIFF
--- a/src/eradiate/experiments/_atmosphere.py
+++ b/src/eradiate/experiments/_atmosphere.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import logging
 import typing as t
+import warnings
 
 import attrs
 
@@ -14,6 +15,7 @@ from ._helpers import (
 )
 from ..attrs import AUTO, define, documented
 from ..scenes.atmosphere import (
+    AbstractHeterogeneousAtmosphere,
     Atmosphere,
     HeterogeneousAtmosphere,
     HomogeneousAtmosphere,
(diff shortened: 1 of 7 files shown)


src/eradiate/scenes/integrators/_path_tracers.py
```diff
@@ -43,6 +43,10 @@ class MonteCarloIntegrator(Integrator):
         init_type="bool, optional",
     )
 
+    @property
+    def extremum_compatible(self):
```
leroyvn, line 47: Missing docstring

src/eradiate/scenes/atmosphere/_core.py
```diff
@@ -103,23 +103,18 @@ class Atmosphere(CompositeSceneElement, ABC):
         default='"plane_parallel"',
     )
 
-    extremum_resolution: tuple[int, int, int] = documented(
+    use_residual_ratio_tracking: bool = documented(
```
leroyvn: How about using here a string flag, *e.g.* `tracking="rrt"` with a sensible default? That would allow to offer more options if ever relevant. Otherwise, I would just mirror the Mitsuba interface here and name this parameter `use_rrt`.

Also, it looks like you removed `extremum_resolution`: is it redefined downstream in the class hierarchy?

src/eradiate/experiments/_atmosphere.py
```diff
@@ -166,6 +168,19 @@ def _normalize_integrator(self) -> None:
...
         """
+        if isinstance(self.atmosphere, AbstractHeterogeneousAtmosphere):
+            if (
+                self.atmosphere.extremum_resolution != (1, 1, 1)
+                and not self.integrator.extremum_compatible
+            ):
+                warnings.warn(
+                    UserWarning(
+                        "Extremum structures are not compatible with "
+                        f"``{type(self.integrator).__name__}`` and "
```
leroyvn:
```suggestion
                        f"{type(self.integrator).__name__} and "
```

leroyvn (changes requested):
Hi @frettini, this is a very quick feedback. Overall the approach looks good to me.

src/eradiate/scenes/atmosphere/_core.py
frettini: I'd rather mirror Mitsuba as it maps better to what happens in code. I don't understand why git shows it as removed. It still appears in AbstractHeterogeneousAtmosphere where it already was..

frettini:
@leroyvn , thank you for the feedback, I have applied changes and propagated the user warning check to other integrators.

PR #570, opened 2026-05-13
Fix SolarIrradianceSpectrum.datetime not supporting datetime objects

mone27:
# Description

Fixes the bug where `SolarIrradianceSpectrum.datetime` can only be initialised with a string and not also a `datetime` object (as suggested in the docs)

I also added a test for the correct parsing of datetime

# Checklist


leroyvn:
Hi @mone27, thanks a lot for reporting this issue and proposing a fix. I will review this ASAP; I however have to think a little about how to solve this: it might actually be better to go a little further and accept a broader range of types (e.g. NumPy's `datetime64`). I will let you know.

leroyvn:
I will close this PR now since the underlying bug was solved by #572. Thank you @mone27!

PR #571, opened 2026-05-18
Use `stokes_moment` integrator instead of separate nested `stokes` and `moment`

frettini:
# Description

Hello,
This PR enforces the use of the `stokes_moment` integrator when either the stokes components, the variance, or both elements are required as output. This change solves the nesting order constraint imposed by recent performance fixes in the stokes component. The pipeline had to be modified to process all stokes components again. The `stokes_moment` integrator tries to reproduce the AOV order and naming of the integrators it is derived from. There is an exception when only the Stokes vector is required, and the name of the nested integrator is preprended to the AOV i.e `nested.S0`. This change was also reflected in the processing changes.

The breaking unit tests were modified accordingly. All tests pass, including regression tests, and some IPRT cases were also run again.

# Checklist

### DIFF
--- a/tests/01_unit/pipelines/test_logic.py
+++ b/tests/01_unit/pipelines/test_logic.py
@@ -222,15 +222,12 @@ def test_01_gather_bitmaps(mode, experiment, gather_bitmaps):
     solar_angle_sizes = {"sza": 1, "saa": 1}
     film_sizes = {"y_index": 32, "x_index": 32}
     all_sizes = {**spectral_sizes, **film_sizes, **solar_angle_sizes}
+    all_sizes_pol = {**all_sizes, "stokes": 4}
 
     expected_sizes = {
         "spp": spectral_sizes,
-        "radiance_raw": all_sizes
-        if not mode.is_polarized
-        else {**all_sizes, "stokes": 4},
-        "radiance_m2_raw": all_sizes
-        if not mode.is_polarized
-        else {**all_sizes, "stokes": 1},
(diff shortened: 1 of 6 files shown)


tests/01_unit/pipelines/test_logic.py
```diff
@@ -222,15 +222,12 @@ def test_01_gather_bitmaps(mode, experiment, gather_bitmaps):
...
     expected_sizes = {
         "spp": spectral_sizes,
-        "radiance_raw": all_sizes
-        if not mode.is_polarized
-        else {**all_sizes, "stokes": 4},
-        "radiance_m2_raw": all_sizes
-        if not mode.is_polarized
-        else {**all_sizes, "stokes": 1},
+        "radiance_raw": all_sizes if not mode.is_polarized else all_sizes_pol,
+        "radiance_m2_raw": all_sizes if not mode.is_polarized else all_sizes_pol,
```
leroyvn: Since you factored the new dimension sizes into a variable, negating the condition no longer helps with readability, so I'll suggest:

```suggestion
        "radiance_raw": all_sizes_pol if mode.is_polarized else all_sizes,
        "radiance_m2_raw": all_sizes_pol if mode.is_polarized else all_sizes,
```

leroyvn (changes requested):
Hi @frettini, no major comment, it looks good. I'll merge as soon as the PR will be ready. Thanks!

frettini:
I have applied the changes, updated the changelog and submodule.

PR #572, opened 2026-05-18
`SolarIrradiance`: Support datetime-like objects

leroyvn:
# Description

This PR updates `SolarIrradianceSpectrum`'s `datetime` attribute to support conversion of various datetime-like types. Now, the following types are accepted:

* `datetime.datetime`: Pivot data type, used internally for storage.
* `str`: ISO format string specification, parsed with `dateutil` as a `datetime.datetime`.
* `pd.Timestamp`: Pandas's own datetime type, converted to `datetime.datetime`.
* `np.datetime64`: NumPy's own datetime type, converted to `datetime.datetime`.

I also refactored the unit tests to use classes for scoping (new preferred pattern).

Closes #570.

# Checklist

### DIFF
--- a/docs/release_notes/v1.2.x.md
+++ b/docs/release_notes/v1.2.x.md
@@ -10,14 +10,16 @@
 
 * Improved error messages when initializing a {class}`.MolecularAtmosphere` with
   an incorrect value for the ``absorption_data`` parameter ({ghpr}`566`).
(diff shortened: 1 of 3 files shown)


leroyvn:
Hi @mone27, I worked on solving #570 — and a bit more. We now accept a range of datetime-like objects, including Python's `datetime`. The fix will be included in our next release.

PR #584, opened 2026-07-29
[3DATM 1/7] Add a particles v1 data format specification extending aer core v2

wint3ria:
# Description

Add an extension to the aer_core_v2 data format to support varying r_eff and v_eff distributions for particles phase functions.

This PR also includes two changes related to minor issues I faced:
 - Deprecation warning issued by numpy due to netCDF4 not supporting NumPy >= 2.5
 - tests/03_regression/romc/test_het01.py threshold being too tight.

Let me know if you want to address any of these differently.

# Checklist

### DIFF
--- a/docs/data/formats/aer.rst
+++ b/docs/data/formats/aer.rst
@@ -122,6 +122,64 @@ Normalization conventions
       * ``imom``
       * ``nmommax``
 
+.. _sec-data-formats-prt_v1:
+
+Prt v1 [``prt_v1``]
+--------------------
+
+Extension of ``aer_core_v2`` for particle properties. This data format
+indexes properties defined by ``aer_core_v2`` on two extra dimensions,
+``reff`` and ``veff``, the mean and variance of the particle size
+distribution, respectively. It shares the same normalization conventions
+as ``aer_core_v2``.
+
+Format
+    ``xarray.Dataset`` (in-memory), NetCDF (storage)
+
+Dimensions
+    * ``w``: radiation wavelength
+    * ``reff``: effective radius of the particles
+    * ``veff``: variance of the particles radii
+    * ``phamat``: nonzero coefficients in the phase matrix
+    * ``iangle``: angular data points
+    * ``imom``: Legendre coefficients for the (1,1) phase matrix component
+
+Coordinates
+    *When relevant, units are required and specified in the "units" metadata field.*
+
+    * ``w(w)``: float [length]: wavelength
+    * ``reff(reff)``: float [length]: effective radius
+    * ``veff(veff)``: float [—]: variance of particles radii
+    * ``phamat(phamat)``: str [—]: row and column indices of the phase matrix coefficient
+      (*e.g.* "11", "12", etc.)
+    * ``theta(w, reff, veff, iangle)``: float [angle]: scattering angle θ; NaN-padded beyond
+      ``nangles[iw, ireff, iveff]``
+    * ``mu(w, reff, veff, iangle)``: float [—]: value of cos θ; NaN-padded beyond 
+      ``nangles[iw, ireff, iveff]``
+
+Data variables
+    *When relevant, units are required and specified in the "units" metadata field.*
+
+    * ``ext(w, reff, veff)`` float [1 / length]: extinction coefficient per unit concentration
+    * ``ssa(w, reff, veff)`` float [—]: single-scattering albedo
+    * ``phase(phamat, w, reff, veff, iangle)`` float [1 / sr]: value of the phase matrix;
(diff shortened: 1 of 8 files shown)


docs/data/formats/aer.rst
```diff
@@ -122,6 +122,64 @@ Normalization conventions
...
+    * ``pmom(w, reff, veff, imom)`` float [—], optional: Legendre expansion coefficients 
+      of the (1,1) phase matrix element; NaN-padded beyond ``nmom[iw, ireff, iveff]``
+    * ``nmom(w, reff, veff)`` int [—], required if ``pmom`` is present: number of valid
+      Legendre coefficients per wavelength; must be consistent with the NaN
+      padding of ``pmom``
+
+.. note::
+
+    * Data are sorted in ascending order of ``w``, ``reff``, ``veff`` and ``mu``
+    * ``reff`` and ``veff`` are regularly spaced
```
leroyvn, line 179: This is a strong limitation. Is it strictly necessary?

docs/data/formats/aer.rst
```diff
@@ -122,6 +122,64 @@ Normalization conventions
       * ``imom``
       * ``nmommax``
 
+.. _sec-data-formats-part_core_v1:
+
+Part-Core v1 [``part_core_v1``]
+-------------------------------
```
leroyvn: I've been using a trigram for format family identification. Also accounting for our conversation earlier, I think we can drop the `core` variant specifier and simply state compatibility and provide automatic promotion functions. So here is my suggestion:

```suggestion
Prt v1 [``prt_v1``]
-------------------
```

docs/data/formats/aer.rst
```diff
@@ -122,6 +122,64 @@ Normalization conventions
...
+    * ``phase(phamat, w, reff, veff, iangle)`` float [1 / sr]: value of the phase matrix;
+      NaN-padded beyond ``nangles[iw, ireff, iveff]``
+    * ``nangles(w, reff, veff)`` int [—]: number of valid angular samples per wavelength, 
+      reff and veff; must be consistent with the NaN padding of ``theta``, ``mu``, and 
+      ``phase``
+    * ``pmom(w, reff, veff, imom)`` float [—], optional: Legendre expansion coefficients 
+      of the (1,1) phase matrix element; NaN-padded beyond ``nmom[iw, ireff, iveff]``
+    * ``nmom(w, reff, veff)`` int [—], required if ``pmom`` is present: number of valid
+      Legendre coefficients per wavelength; must be consistent with the NaN
+      padding of ``pmom``
```
leroyvn: Are you sure if this makes sense? What do you think @emdec?

leroyvn (commented):
Hi @wint3ria, this is a first pass. The Legendre coefficient part has to be settled to decide whether the `prt` family should be considered an extension of the `aer` family, or if it has incompatibilities.

docs/data/formats/aer.rst
wint3ria: Concerning the ascending constraint, this is a fair constraint to add to indexing quantities, that xarray might enforce anyway.

Irregular grids add an extra binary search on `reff` and `veff` coordinates, although these are rather small and performed once per interaction. Since I don't have any use case for irregular grids, I did not implement any for these coordinates in the mitsuba plugin, so currently the constraint is strictly necessary here in Eradiate. Now a few elements:
 - This is a clear limitation of the new mitsuba `particlefieldphase` plugin
 - Releasing this constraint would be backward compatible with the current proposed format
 - This is not a complicated problem, I can change this now or later with minimal effort
 - Implementation at the Eradiate level will be the same, regardless of this constraint

For now I will continue with the current implementation and note this as an iterative improvement, to prioritize the delivery of the rest of the work on particles fields. The current limitation still allows us to validate against IPRT cases

docs/data/formats/aer.rst
wint3ria: The way I see it, pmom(w, reff, veff, imom) is not meant to be used as an interpolated quantity for configuration of the particlephase plugin. This is necessary to load a complete reff indexed libradtran formatted dataset, and we decided to add the veff coordinate. DISORT users must select a (reff, veff) couple to work on or perform their own interp logic, thus collapsing the prt_v1 format to aer_core_v2 that is meant to be used with this model

docs/data/formats/aer.rst
wint3ria: Done

wint3ria:
Closed to prefer a [stacked PR structure](https://docs.github.com/en/pull-requests/get-started/about-stacked-prs)

PR #586, opened 2026-07-31
[3DATM 3/7] Implement a coordinate system for 3D atmospheres

wint3ria:
# Description

Implement a coordinate system for gridded atmospheres and adapt the codebase to this new feature. The `GridCoords` interface effectively replaces the `ZGrid` class, introducing many small mechanical changes in the codebase. This abstraction indexes the extent of the 3D space where the atmospheric data and optical properties are defined. It generalizes `ZGrid` to 3 dimensions.

Two subclasses of `GridCoords`, `PlaneParallelGridCoords` and `SphericalShellGridCoords`, implement a Cartesian and a geocentric coordinate system respectively.

The `Geometry` keeps the same role of defining the shape of the atmosphere volume, but the different components are now mapped on the `GridCoords` subregion of the shape. The `Geometry` subclasses are updated to introduce wrap mode parametrization, although this feature is not implemented yet (see #589).

A new set of tools is proposed to facilitate the configuration of a mitsuba `GridVolume` instances from a `Geometry` instance and its associated `GridCoords`.

Lastly, to support end user manipulation of these coordinates, a small visualization tool is proposed for notebook cells. This module is implemented in pure SVG without any external library, and only output a few diagrams representing the grids or atmosphere. This allows a fast render time for interactive sessions.

Here is an example representation of a default 1D grid in Eradiate:
image

The scale is not realistic. Only 3 different sizes can be assigned to an axis, and sizes are attributed by order of length of the axes' extent. In the example above, the default coordinate grid is 120km high, and both X and Y extents are the same size at 5e6 km and thus represented with the same size. Thus this representation is only a convenient visual cue to help debugging or configuring a scene.

Here is an example with another narrower grid geometry:
image

For `Geometry` instances, a similar visualization uses a logarithmic scale to represent the extent of the inner grid compared to the atmosphere's shape:
image

Another representation is used for spherical geometries:
image

Note that although these representations are implemented in this PR, the actual wrap mode support is implemented in the subsequent #589. Using a grid smaller than the atmosphere shape at this stage is undefined behavior.

Examples taken from https://github.com/eradiate/eradiate-tutorials/pull/5

# Checklist

### DIFF
--- a/docs/data/formats/aer.rst
+++ b/docs/data/formats/aer.rst
@@ -122,6 +122,64 @@ Normalization conventions
       * ``imom``
       * ``nmommax``
 
+.. _sec-data-formats-prt_v1:
+
+Prt v1 [``prt_v1``]
+--------------------
+
+Extension of ``aer_core_v2`` for particle properties. This data format
+indexes properties defined by ``aer_core_v2`` on two extra dimensions,
+``reff`` and ``veff``, the mean and variance of the particle size
+distribution, respectively. It shares the same normalization conventions
+as ``aer_core_v2``.
+
+Format
+    ``xarray.Dataset`` (in-memory), NetCDF (storage)
+
+Dimensions
+    * ``w``: radiation wavelength
+    * ``reff``: effective radius of the particles
+    * ``veff``: variance of the particles radii
+    * ``phamat``: nonzero coefficients in the phase matrix
+    * ``iangle``: angular data points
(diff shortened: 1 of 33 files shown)


wint3ria:
Closed to prefer a [stacked PR structure](https://docs.github.com/en/pull-requests/get-started/about-stacked-prs)

PR #592, opened 2026-08-03
Refactor regression testing framework

leroyvn:
# Description

Merge after #582 and #594.

This PR refactors the regression testing framework. The `eradiate.test_tools.regression` module is restructured around a smaller, better defined set of statistical tests, with improved diagnostics when a regression test fails. Changes are as follows:

* The the test hierarchy is trimmed down to `RMSETest` and `ZTest`, both deriving from a reworked `RegressionTest` base. The chi-square and Student t-test variants are dropped. The Šidák correction is now exposed as a plain `sidak_family_p_value()` helper instead of a dedicated test class.
* Failed regression tests now raise a `RegressionTestFailure` exception carrying the test outcome, instead of returning  `False`.
* Diagnostic charts, *e.g.* showing statistical test results, are added to comparison charts. Comparison plots for cases that include multiple wavelengths show multiple wavelengths at once, encoded with an isoluminant colormap (`colorcet`)
* Diagnostic chart generation is gated on the existing `--plot` flag (consistently with the comparison plot they are added to).
* Added a `--update-references` flag (and matching `update_references` fixture): without it, a missing reference dataset is an error instead of being silently archived as a reference candidate.
* Updated all regression tests to the new API, added a system-level check (`tests/02_system/test_regression_framework.py`) and extended the unit tests.
* The obsolete `resources/regression_testing_framework_benchmark` script is dropped.
* The contributing guide and API reference are updated.

This PR adds [Colorcet](https://github.com/holoviz/colorcet) as a new dependency to provide isoluminant colormaps. This can be considered low risk, as replacing it if unmaintained is very easy.

# Checklist


leroyvn:
Closing (superseded by #596 and #598).

PR #593, opened 2026-08-04
[enhancement] Refactor scene traversal by moving it to the kernel, allowing support for traversal of multiple parents.

frettini:
# Description

Hello,
This PR is linked to [kernel PR 37](https://github.com/eradiate/mitsuba3/pull/37). 
The aims of this PR are to:
- Move the kernel traversal logic in the kernel, enabling better separation of concern. It also ensures that it is possible to use the kernel on its own.
- Ensure that the name id pinning is properly ported over and that multiple parent node traversal properly works.

Moving the traversal logic in the kernel means that it is not possible to pass the content of `KernelSceneParameterMap` in the traverse function as it is `Eradiate` specific. This, in turn, means that populating the `uparam.parameter_id` needs to be done after the traversal, as opposed to during.  

The `mi_traverse` remains as is. The test suite was run and passed, and a test for multiple parent traversal was added to `test_render.py`.

Now that the traversal function is in the kernel, it might make sense to move parts of `test_render.py` in there too. However, I still think it valuable to have those tests run when testing Eradiate, hence why I refrained from moving them, let me know what you think about that.

This PR includes a bump of the Mitsuba submodule.

# Checklist

### DIFF
--- a/docs/release_notes/v1.3.x.md
+++ b/docs/release_notes/v1.3.x.md
@@ -32,3 +32,4 @@
   renders) are now written to the `reports/` directory ({ghpr}`582`).
 * 🖥️ Move `figure_to_html()` from `test_tools.regression` to `test_tools.report`
   ({ghpr}`582`).
+* 🖥️ Move traversal logic of the mitsuba scene in the kernel code ({ghpr}`593`).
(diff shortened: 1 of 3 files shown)


leroyvn (changes requested):
Hi @frettini, thanks for this update. Looks good to me!

PR #589, opened 2026-07-31
[3DATM 6/7] Support texture wrapmodes for 3D atmosphere boundary conditions

wint3ria:
# Description

Implement wrapmode support for 3D atmospheres. By default the `WrapMode.CLAMP` extends the atmospheric properties defined on the `GridCoords` towards the entire volume extent:
image

`WrapMode.REPEAT` and `WrapMode.MIRROR` implement periodic and symetric wrap functions respectively

image

The three above images leverage changes made in the 3D `ParticleEnsemble` class, configured as a checkerboard of 20 meter high particles cubes of $OT\approx5$.

Wrapmode is supported in both plane parallel and spherical geometries. Below is an example of a circular ParticleEnsemble located on ~Earth-sized sphere:

image

Examples taken from https://github.com/eradiate/eradiate-tutorials/pull/5.

# Checklist

- [x] The code follows the relevant coding guidelines
- [x] The code generates no new warnings
- [x] The code is appropriately documented
- [x] The code is tested to prove its function
- [x] The feature branch is rebased on the current state of the `next` branch
- [x] I updated the change log if relevant
- [x] I give permission that the Eradiate project may redistribute my contributions under the terms of its license

PR #600, opened 2026-09-09
[3DATM 2/6] Coordinates and grid facility

wint3ria:
# Description

Implement a coordinate system for gridded atmospheres and adapt the codebase to this new feature. The `GridCoords` interface effectively replaces the `ZGrid` class, introducing many small mechanical changes in the codebase. This abstraction indexes the extent of the 3D space where the atmospheric data and optical properties are defined. It generalizes `ZGrid` to 3 dimensions.

Two subclasses of `GridCoords`, `PlaneParallelGridCoords` and `SphericalShellGridCoords`, implement a Cartesian and a geocentric coordinate system respectively.

The `Geometry` keeps the same role of defining the shape of the atmosphere volume, but the different components are now mapped on the `GridCoords` subregion of the shape. The `Geometry` subclasses are updated to introduce wrap mode parametrization, although this feature is not implemented yet (see #589).

A new set of tools is proposed to facilitate the configuration of a mitsuba `GridVolume` instances from a `Geometry` instance and its associated `GridCoords`.

Lastly, to support end user manipulation of these coordinates, a small visualization tool is proposed for notebook cells. This module is implemented in pure SVG without any external library, and only output a few diagrams representing the grids or atmosphere. This allows a fast render time for interactive sessions.

Here is an example representation of a default 1D grid in Eradiate:
image

The scale is not realistic. Only 3 different sizes can be assigned to an axis, and sizes are attributed by order of length of the axes' extent. In the example above, the default coordinate grid is 120km high, and both X and Y extents are the same size at 5e6 km and thus represented with the same size. Thus this representation is only a convenient visual cue to help debugging or configuring a scene.

Here is an example with another narrower grid geometry:
image

For `Geometry` instances, a similar visualization uses a logarithmic scale to represent the extent of the inner grid compared to the atmosphere's shape:
image

Another representation is used for spherical geometries:
image

Note that although these representations are implemented in this PR, the actual wrap mode support is implemented in the subsequent #589. Using a grid smaller than the atmosphere shape at this stage is undefined behavior.

Examples taken from https://github.com/eradiate/eradiate-tutorials/pull/5

# Checklist

### DIFF
--- a/docs/release_notes/v1.3.x.md
+++ b/docs/release_notes/v1.3.x.md
@@ -25,6 +25,10 @@
 
 ### Internal changes
 
+* 🖥️ `ZGrid` is replaced throughout the atmosphere stack by the new
+  `grid.py` facility (`GridCoords`, `PlaneParallelGridCoords`,
+  `SphericalShellGridCoords`) and a `generate_gridvolume()` helper for
+  building kernel volumes ({ghpr}`586`).
 * 🖥️ Documentation now builds with Sphinx 9 and Python 3.11 ({ghpr}`580`).
 * 🖥️ Local docs build now mocks Mitsuba and Dr.Jit like the Read The Docs build
   does ({ghpr}`580`).
--- a/src/eradiate/_repr_diagrams.py
+++ b/src/eradiate/_repr_diagrams.py
@@ -1,0 +1,740 @@
+"""SVG diagrams for GridCoords/SceneGeometry HTML reprs."""
+
+from __future__ import annotations
+
+import math
+import xml.etree.ElementTree as ET
+
+import pint
+
+from .units import unit_registry as ureg
+
+
+def _attrib(attrib: dict) -> dict[str, str]:
+    """Translate Python-friendly (underscored) attribute names to their SVG/CSS (hyphenated) form and stringify values, dropping ``None`` entries."""
+    return {k.replace("_", "-"): str(v) for k, v in attrib.items() if v is not None}
+
+
+def svg(width: float, height: float, *children: ET.Element, **attrib) -> ET.Element:
+    """Build an <svg> root sized to (width, height) with a matching viewBox."""
+    el = ET.Element(
+        "svg",
+        _attrib(
+            {
+                "width": f"{width}",
+                "height": f"{height}",
+                "viewBox": f"0 0 {width} {height}",
+                "xmlns": "http://www.w3.org/2000/svg",
+                **attrib,
+            }
+        ),
+    )
+    el.extend(children)
+    return el
+
+
+def g(*children: ET.Element, **attrib) -> ET.Element:
+    """Build a <g> group element."""
+    el = ET.Element("g", _attrib(attrib))
+    el.extend(children)
+    return el
+
+
+def line(x1: float, y1: float, x2: float, y2: float, **attrib) -> ET.Element:
+    """Build a <line> element."""
+    return ET.Element(
+        "line",
+        _attrib({"x1": f"{x1}", "y1": f"{y1}", "x2": f"{x2}", "y2": f"{y2}", **attrib}),
+    )
+
+
+def rect(x: float, y: float, width: float, height: float, **attrib) -> ET.Element:
+    """Build a <rect> element."""
+    return ET.Element(
+        "rect",
+        _attrib(
+            {
+                "x": f"{x}",
+                "y": f"{y}",
+                "width": f"{width}",
+                "height": f"{height}",
+                **attrib,
+            }
+        ),
+    )
+
+
+def circle(cx: float, cy: float, r: float, **attrib) -> ET.Element:
+    """Build a <circle> element."""
+    return ET.Element(
+        "circle", _attrib({"cx": f"{cx}", "cy": f"{cy}", "r": f"{r}", **attrib})
+    )
+
+
+def ellipse(cx: float, cy: float, rx: float, ry: float, **attrib) -> ET.Element:
+    """Build an <ellipse> element."""
+    return ET.Element(
+        "ellipse",
+        _attrib({"cx": f"{cx}", "cy": f"{cy}", "rx": f"{rx}", "ry": f"{ry}", **attrib}),
+    )
+
+
+def polygon(points: list[tuple[float, float]], **attrib) -> ET.Element:
+    """Build a <polygon> element from a sequence of (x, y) points."""
+    pts = " ".join(f"{x},{y}" for x, y in points)
+    return ET.Element("polygon", _attrib({"points": pts, **attrib}))
+
+
+def path(d: str, **attrib) -> ET.Element:
+    """Build a <path> element with the given path data."""
+    return ET.Element("path", _attrib({"d": d, **attrib}))
+
+
+def text(x: float, y: float, content: str, **attrib) -> ET.Element:
+    """Build a <text> element."""
+    el = ET.Element("text", _attrib({"x": f"{x}", "y": f"{y}", **attrib}))
+    el.text = content
+    return el
+
+
+def el(
+    tag: str, *children: ET.Element, text: str | None = None, **attrib
+) -> ET.Element:
+    """Build a generic (typically non-SVG, e.g. HTML) element with optional text content and children."""
+    node = ET.Element(tag, _attrib(attrib))
+    node.text = text
+    node.extend(children)
+    return node
+
+
+def tostring(node: ET.Element) -> str:
+    """Serialize an element tree to an XML string."""
+    return ET.tostring(node, encoding="unicode")
+
+
+def _division_fractions(
+    n_cells: int, *, max_shown: int = 5, gap_scale: float = 2.0
+) -> tuple[list[float], tuple[float, float] | None]:
+    """Division fractions for up to *max_shown* grid segments, plus the compressed-gap range (*gap_scale* times as wide as a regular cell) when *n_cells* exceeds *max_shown*."""
+    if n_cells <= 1:
+        return [], None
+    if n_cells <= max_shown:
+        return [i / n_cells for i in range(1, n_cells)], None
+
+    n_head = n_tail = 2
+    unit = 1.0 / (n_head + n_tail + gap_scale)
+    lines = [
+        unit,
+        2 * unit,
+        (2 + gap_scale) * unit,
+        (3 + gap_scale) * unit,
+    ]
+    gap = (2 * unit, (2 + gap_scale) * unit)
+    return lines, gap
+
+
+_AXIS_SCALES = (0.55, 1.0, 1.6)  # (small, medium, large) edge-length multipliers
+
+
+def _axis_scales(
+    span_x: pint.Quantity,
+    span_y: pint.Quantity,
+    span_z: pint.Quantity,
+    scales: tuple[float, float, float] = _AXIS_SCALES,
+) -> tuple[float, float, float]:
+    """Rank each axis by physical extent into one of 3 predefined *scales* (small/medium/large, ties share a scale, non-comparable axes default to medium)."""
+    spans = [span_x, span_y, span_z]
+    small, medium, large = scales
+
+    def _cmp(i: int, j: int) -> int | None:
+        try:
+            if spans[i] < spans[j]:
+                return -1
+            if spans[i] > spans[j]:
+                return 1
+            return 0
+        except pint.DimensionalityError:
+            return None
+
+    result = [medium, medium, medium]
+    for i in range(3):
+        others = [j for j in range(3) if j != i and _cmp(i, j) is not None]
+        if not others:
+            continue
+        is_min = all(_cmp(i, j) <= 0 for j in others)
+        is_max = all(_cmp(i, j) >= 0 for j in others)
+        if is_min and not is_max:
+            result[i] = small
+        elif is_max and not is_min:
+            result[i] = large
+    return tuple(result)
+
+
+def _format_extent(lo: pint.Quantity, hi: pint.Quantity) -> str:
+    return f"{lo:~P.4g} … {hi:~P.4g}"
+
+
+def _visual_fraction(
+    ratio: float, *, floor: float = 0.15, log_floor_ratio: float = 1e-6
+) -> float:
+    """Map a linear size ratio in (0, 1] to a visual fraction in [*floor*, 1] on a log scale."""
+    if ratio >= 1.0:
+        return 1.0
+    if ratio <= log_floor_ratio:
+        return floor
+    t = math.log10(ratio) / math.log10(log_floor_ratio)  # 0 at ratio=1, 1 at floor
+    return floor + (1.0 - t) * (1.0 - floor)
+
+
+def _add(
+    p: tuple[float, float], v: tuple[float, float], f: float = 1.0
+) -> tuple[float, float]:
+    return (p[0] + v[0] * f, p[1] + v[1] * f)
+
+
+def _pt(
+    origin: tuple[float, float],
+    va: tuple[float, float],
+    fa: float,
+    vb: tuple[float, float],
+    fb: float,
+) -> tuple[float, float]:
+    return _add(_add(origin, va, fa), vb, fb)
+
+
+def _cuboid_vertices(
+    p0: tuple[float, float],
+    ux: tuple[float, float],
+    uy: tuple[float, float],
+    uz: tuple[float, float],
+) -> tuple[
+    tuple[float, float],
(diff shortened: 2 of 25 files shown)


docs/release_notes/v1.3.x.md
```diff
@@ -25,6 +25,10 @@
 
 ### Internal changes
 
+* 🖥️ `ZGrid` is replaced throughout the atmosphere stack by the new
+  `grid.py` facility (`GridCoords`, `PlaneParallelGridCoords`,
+  `SphericalShellGridCoords`) and a `generate_gridvolume()` helper for
+  building kernel volumes ({ghpr}`586`).
```
leroyvn, line 31:
```suggestion
  building kernel volumes ({ghpr}`600`).
```

src/eradiate/scenes/atmosphere/_core.py
```diff
@@ -172,7 +177,6 @@ def phase(self) -> PhaseFunction:
         .PhaseFunction
             Phase function associated with the atmosphere.
         """
-        pass
```
leroyvn: Was this done by the linter?

src/eradiate/scenes/atmosphere/_core.py
```diff
@@ -666,12 +679,12 @@ def eval_transmittance(
         try:
             sigma = eval_sigma[interaction](si=si)
         except KeyError as e:
-            raise ValueError(
+            raise TypeError(
```
leroyvn: Please don't change that unless there is a very compelling reason. This can break compatibility for callers.

src/eradiate/scenes/atmosphere/_core.py
```diff
@@ -692,109 +705,48 @@ def _template_medium(self) -> dict:
...
+                extremum = {
+                    "type": "extremum_grid",
+                    "volume": {"type": "ref", "id": sigma_t_id},
+                    "resolution": self.extremum_resolution,
+                    "to_world": self.geometry.atmosphere_volume_to_world,
                 }
-
         else:
-            raise ValueError(
+            raise TypeError(
```
leroyvn: Ditto

src/eradiate/scenes/atmosphere/_core.py
```diff
@@ -692,109 +705,48 @@ def _template_medium(self) -> dict:
...
-
-        elif isinstance(self.geometry, SphericalShellGeometry):
-            volume_rmin = self.geometry.atmosphere_volume_rmin
-            to_world = self.geometry.atmosphere_volume_to_world
+        volumes = {
+            "albedo": generate_gridvolume(
+                self.geometry,
+                self.eval_albedo,
+                units=ureg.dimensionless,
+                dtype=np.float64,
```
leroyvn, line 714: What is the reason to explicitly enforce float64 here instead of letting Numpy or Dr.Jit promote if necessary?

src/eradiate/scenes/atmosphere/_heterogeneous.py
```diff
@@ -117,29 +114,12 @@ def _particle_layers_validator(self, attribute, value):
...
-        init_type=".ZGrid, optional",
-    )
-
     @property
     def components(self) -> list[MolecularAtmosphere | ParticleLayer]:
         """
         Returns
         -------
-        list of .AbstractHeterogeneousAtmosphere
+        list
```
leroyvn, line 122: Please don't change that

src/eradiate/scenes/atmosphere/_heterogeneous.py
```diff
@@ -20,16 +20,13 @@
...
+from ...grid import GridCoords
 from ...kernel import SearchSceneParameter
-from ...radprops import ZGrid
 from ...spectral.index import SpectralIndex
 from ...units import unit_context_config as ucc
 from ...units import unit_registry as ureg
 from ...util.misc import cache_by_id
 
-if TYPE_CHECKING:
-    from axsdb import AbsorptionDatabase
```
leroyvn, line 31: This is intentional: no reason to import this type here outside of type checking as the class is not used anywhere. Does this break something?

src/eradiate/scenes/_gridvolume.py
```diff
@@ -0,0 +1,284 @@
...
+        Parameter lookup strategy forwarded to :class:`.SceneParameter`.
+        Only used when ``flag`` is set.
+
+    filter_type : str, optional
+        Mitsuba filter type string. Defaults to ``str(geometry.filter_type)``.
+
+    wrap_mode : str, optional
+        Mitsuba wrap mode string. Defaults to ``str(geometry.wrap_mode)``.
+
+    spectral_index : bool, optional, default: False
```
leroyvn, line 220:
```suggestion
    spectral_index : bool, default: False
```

src/eradiate/scenes/_gridvolume.py
```diff
@@ -0,0 +1,284 @@
...
+        If ``True``, pass ``ctx.si`` instead of ``ctx`` to ``eval_grid``.
+
+    units : pint.Unit, optional
+    If set, convert grid values to these units before processing.
```
leroyvn, line 224:
```suggestion
        If set, convert grid values to these units before processing.
```

src/eradiate/scenes/geometry.py
```diff
@@ -10,90 +11,137 @@
...
+    def __str__(self):
+        return self.name.lower()
+
+
+class FilterType(Enum):
+    TRILINEAR = 0
+    NEAREST = 1
+
+    def __str__(self) -> str:
+        return self.name.lower()
```
leroyvn, line 38: Unless the int values are used, I'd prefer to keep consistency with current enum definitions (*e.g.* `QuadType`). The converters defined in the updated `SceneGeometry` would then automatically convert strings.

src/eradiate/grid.py
```diff
@@ -0,0 +1,1183 @@
...
+from abc import ABC, abstractmethod
+
+import mitsuba as mi
+import numpy as np
+import pint
+import pinttr
+import xarray as xr
+from pinttr.util import ensure_units
+
+from eradiate.kernel.transform import map_unit_cube
```
leroyvn, line 12: Please use relative imports

src/eradiate/grid.py
```diff
@@ -0,0 +1,1183 @@
+from __future__ import annotations
+
+from abc import ABC, abstractmethod
+
+import mitsuba as mi
+import numpy as np
+import pint
+import pinttr
+import xarray as xr
+from pinttr.util import ensure_units
```
leroyvn, line 10:
```suggestion
from pinttrs.util import ensure_units
```

src/eradiate/_repr_diagrams.py
leroyvn, line 1: Let's relocate this module under the `util` namespace.

I generally like the idea of a repr for notebooks, but I'll need more time to figure out how to organize this. Let's keep at as it is for now, with no guarantee for stability.

src/eradiate/scenes/atmosphere/_core.py
```diff
@@ -10,26 +10,31 @@
 import xarray as xr
 from axsdb import AbsorptionDatabase
 
+from eradiate.kernel._render import SearchSceneParameter
```
leroyvn, line 13:
```suggestion
from ...kernel import SearchSceneParameter
```

src/eradiate/grid.py
```diff
@@ -0,0 +1,1183 @@
...
+        ensure_units(bottom, default_units=ucc.get("length"))
+        ensure_units(width, default_units=ucc.get("length"))
+
+        levels = ureg.convert(
+            np.arange(
+                bottom.m_as(ureg.m), (top + step * 0.1).m_as(ureg.m), step.m_as(ureg.m)
+            ),
+            ureg.m,
+            ucc.get("length"),
+        )
```
leroyvn: Cache the units instead:

```suggestion
        units_length = ucc.get("length")
        ensure_units(top, default_units=units_length)
        ensure_units(bottom, default_units=units_length)
        ensure_units(width, default_units=units_length)

        levels = ureg.convert(
            np.arange(
                bottom.m_as(ureg.m), (top + step * 0.1).m_as(ureg.m), step.m_as(ureg.m)
            ),
            ureg.m,
            units_length,
        )
```

src/eradiate/scenes/atmosphere/_core.py
```diff
@@ -487,7 +487,7 @@ def eval_radprops(
                 },
             ),
             "albedo": (
-                "z_layer",
+                ("x_cells", "y_cells", "z_layer"),
```
leroyvn: For consistency, please use singular here:

```suggestion
                ("x_cell", "y_cell", "z_layer"),
```

src/eradiate/grid.py
```diff
@@ -0,0 +1,1183 @@
+from __future__ import annotations
+
+from abc import ABC, abstractmethod
+
+import mitsuba as mi
+import numpy as np
+import pint
+import pinttr
```
leroyvn, line 8: Please use pinttrs ("New" API)

```suggestion
import pinttrs
```

src/eradiate/grid.py
```diff
@@ -0,0 +1,1183 @@
...
+        if not np.isscalar(value.magnitude):
+            raise ValueError("layer height must be a scalar")
+
+    _total_height: pint.Quantity = pinttr.field(
+        units=ucc.deferred("length"),
+        on_setattr=None,
+    )
+
+    @staticmethod
+    def make_onedim_from_levels(levels: pint.Quantity):
```
leroyvn, line 81: To resonate a bit with NumPy, I suggest:
```suggestion
    def make_1d_from_levels(levels: pint.Quantity):
```
Can you please add a return type? Also: why restrict to plane-parallel?

src/eradiate/grid.py
```diff
@@ -0,0 +1,1183 @@
...
+        default = GridCoords.make_default()
+
+        return PlaneParallelGridCoords(
+            levels=levels,
+            edges_x=default.edges_x,
+            edges_y=default.edges_y,
+        )
+
+    @staticmethod
+    def make_onedim_arange(
```
leroyvn, line 99:
```suggestion
    def make_1d_arange(
```

src/eradiate/grid.py
leroyvn, line 136: Cache the units instead:

```suggestion
        units_length = ucc.get("length")
        ensure_units(top, default_units=units_length)
        ensure_units(bottom, default_units=units_length)
        ensure_units(width, default_units=units_length)

        levels = ureg.convert(
            np.arange(
                bottom.m_as(ureg.m), (top + step * 0.1).m_as(ureg.m), step.m_as(ureg.m)
            ),
            ureg.m,
            units_length,
        )
```

src/eradiate/grid.py
```diff
@@ -0,0 +1,1183 @@
...
+        return PlaneParallelGridCoords.from_extent_and_resolution(
+            levels=levels,
+            extent_x=width,
+            extent_y=width,
+            n_cells_x=1,
+            n_cells_y=1,
+        )
+
+    @staticmethod
+    def make_default():
```
leroyvn, line 147: What is the returned type?

src/eradiate/grid.py
```diff
@@ -0,0 +1,1183 @@
...
+
+        bottom = 0.0 * ureg.km
+        top = 120.0 * ureg.km
+        step = 100.0 * ureg.m
+        width = 1e6 * ureg.km
+
+        return GridCoords.make_onedim_arange(top, bottom, step, width)
+
+    @classmethod
+    def convert(cls, value: t.Any) -> t.Any:
```
leroyvn, line 164:
```suggestion
    def convert(cls, value: Any) -> Any:
```
(the typing import is also missing)

src/eradiate/grid.py
```diff
@@ -0,0 +1,1183 @@
...
+        """
+        Convert dictionary type of parameters to a GridCoords.
```
leroyvn, line 166:
```suggestion
        Convert dictionary type of parameters to a ``GridCoords``.
```

leroyvn, line 166: This does more than just convert dicts: it also interprets strings and quantities. Please revisit the docstring for consistency.

src/eradiate/grid.py
```diff
@@ -0,0 +1,1183 @@
...
+
+        Returns the default plane parallel ``GridCoords`` if value
+        is ``"plane_parallel_coords"``. If value is a ``pint.Quantity``,
+        attempts to construct a 1D atmosphere with the layer altitudes
+        contained in the quantity. If value is a ``np.ndarray``, interpret it
+        as an array of layer altitudes in meters.
+
+        If value is a dict, expects a ``"type"`` key with the value
+        ``"plane_parallel_coords"`` or ``"spherical_shell_geometry"``.
+        Other keys are passed to the relevant subclass constructor.
```
leroyvn, line 176: Overall a bit more structure would be nice here (basically disclose the logic like *e.g.* `convert_absdb()` does). The no-op case is missing.

src/eradiate/grid.py
```diff
@@ -0,0 +1,1183 @@
...
+        """Number of edges in the Z direction (equal to number of levels)."""
+        return self.n_levels
+
+    @property
+    def shape(self) -> tuple[int, int, int]:
+        """Grid dimensions sizes, or numpy shape. (X, Y, Z) order."""
+        return (self.n_cells_x, self.n_cells_y, self.n_cells_z)
+
+    @property
+    def onedim(self) -> bool:
```
leroyvn, line 306: I'd rather go with this:
```suggestion
    def is_1d(self) -> bool:
```

src/eradiate/grid.py
```diff
@@ -0,0 +1,1183 @@
...
+        return len(self.edges_y)
+
+    def __repr__(self) -> str:
+        return (
+            f"PlaneParallelGridCoords(\n"
+            f"  x: [{self.edges_x[0]:~P}, {self.edges_x[-1]:~P}], {self.n_cells_x} cells\n"
+            f"  y: [{self.edges_y[0]:~P}, {self.edges_y[-1]:~P}], {self.n_cells_y} cells\n"
+            f"  z: [{self.levels[0]:~P}, {self.levels[-1]:~P}], {self.n_layers} layers\n"
+            f")"
+        )
```
leroyvn, line 552: Also set `repr=False` in `define()` above.

I'd prefer a flat repr that doesn't have linebreaks (plays better with parents when not relying on rich):
         
```suggestion
    def __repr__(self) -> str:
        return (
            "<PlaneParallelGridCoords "
            f"x={self.edges_x[[0, -1]]:~P} ({self.n_cells_x} cells) "
            f"y={self.edges_y[[0, -1]]:~P} ({self.n_cells_y} cells) "
            f"z={self.levels[[0, -1]]:~P} ({self.n_layers} layers) "
            f">"
        )
```

src/eradiate/grid.py
```diff
@@ -0,0 +1,1183 @@
...
+        return len(self.colatitudes)
+
+    def __repr__(self) -> str:
+        return (
+            f"SphericalShellGridCoords(\n"
+            f"  azimuth:    [{self.azimuths[0]:~P}, {self.azimuths[-1]:~P}], {self.n_cells_x} sectors\n"
+            f"  colatitude: [{self.colatitudes[0]:~P}, {self.colatitudes[-1]:~P}], {self.n_cells_y} bands\n"
+            f"  z:          [{self.levels[0]:~P}, {self.levels[-1]:~P}], {self.n_layers} layers\n"
+            f")"
+        )
```
leroyvn, line 1130: Ditto

src/eradiate/scenes/atmosphere/_core.py
```diff
@@ -469,16 +467,18 @@ def eval_radprops(
...
-        albedo = self.eval_albedo(si, zgrid).m_as(ureg.dimensionless)
+        sigma_t = self.eval_sigma_t(si, grid)
+        sigma_t = np.broadcast_to(sigma_t, grid.shape)
+        albedo = self.eval_albedo(si, grid).m_as(ureg.dimensionless)
+        albedo = np.broadcast_to(albedo, grid.shape)
 
         data_vars = {
             "sigma_t": (
-                "z_layer",
+                ("x_cells", "y_cells", "z_layer"),
```
leroyvn, line 481: For consistency, let's use the singular:

```suggestion
                ("x_cell", "y_cell", "z_layer"),
```

Same at other sites where this is also used.

leroyvn (changes requested):
Hi @wint3ria, thanks for this PR! A few specific comments below. More general comments (subject to internal discussion) on our chat.

PR #602, opened 2026-09-09
[3DATM 3/6] Replace the blendphase plugin with the multiphase plugin

wint3ria:
# Description

Requires https://github.com/eradiate/mitsuba3/pull/46

This PR introduces the MultiPhaseFunction, able to blend an arbitrary number of sub-phases in one `multiphase` plugin. It simplifies both the weight calculation logic and the update dicts handling:

The now deprecated BlendPhaseFunction class used the `blendphase` plugin. This plugin allows to blend the phase functions of only 2 components. Eradiate typically needs to blend several phases together, resulting in the composition of a binary tree of `blendphase` plugins, weighted conditionally. This tree structure then needs to be mapped onto the flat key structure indexing the mitsuba update/traverse function.

The multiphase plugin thus allows to greatly simplify all this logic.

# Checklist


src/eradiate/scenes/atmosphere/_heterogeneous.py
```diff
@@ -114,6 +114,15 @@ def _particle_layers_validator(self, attribute, value):
                 "scaled individually"
             )
 
+    use_mis: bool = documented(
```
leroyvn: Since this argument is forwarded to the phase function, I suggest:
```suggestion
    use_phase_mis: bool = documented(
```

tests/01_unit/scenes/atmosphere/test_heterogeneous.py
```diff
@@ -363,23 +373,23 @@ def test_heterogeneous_absorbing_mol_atm(
...
+    weights = np.squeeze(mi_wrapper.parameters["weight_1.volume.data"])
 
     # Extract phase function weights
     inside_particle_layer = (atmosphere.geometry.grid.layers >= pl_bottom) & (
         atmosphere.geometry.grid.layers <= pl_top
     )
 
     # Outside the particle layer, the phase function weight should be zero.
-    assert np.all(weights[~inside_particle_layer] == 0.0)
+    assert np.all(weights.T[~inside_particle_layer] == 0.0)
```
leroyvn: Is the transpose needed because the previous pattern worked only by accident, or did the data shape change?

tests/01_unit/scenes/phase/test_multi.py
```diff
@@ -0,0 +1,359 @@
...
+    "weights, use_small_grid",
+    [
+        ([0.25, 0.75], False),
+        ([0.0, 1.0], False),
+        ([1.0, 0.0], False),
+        ([np.linspace(1.0, 0.0, 4), np.linspace(0.0, 1.0, 4)], True),
+    ],
+    ids=["scalar_even", "scalar_all_second", "scalar_all_first", "array"],
+)
+def test_multi_kernel_dict_2_components(mode_mono, weights, use_small_grid):
```
leroyvn: Given that the behaviour with any number of components should be the same, does this test add something compared to the 3-component one?

leroyvn (changes requested):
Hi @wint3ria, thanks a lot! Just a few comments.

PR #603, opened 2026-09-09
[3DATM 4/6] Spatially varying particle ensembles

wint3ria:
# Description

Adapt the `ParticleLayer` class to 3D grids. The class is now named `ParticleEnsemble`, referring the the unique distribution of particles shape they represent. This class can now handle variability in its geometry and OT alongside the X and Y coordinates of the scene. 

The vertical extent of the `ParticleEnsemble` can vary
image

Optical thickness can accept a 2D array of the shape of the grid's X and Y coordinates
image

`ArrayDistribution` class is also adapted to support X and Y dimensions
image

This class can be used for parametric configuration of 3D particles plumes geometry. Examples taken from https://github.com/eradiate/eradiate-tutorials/pull/5

# Checklist

### DIFF
--- a/docs/reference_api/scenes.rst
+++ b/docs/reference_api/scenes.rst
@@ -171,7 +171,7 @@ Quick access
    HomogeneousAtmosphere
    HeterogeneousAtmosphere
    MolecularAtmosphere
-   ParticleLayer
+   ParticleEnsemble
 
 **Particle distributions**
 
--- a/docs/release_notes/v1.3.x.md
+++ b/docs/release_notes/v1.3.x.md
@@ -13,6 +13,9 @@
   `r_eff`/`v_eff` support ({ghpr}`584`).
 * New `ppr_v1` data format for spatially varying particle profiles, like
   clouds ({ghpr}`585`).
+* {class}`.ParticleEnsemble`'s `bottom` and `top` altitudes can now vary per
+  horizontal grid cell. `tau_ref` can vary per cell too. Particle properties
+  are the same in every cell ({ghpr}`588`).
 
 ### Changed
 
@@ -38,6 +41,9 @@
 * 🖥️ `RayleighPhaseFunction` now builds its depolarization-factor volume with
   `generate_gridvolume()`, instead of hand-rolled `DictParameter`/
   `SceneParameter` code ({ghpr}`587`).
+* 🖥️ `ArrayParticleDistribution` now accepts 3D arrays (one per horizontal
+  cell), not just 1D. `validators.is_positive` now works on arrays too
+  ({ghpr}`588`).
 * 🖥️ Documentation now builds with Sphinx 9 and Python 3.11 ({ghpr}`580`).
 * 🖥️ Local docs build now mocks Mitsuba and Dr.Jit like the Read The Docs build
   does ({ghpr}`580`).
--- a/src/eradiate/scenes/atmosphere/_core.py
+++ b/src/eradiate/scenes/atmosphere/_core.py
@@ -61,8 +61,8 @@ atmosphere_factory.register_lazy_batch(
             {},
         ),
         (
-            "_particle_layer.ParticleLayer",
-            "particle_layer",
+            "_particle_ensemble.ParticleEnsemble",
+            "particle_ensemble",
             {},
         ),
     ],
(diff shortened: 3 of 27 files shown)


docs/reference_api/scenes.rst
```diff
@@ -171,7 +171,7 @@ Quick access
    HomogeneousAtmosphere
    HeterogeneousAtmosphere
    MolecularAtmosphere
-   ParticleLayer
+   ParticleEnsemble
```
leroyvn: See our offline discussion about the nomenclature.

docs/release_notes/v1.3.x.md
```diff
@@ -38,6 +41,9 @@
 * 🖥️ `RayleighPhaseFunction` now builds its depolarization-factor volume with
   `generate_gridvolume()`, instead of hand-rolled `DictParameter`/
   `SceneParameter` code ({ghpr}`587`).
+* 🖥️ `ArrayParticleDistribution` now accepts 3D arrays (one per horizontal
+  cell), not just 1D. `validators.is_positive` now works on arrays too
```
leroyvn: This deserves a separate line in the log. Also: I think it's better to rename the validator to `all_positive()`.

src/eradiate/scenes/atmosphere/_core.py
```diff
@@ -784,7 +784,7 @@ def _params_medium(self) -> dict[str, SceneParameter]:
             albedo_key = "albedo.volume.data"
             sigma_t_key = "sigma_t.volume.data"
         else:
-            raise ValueError(
+            raise TypeError(
```
leroyvn: Please don't change this, it's breaking and should be a separate change set (as part of a linter pass).

src/eradiate/scenes/atmosphere/_core.py
```diff
@@ -61,8 +61,8 @@
             {},
         ),
         (
-            "_particle_layer.ParticleLayer",
-            "particle_layer",
+            "_particle_ensemble.ParticleEnsemble",
+            "particle_ensemble",
```
leroyvn: I am thinking to keep an alias for compatibility (with a deprecation warning), but that would probably not be enough.

src/eradiate/scenes/atmosphere/_particle_dist.py
```diff
@@ -136,7 +138,7 @@ def _rate_validator(self, attribute, value):
                 f"while validating {attribute.name}: rate must be strictly positive"
             )
 
-    def __init__(self, rate: float = None, scale: float = None):
+    def __init__(self, rate: Optional[float] = None, scale: Optional[float] = None):
```
leroyvn: Since you have `from __future__ import annotations`:
```suggestion
    def __init__(self, rate: float | None = None, scale: float | None = None):
```

src/eradiate/scenes/atmosphere/_particle_ensemble.py
```diff
@@ -261,18 +320,18 @@ def eval_mfp(self, ctx: KernelContext) -> pint.Quantity:
     @cache_by_id
     def _eval_albedo_impl(self, w: pint.Quantity, grid: GridCoords) -> pint.Quantity:
         # Return albedo from dataset (without accounting for bypass switches)
-        # This routine is vectorized and returns an array of shape
-        # (n_wavelengths, n_layers)
+        # This routine returns an array of shape (n_wavelengths, [x, y, ]n_layers)
```
leroyvn: If nothing blocks it elsewhere, it might be simpler to return an array with always 4 dimensions.

tests/01_unit/scenes/atmosphere/test_3d.py
leroyvn: I suggest renaming `test_integration_3d.py`

src/eradiate/validators.py
```diff
@@ -56,14 +56,15 @@ def is_vector3(instance, attribute, value):
 
 def is_positive(_, attribute, value):
```
leroyvn: Let's rename this `all_positive()`.

leroyvn (changes requested):
Hi @wint3ria, thanks for this! Here are some specific comment, in addition to our discussions.

